PR #2, opened 2018-02-21
Preferences

Redth:
WIP: Adding preferences implementation

### DIFF
--- a/F50/F50.csproj
+++ b/F50/F50.csproj
@@ -22,7 +22,7 @@
     <Copyright>Copyright 2018</Copyright>
     <RepositoryUrl>https://github.com/xamarin/f50</RepositoryUrl>
     <PackageReleaseNotes>See: https://github.com/xamarin/f50 </PackageReleaseNotes>
-    <LangVersion>default</LangVersion>
+    <LangVersion>7.1</LangVersion>
     <DefineConstants>$(DefineConstants);</DefineConstants>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
--- a/F50/F50.csproj
+++ b/F50/F50.csproj
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;MonoAndroid80;uap10.0.16299</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">netstandard1.0;netstandard2.0;Xamarin.iOS10;MonoAndroid80;uap10.0.16299</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard1.0;netstandard2.0;Xamarin.iOS10;MonoAndroid80</TargetFrameworks>
     <AssemblyName>Xamarin.F50</AssemblyName>
     <RootNamespace>Xamarin.F50</RootNamespace>
     <PackageId>Xamarin.F50</PackageId>
--- a/F50/Platform/Platform.android.cs
+++ b/F50/Platform/Platform.android.cs
@@ -9,11 +9,11 @@ namespace Xamarin.F50
 		static Application application;
 		static ActivityLifecycleContextListener lifecycleListener;
 
-		internal static Context CurrentContext
-			=> lifecycleListener?.Context ?? application?.ApplicationContext;
+		internal static Context CurrentContext =>
+			lifecycleListener?.Context ?? application?.ApplicationContext;
 
-		internal static Activity CurrentActivity
-			=> lifecycleListener?.Activity;
+		internal static Activity CurrentActivity =>
+			lifecycleListener?.Activity;
 
 		public static void Init(Activity activity, Bundle bundle)
 		{
@@ -27,11 +27,11 @@ namespace Xamarin.F50
 	{
 		Activity currentActivity = null;
 
-		public Context Context 
-			=> currentActivity ?? Application.Context;
+		public Context Context =>
+			currentActivity ?? Application.Context;
 
-		public Activity Activity
-			=> currentActivity;
+		public Activity Activity =>
+			currentActivity;
 
 		public void OnActivityCreated(Activity activity, Bundle savedInstanceState)
 		{
--- a/F50/Platform/Platform.android.cs
+++ b/F50/Platform/Platform.android.cs
@@ -1,16 +1,16 @@
 ﻿using Android.App;
 using Android.Content;
 using Android.OS;
+using System;
 
 namespace Xamarin.F50
 {
 	public partial class Platform
 	{
-		static Application application;
 		static ActivityLifecycleContextListener lifecycleListener;
 
 		internal static Context CurrentContext =>
-			lifecycleListener?.Context ?? application?.ApplicationContext;
+			lifecycleListener?.Context ?? Application.Context;
 
 		internal static Activity CurrentActivity =>
 			lifecycleListener?.Activity;
@@ -18,20 +18,27 @@ namespace Xamarin.F50
 		public static void Init(Activity activity, Bundle bundle)
 		{
 			lifecycleListener = new ActivityLifecycleContextListener();
-			application = activity.Application;
-			application.RegisterActivityLifecycleCallbacks(lifecycleListener);
+			activity.Application.RegisterActivityLifecycleCallbacks(lifecycleListener);
 		}
 	}
 
 	class ActivityLifecycleContextListener : Java.Lang.Object, Application.IActivityLifecycleCallbacks
 	{
-		Activity currentActivity = null;
+		WeakReference<Activity> currentActivity = null;
 
 		public Context Context =>
-			currentActivity ?? Application.Context;
+			Activity ?? Application.Context;
 
-		public Activity Activity =>
-			currentActivity;
+		public Activity Activity
+		{
+			get
+			{
+				Activity a;
+				if (currentActivity.TryGetTarget(out a))
+					return a;
+				return null;
+			}
+		}
 
 		public void OnActivityCreated(Activity activity, Bundle savedInstanceState)
 		{
@@ -43,11 +50,12 @@ namespace Xamarin.F50
 
 		public void OnActivityPaused(Activity activity)
 		{
+			currentActivity.SetTarget(null);
 		}
 
 		public void OnActivityResumed(Activity activity)
 		{
-			currentActivity = activity;
+			currentActivity.SetTarget(activity);
 		}
 
 		public void OnActivitySaveInstanceState(Activity activity, Bundle outState)
--- a/F50/Platform/Platform.android.cs
+++ b/F50/Platform/Platform.android.cs
@@ -1,0 +1,65 @@
+﻿using Android.App;
+using Android.Content;
+using Android.OS;
+
+namespace Xamarin.F50
+{
+	public partial class Platform
+	{
+		static Application application;
+		static ActivityLifecycleContextListener lifecycleListener;
+
+		internal static Context CurrentContext
+			=> lifecycleListener?.Context ?? application?.ApplicationContext;
+
+		internal static Activity CurrentActivity
+			=> lifecycleListener?.Activity;
+
+		public static void Init(Activity activity, Bundle bundle)
+		{
+			lifecycleListener = new ActivityLifecycleContextListener();
+			application = activity.Application;
+			application.RegisterActivityLifecycleCallbacks(lifecycleListener);
+		}
+	}
+
+	internal class ActivityLifecycleContextListener : Java.Lang.Object, Application.IActivityLifecycleCallbacks
+	{
+		Activity currentActivity = null;
+
+		public Context Context 
+			=> currentActivity ?? Application.Context;
+
+		public Activity Activity
+			=> currentActivity;
+
+		public void OnActivityCreated(Activity activity, Bundle savedInstanceState)
+		{
+		}
+
+		public void OnActivityDestroyed(Activity activity)
+		{
+		}
+
+		public void OnActivityPaused(Activity activity)
+		{
+		}
+
+		public void OnActivityResumed(Activity activity)
+		{
+			currentActivity = activity;
+		}
+
+		public void OnActivitySaveInstanceState(Activity activity, Bundle outState)
+		{
+		}
+
+		public void OnActivityStarted(Activity activity)
+		{
+		}
+
+		public void OnActivityStopped(Activity activity)
+		{
+		}
+	}
+}
--- a/F50/Platform/Platform.android.cs
+++ b/F50/Platform/Platform.android.cs
@@ -23,7 +23,7 @@ namespace Xamarin.F50
 		}
 	}
 
-	internal class ActivityLifecycleContextListener : Java.Lang.Object, Application.IActivityLifecycleCallbacks
+	class ActivityLifecycleContextListener : Java.Lang.Object, Application.IActivityLifecycleCallbacks
 	{
 		Activity currentActivity = null;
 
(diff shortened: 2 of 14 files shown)


F50/Platform/Platform.android.cs
```diff
@@ -0,0 +1,65 @@
...
+		{
+		}
+
+		public void OnActivityPaused(Activity activity)
+		{
+		}
+
+		public void OnActivityResumed(Activity activity)
+		{
+			currentActivity = activity;
```
jamesmontemagno: should this happen on create as well? What about when it is paused/closed should we null out? Multiple activities at the same time?

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,130 @@
...
+		}
+
+		public bool Remove(string key)
+		{
+			using (var sharedPreferences = GetSharedPreferences())
+			{
+				if (sharedPreferences == null)
+					return false;
+
+				return sharedPreferences.Edit().Remove(key).Commit();
```
jamesmontemagno: will this also dispose of edit?

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,130 @@
...
+						break;
+					case bool b:
+						value = sharedPreferences.GetBoolean(key, b);
+						break;
+					case long l:
+						value = sharedPreferences.GetLong(key, l);
+						break;
+					case double d:
+						double dbl;
+						if (double.TryParse(sharedPreferences.GetString(key, null), out dbl))
```
jamesmontemagno: Double's are crazy when saving to string: https://github.com/jamesmontemagno/SettingsPlugin/blob/master/src/Plugin.Settings.Android/Settings.cs#L140

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,130 @@
...
+						break;
+				}
+			}
+
+			return (T)value;
+		}
 
+		ISharedPreferences GetSharedPreferences ()
+		{
+			var context = Platform.CurrentContext;
```
jamesmontemagno: We aren't using this at all, so we can remove this. Application.Context is good enough:

```
return string.IsNullOrWhiteSpace(fileName) ?
                PreferenceManager.GetDefaultSharedPreferences(Application.Context) :
                Application.Context.GetSharedPreferences(fileName, FileCreationMode.Private);
```

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,130 @@
...
-using System.Collections.Generic;
-using System.Text;
+using System.Globalization;
+using Android.App;
+using Android.Content;
+using Android.Preferences;
 
 namespace Xamarin.F50
 {
 	public partial class Preferences
```
jamesmontemagno: Do we need locks at all?

F50/Preferences/Preferences.ios.cs
```diff
@@ -1,14 +1,142 @@
...
+					break;
+				case float f:
+					value = userDefaults.FloatForKey(key);
+					break;
+			}
+
+			return (T)value;
+		}
+
+		NSUserDefaults GetUserDefaults() =>
```
jamesmontemagno: Can this just be a property? Will it ever change?

F50/Preferences/Preferences.ios.cs
```diff
@@ -1,14 +1,142 @@
...
+				case int i:
+					value = userDefaults.IntForKey(key);
+					break;
+				case bool b:
+					value = userDefaults.BoolForKey(key);
+					break;
+				case long l:
+					long lv;
+					var lnStr = userDefaults.StringForKey(key);
+					if (long.TryParse(lnStr, out lv))
```
jamesmontemagno: Is convert better? Convert.ToInt64(lnStr, System.Globalization.CultureInfo.InvariantCulture);

F50/Preferences/Preferences.ios.cs
```diff
@@ -1,14 +1,142 @@
...
+					userDefaults.SetString(s, key);
+					break;
+				case int i:
+					userDefaults.SetInt(i, key);
+					break;
+				case bool b:
+					userDefaults.SetBool(b, key);
+					break;
+				case long l:
+					userDefaults.SetString(l.ToString(), key);
```
jamesmontemagno: Should use invariant culture: Convert.ToString(value, System.Globalization.CultureInfo.InvariantCulture) i know i know, but probably best still

F50/Preferences/Preferences.shared.cs
```diff
@@ -10,10 +10,29 @@ namespace Xamarin.F50
...
-        }
-    }
+		public Preferences(string sharedName)
+		{
+			SharedName = sharedName;
+		}
+
+		public string SharedName { get; private set; }
+
+		public string Get(string key, string defaultValue)
```
jamesmontemagno: XML comments?

jamesmontemagno:
Also, can we add some helpers:

```
public static Boolean Get(this Preferences preferences, Boolean defaultValue = default(Boolean), [CallerMemberName] string key = null)
        => preferences.Get(key, defaultValue);
```

Thoughts?

F50/F50.csproj
```diff
@@ -22,7 +22,7 @@
     <Copyright>Copyright 2018</Copyright>
     <RepositoryUrl>https://github.com/xamarin/f50</RepositoryUrl>
     <PackageReleaseNotes>See: https://github.com/xamarin/f50 </PackageReleaseNotes>
-    <LangVersion>default</LangVersion>
+    <LangVersion>7.1</LangVersion>
```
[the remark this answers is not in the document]
Redth: I wanted to use pattern matching which requires 7.1+ which apparently is not what default is.

F50/Platform/Platform.android.cs
Redth: So I don't think on create. I think onresume is all you want to ensure you have the foremost activity always. 

We may need to null it out on pause though....

There shouldn't be multiple activities hitting onresume? Can you think of a way to actually do that?

F50/Preferences/Preferences.android.cs
Redth: Yeah. I wanted to evaluate if we truly did or not first. On android it seems in theory no but I'm seeing some posts that seem to have problems in practice so better safe than sorry here I think.

F50/Preferences/Preferences.android.cs
Redth: Nice catch.

F50/Preferences/Preferences.shared.cs
mattleibow: We are going to be doing docs using mdoc.

F50/Platform/Platform.android.cs
```diff
@@ -0,0 +1,65 @@
...
+		{
+			lifecycleListener = new ActivityLifecycleContextListener();
+			application = activity.Application;
+			application.RegisterActivityLifecycleCallbacks(lifecycleListener);
+		}
+	}
+
+	class ActivityLifecycleContextListener : Java.Lang.Object, Application.IActivityLifecycleCallbacks
+	{
+		Activity currentActivity = null;
```
mattleibow: Should this be a `WeakReference` at all? What are the implications of holding a reference to an object. I know it is in the foreground at this point, but we are effectively keeping a static reference to an object.

F50/Platform/Platform.shared.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Xamarin.F50
+{
+	public partial class Platform
```
mattleibow: I have a question about the name. Right now the namespace is F50, but what will this be in future. Will we ever separate the objects with namespaces? `Xamarin.F50.Preferences`? If we do, then the class name and namespace will clash.

F50/Preferences/Preferences.ios.cs
```diff
@@ -1,14 +1,142 @@
...
+			catch (Exception ex)
+			{
+				Console.WriteLine("Unable to clear items, Message: " + ex.Message);
+				return false;
+			}
+
+			return true;
+		}
+		
+		bool Set<T>(string key, T value)
```
mattleibow: Are generics going to cause any unnecessary overhead on iOS?

F50/Preferences/Preferences.ios.cs
```diff
@@ -1,14 +1,142 @@
...
+					break;
+			}
+
+			try
+			{
+				userDefaults.Synchronize();
+			}
+			catch (Exception ex)
+			{
+				Console.WriteLine("Unable to save: " + key, " Message: " + ex.Message);
```
mattleibow: These "errors", should we be using the native platform logging mechanisms or is `Console` fine?

F50/F50.csproj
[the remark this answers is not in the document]
jamesmontemagno: Love it! <3

F50/Platform/Platform.android.cs
jamesmontemagno: I don't think so......
Android 7 multi-window stuff... idk not sure anyone uses it, but one would always be active I assume

F50/Preferences/Preferences.android.cs
jamesmontemagno: I don't think it is needed at all and we should only require our context when we need it. In fact we should be able to access preferences in a service!

F50/Preferences/Preferences.shared.cs
jamesmontemagno: Ahhh... does that auto do it?

jamesmontemagno: Also does that get put into intellisense like xml docs?

F50/Platform/Platform.shared.cs
jamesmontemagno: We thought about this, but to be honest just easier to have 1 name space. If we have a clash that means we did something wrong for sure.

F50/Preferences/Preferences.shared.cs
mattleibow: I don't think so as it is a separate task.

F50/Preferences/Preferences.ios.cs
jamesmontemagno: Good catch.... I am not sure myself. It shouldn't be required actually, but hmmmmm interesting

F50/Preferences/Preferences.ios.cs
jamesmontemagno: Console is the go to from talking with @praeclarum

jamesmontemagno: I think we should create a "Tag" system like Android does though

F50/Preferences/Preferences.shared.cs
mattleibow: I think we need to think about this, mdoc is nice for separating docs from the code, but it then needs to be put back into the nuget... This is going to be extra work.

mattleibow: The advantage of mdoc is that we don't clutter the code with a 20-line snippet with weird xml escaping in the comment above a method or class. Also, we may want to write some awesome docs at the class, and this may take up space. And, if it is XML, contributors may feel more at east than modifying code to change a doc.

mattleibow: We could do like Xamarin.Forms and have a separate step that creates the package. I do not fear the .nuspec and think we can do better than the automation.

For example this is in the nuget:
```
<file src="...\bin\Release\netstandard1.0\Xamarin.F50.dll" target="lib\netstandard1.0\Xamarin.F50.dll" />
```
that should actually be (note the `ref` instead of `lib`):
```
<file src="...\bin\Release\netstandard1.0\Xamarin.F50.dll" target="ref\netstandard1.0\Xamarin.F50.dll" />
```

mattleibow: We need to support the case where the API is slightly different:

Android:
```csharp
class Platform {
    Init(Context);
}
```
iOS:
```csharp
class Platform {
    Init();
}
```

mattleibow: I just tested the above case, the `--preserve` flag does not remove members and basically forms a union of the APIs

mattleibow: I created this PR to add mdoc: https://github.com/xamarin/f50/pull/3

jamesmontemagno: Do you have an example of how we package up mdoc today in other nugets? It should be completely possible to do this

mattleibow: I do it "manually" for SkiaSharp. I use a nuspec and add the xml files to each lib directory in the nuspec.

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,159 @@
...
+							value = sharedPreferences.GetInt(key, i);
+							break;
+						case bool b:
+							value = sharedPreferences.GetBoolean(key, b);
+							break;
+						case long l:
+							value = sharedPreferences.GetLong(key, l);
+							break;
+						case double d:
+							value = Convert.ToDouble(sharedPreferences.GetString(key, null), NumberFormatInfo.InvariantInfo);
```
jamesmontemagno: Be aware here that this may cause major issues. The string for a double can be saved with an exponent I think and if that is the cause then it wont parse correctly.

Also there are issues with min/max as well which can cause issues. Well known in .NET I guess :(

```
 if (!double.TryParse(savedDouble, NumberStyles.Number | NumberStyles.AllowExponent, CultureInfo.InvariantCulture, out double outDouble))
                        {
                            var maxString = Convert.ToString(double.MaxValue, CultureInfo.InvariantCulture);
                            outDouble = savedDouble.Equals(maxString) ? double.MaxValue : double.MinValue;
                        }
```

F50/Preferences/Preferences.android.cs
```diff
@@ -1,14 +1,159 @@
...
+		}
+
+		T Get<T> (string key, T defaultValue)
+		{
+			lock (locker)
+			{
+				object value = null;
+				using (var sharedPreferences = GetSharedPreferences())
+				{
+					if (sharedPreferences == null)
```
jamesmontemagno: should we return the default value?

F50/Preferences/Preferences.android.cs
```diff
@@ -16,82 +16,64 @@ public bool ContainsKey(string key)
 			{
 				using (var sharedPreferences = GetSharedPreferences())
 				{
-					if (sharedPreferences == null)
-						return false;
-
 					return sharedPreferences.Contains(key);
```
jamesmontemagno: What if it is null? Should we do sharedPreferences?.Contains(key) ?? false;

F50/Preferences/Preferences.android.cs
```diff
@@ -16,82 +16,64 @@ public bool ContainsKey(string key)
...
+				using (var editor = sharedPreferences.Edit())
 				{
-					if (sharedPreferences == null)
-						return false;
-
-					using (var editor = sharedPreferences.Edit())
-					{
-						return editor.Remove(key).Commit();
-					}
+					editor.Remove(key).Commit();
```
jamesmontemagno: Same here.. editor?.Remove(key)?.Commit(); ?

F50/Preferences/Preferences.android.cs
Redth: There's nothing in the docs that suggests this should actually ever be null... It looks like worst case if you pass a null context into GetSharedPreferences(..) you'll get a java side null pointer exception thrown...

F50/Preferences/Preferences.android.cs
Redth: Same here, I don't see anything that refers to Edit() returning null, so I think we're ok to just bubble up any weird exceptions that we might encounter for now.

F50/Preferences/Preferences.android.cs
jamesmontemagno: Well, we return null if the context is null....

F50/Preferences/Preferences.android.cs
jamesmontemagno: roger that

F50/Preferences/Preferences.android.cs
jamesmontemagno: We should heavily unit test this one :)

F50/Preferences/Preferences.android.cs
Redth: We should probably throw an exception in that case with some reasonable explanation to make sure the app has been initialized.. I'm not sure Application.Context should ever be null?

Redth:
build

Redth:
build

Redth:
build

Redth:
build

Redth:
build

PR #3, opened 2018-02-21
WIP: Added the bits to get mdoc working

mattleibow:
Added the bits to get a nice documentation story.

Cake is just used so that we don't have to use a batch file or something worse. I basically `nuget restore` and then `msbuild` the solution. Once that is done, I copy the entire output to an `output` folder.

There are 3 main tasks:

 - "Build"
   This runs `nuget` and `msbuild`, and then copies everything to the output directory.
 - "UpdateDocs"
   This updates the repository docs.
 - "GenDocs"
   This generates the msxml docs for intellisense and saves it to the output directory.

The default task is just the "Build" target. There is a nice "CI" target that does everything.


F50/Preferences/Preferences.android.cs
```diff
@@ -6,6 +6,10 @@ namespace Xamarin.F50
 {
 	public partial class Preferences
 	{
+		public Preferences(global::Android.Content.Context context)
+		{
+		}
```
mattleibow: An example of a platform-specific overload.

build.cake
```diff
@@ -0,0 +1,95 @@
...
+            .Append("export-msxdoc")
+            .AppendSwitchQuoted("--out", "=", "./output/Xamarin.F50.xml")
+            .AppendQuoted("./docs/en/")
+    });
+});
+
+Task("Default")
+    .IsDependentOn("Build");
+
+Task("CI")
```
mattleibow: The "CI" target is what happens on CI...

docs/en/Xamarin.F50/Preferences.xml
```diff
@@ -0,0 +1,81 @@
...
+        <AssemblyVersion>1.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <Parameters />
+      <Docs>
+        <summary>To be added.</summary>
+        <remarks>To be added.</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName=".ctor">
+      <MemberSignature Language="C#" Value="public Preferences (Android.Content.Context context);" />
```
mattleibow: NOTE: the platform-specific member lives here.

build.cake
```diff
@@ -0,0 +1,95 @@
...
+    refDirs.AddRange(GetDirectories(refAssemblies + "Xamarin.Mac/v2.0"));
+
+    // the assemblies to generate documentation for
+    var assemblies = GetFiles("./output/*/*.dll");
+
+    // generate the docs
+    var args = new ProcessArgumentBuilder()
+        .Append("update")
+        .Append("--preserve")
+        .AppendSwitchQuoted("--out", "=", "./docs/en/");
```
mattleibow: The "en" bit is the language (English). We may want to support localization...

mattleibow:
At this point, the .nupkg does not contain the docs (actually it does, but this is the auto-gen docs from msbuild which is empty). I will look at getting that in next, but I think feedback is in order first.

F50/Preferences/Preferences.android.cs
Redth: Can we leave this bit out of this PR?  We may add these sorts of things for API's at some point, but I don't think Preferences will ever have it, as we simply need application context for it.

build.cake
```diff
@@ -0,0 +1,95 @@
...
+    EnsureDirectoryExists("./output/");
+    CopyDirectory($"./F50/bin/{configuration}", "./output/");
+});
+
+Task("UpdateDocs")
+    .IsDependentOn("Build")
+    .Does(() =>
+{
+    // the reference folders to locate assemblies
+    var refAssemblies = "C:/Program Files (x86)/Microsoft Visual Studio/*/*/Common7/IDE/ReferenceAssemblies/Microsoft/Framework/";
```
Redth: Do you know where these reside on the mac? Can we take a look at making it work here too?

docs/en/Xamarin.F50/Preferences.xml
Redth: So mdoc has no specific way of associating docs with platforms?

F50/Preferences/Preferences.android.cs
mattleibow: OK, this was just an example.

docs/en/Xamarin.F50/Preferences.xml
mattleibow: The docs just describe the API, in a platform-independent way. We could create separate docs for each assembly, but I wanted to add them all together so that the one docs can have everything in one place. Of course, having this item in the docs doesn't mean that it will appear in the intellisense as this member does not exists.

build.cake
mattleibow: We can, I was just getting the Windows bits working.

Redth:
build

mattleibow:
This PR has been superseded by #5

PR #5, opened 2018-02-23
Add mdoc

mattleibow:
Building the project will automatically run mdoc to generate the docs.

Once the project is compiled, a task can be run to take all the output assemblies and update the docs:

```
msbuild /t:MDocUpdateDocs /p:Cnfiguration=Release
```

This is for the #16 issue.

An example case of using everything could be this:

1. Write some nice new API
2. Build the solution (`msbuild /p:Configuration=Release`)
3. Update the docs (`msbuild /t:MDocUpdateDocs /p:Cnfiguration=Release`)
4. Submit PR

**NOTE: The `UpdateDocsWithMDoc` task is best run on Windows as the UWP library has to be processed. However, if this is run on macOS, then an assembly element will be missing from the `./docs/en/index.xml` file. Be careful not to commit this.**


Caboodle/Caboodle.csproj
```diff
@@ -24,36 +25,89 @@
...
+
+  <!--
+    ============================================================
+    AddMDocFilesToPackage
+
+    Make sure the mdoc output goes into the final package.
+    This runs during the package phase.
+    ============================================================
+  -->
+  <Target Name="AddMDocFilesToPackage" BeforeTargets="_GetBuildOutputFilesWithTfm" DependsOnTargets="SetupMDoc">
```
Redth: Should this not depend on `GenerateDocsWithMDoc`?

mattleibow: No as they are separate steps.

`GenerateDocsWithMDoc` runs as part of the core compilation of the individual assemblies. `AddMDocFilesToPackage` is a special task that is run when the .nuspec is being created. If there is a dependency, then it will fail as the `Pack` task does not actually start the compile.

mattleibow:
An interesting thing happens on mac... After running the `UpdateDocsWithMDoc` task, it fails with an error saying that it couldn't find that task that it just ran...

Not sure why, but since the job has already run, it should be safe to ignore.

https://gist.github.com/mattleibow/0942417c7b8aebadfb9e7184eba9a562

mattleibow:
build

mattleibow:
Closes #16

Issue #11, opened 2018-02-24
Clipboard API's


Redth:
Clipboard manipulation:

```csharp 
class Clipboard {
    void SetText(string text);
    void SetText(string text, string label);
    bool HasText { get; }
    Task<string> GetTextAsync();
}
```

 - UWP: https://docs.microsoft.com/en-us/windows/uwp/app-to-app/copy-and-paste
 - Android: https://developer.android.com/reference/android/content/ClipboardManager.html
 - iOS https://developer.apple.com/documentation/uikit/uipasteboard

Redth:
This needs API's drafted for sharing text content...

jamesmontemagno:
I changed to just clipboard :)

So SetText is easy, we should have a GetText.

Should analyze if they need to be async or not....

mattleibow:
@jamesmontemagno I had a look, and it doesn't appear to need async. Also, it doesn't matter since a copy/cut operation is fire-and-forget - there is no reason to get the app to wait until the clipboard accepts the text.

jamesmontemagno:
I think maybe get text on uwp may need async... ?

mattleibow:
Hmm. Appears so. The get on UWP is async, iOS and Android are not.

mattleibow:
I suppose we can have a `bool SetText()` and a `Task<bool> GetTextAsync()`...

mattleibow:
Also, iOS and UWP support copying images to the clipboard, do we want to do that?

Redth:
Images add a whole level of complexity which we do not need for v1. Also not sure how many people would realistically use it. Let's leave it out for now.

Also why bool return types? Let us throw an exception if something fails, not return false. 

GetTextAsync should be Task<string>.

jamesmontemagno:
I like the idea of exceptions in general for it.

jamesmontemagno:
Also agree images for this and for the share app should be v2 as Android can be complex.

PR #24, opened 2018-02-25
GH-11: Add Clipboard API

F1nZeR:
related to #11 

### PR Checklist ###

- [x] Consolidate commits as makes sense
- [x] Has Samples

### DIFF
--- a/Caboodle/Clipboard/Clipboard.android.cs
+++ b/Caboodle/Clipboard/Clipboard.android.cs
@@ -24,7 +24,8 @@ namespace Microsoft.Caboodle
 
         public static bool HasText => ClipboardManager.HasPrimaryClip;
 
-        public static Task<string> GetTextAsync() =>
-            Task.FromResult(HasText ? clipboardManager.PrimaryClip.GetItemAt(0).Text : null);
+        public static Task<string> GetTextAsync() => HasText
+            ? Platform.InvokeOnMainThread(() => clipboardManager.PrimaryClip.GetItemAt(0).Text)
+            : Task.FromResult<string>(null);
     }
 }
--- a/Caboodle/Clipboard/Clipboard.android.cs
+++ b/Caboodle/Clipboard/Clipboard.android.cs
@@ -20,7 +20,8 @@ namespace Microsoft.Caboodle
             }
         }
 
-        public static void SetText(string text) => ClipboardManager.PrimaryClip = ClipData.NewPlainText("Text", text);
+        public static Task SetTextAsync(string text) =>
+            Platform.InvokeOnMainThread(() => ClipboardManager.PrimaryClip = ClipData.NewPlainText("Text", text));
 
         public static bool HasText => ClipboardManager.HasPrimaryClip;
 
--- a/Caboodle/Clipboard/Clipboard.android.cs
+++ b/Caboodle/Clipboard/Clipboard.android.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        static ClipboardManager clipboardManager;
+
+        static ClipboardManager ClipboardManager
+        {
+            get
+            {
+                if (clipboardManager == null || clipboardManager.Handle == IntPtr.Zero)
+                    clipboardManager = (ClipboardManager)Application.Context.GetSystemService(Context.ClipboardService);
+
+                return clipboardManager;
+            }
+        }
+
+        public static void SetText(string text) => ClipboardManager.PrimaryClip = ClipData.NewPlainText("Text", text);
+
+        public static bool HasText => ClipboardManager.HasPrimaryClip;
+
+        public static Task<string> GetTextAsync() =>
+            Task.FromResult(HasText ? clipboardManager.PrimaryClip.GetItemAt(0).Text : null);
+    }
+}
--- a/Caboodle/Clipboard/Clipboard.ios.cs
+++ b/Caboodle/Clipboard/Clipboard.ios.cs
@@ -5,7 +5,8 @@ namespace Microsoft.Caboodle
 {
     public static partial class Clipboard
     {
-        public static void SetText(string text) => UIPasteboard.General.String = text;
+        public static Task SetTextAsync(string text) =>
+            Platform.InvokeOnMainThread(() => UIPasteboard.General.String = text);
 
         public static bool HasText => UIPasteboard.General.HasStrings;
 
--- a/Caboodle/Clipboard/Clipboard.ios.cs
+++ b/Caboodle/Clipboard/Clipboard.ios.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Caboodle
 
         public static bool HasText => UIPasteboard.General.HasStrings;
 
-        public static Task<string> GetTextAsync() => Task.FromResult(UIPasteboard.General.String);
+        public static Task<string> GetTextAsync() =>
+            Platform.InvokeOnMainThread(() => UIPasteboard.General.String);
     }
 }
--- a/Caboodle/Clipboard/Clipboard.ios.cs
+++ b/Caboodle/Clipboard/Clipboard.ios.cs
@@ -1,0 +1,14 @@
+﻿using System.Threading.Tasks;
+using UIKit;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        public static void SetText(string text) => UIPasteboard.General.String = text;
+
+        public static bool HasText => UIPasteboard.General.HasStrings;
+
+        public static Task<string> GetTextAsync() => Task.FromResult(UIPasteboard.General.String);
+    }
+}
--- a/Caboodle/Clipboard/Clipboard.uwp.cs
+++ b/Caboodle/Clipboard/Clipboard.uwp.cs
@@ -8,11 +8,13 @@ namespace Microsoft.Caboodle
 {
     public static partial class Clipboard
     {
-        public static void SetText(string text)
+        public static Task SetTextAsync(string text)
         {
             var dataPackage = new DataPackage();
             dataPackage.SetText(text);
             SetContent(dataPackage);
+
+            return Task.CompletedTask;
         }
 
         public static bool HasText => GetContent().Contains(StandardDataFormats.Text);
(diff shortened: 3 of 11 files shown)


Caboodle/Caboodle.csproj
```diff
@@ -58,4 +58,4 @@
     <Compile Include="**\*.ios.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
-</Project>
+</Project>
```
Redth: Please revert these changes to the .csproj

Looks like a tab change and then some newline stuff crept in.

mattleibow:
Congrats @F1nZeR on being the first to send in a PR! Hope to see you around on other issues and PRs!

bholmes:
bumping this thread to test if the CLA bot will respond.

mattleibow:
@F1nZeR I see you signed the CLA, thanks.

mattleibow:
I just noticed that you are using tabs here. I am sure we noticed that we switched from tabs to spaces, so it would be great if you could just changes those.

mattleibow:
@jamesmontemagno @Redth I think this is ready to be merged in, after the tabs are spaces.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        public static void SetText(string text)
+        {
+            SetText(text, "text");
+        }
+
+        public static void SetText(string text, string label)
```
mattleibow: I think we should remove the overloads with the `label` parameter as it provides no value. For android, we should just use `"Text"`.

F1nZeR:
Yes, totally agree.

Caboodle/Clipboard/Clipboard.ios.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using UIKit;
+using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        public static void SetText(string text)
```
mattleibow: Is there are reason this was not an expression bodied member?

mattleibow:
Beautiful! Thanks @F1nZeR for putting up with all our changes back and forth. We hope to create the perfect API for developers, thanks for jumping in so quick - even when we just started this project last week!

I have to wait for another final approval on this PR and then we can merge.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,34 @@
...
+using Android.Views;
+using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        public static void SetText(string text)
+        {
+            var clipboardManager = (ClipboardManager)Application.Context.GetSystemService(Context.ClipboardService);
```
jamesmontemagno: Should make this a static helper such as:
https://github.com/xamarin/Caboodle/blob/connectivity/Caboodle/Connectivity/Connectivity.android.cs#L17-L26

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,34 @@
...
+            var clip = ClipData.NewPlainText("Text", text);
```
jamesmontemagno: no need to create var clip, just set it.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,34 @@
...
+            clipboardManager.PrimaryClip = clip;
+        }
+
+        public static bool HasText
+        {
+            get
+            {
+                var clipboardManager = (ClipboardManager)Application.Context.GetSystemService(Context.ClipboardService);
```
jamesmontemagno: if we cleanup the manager code we can easily make expression bodied here.

Caboodle/Clipboard/Clipboard.uwp.cs
```diff
@@ -0,0 +1,36 @@
...
+            var dataPackage = new DataPackage();
+            dataPackage.SetText(text);
+            Windows.ApplicationModel.DataTransfer.Clipboard.SetContent(dataPackage);
+        }
+
+        public static bool HasText
+        {
+            get
+            {
+                var clipboardContent = Windows.ApplicationModel.DataTransfer.Clipboard.GetContent();
```
jamesmontemagno: Could make expression bodied member here as well, no need to create the var

jamesmontemagno: I would do:  
```
public static bool HasText =>
    Windows.ApplicationModel.DataTransfer.Clipboard.GetContent()?.Contains(StandardDataFormats.Text) ?? false;

**Comment by jamesmontemagno** [comment]
I have a few small tweaks here to cleanup some of the duplication of code in Android.

File: Caboodle/Clipboard/Clipboard.uwp.cs (hunk 7 above)
Comment (mattleibow) [reply Caboodle/Clipboard/Clipboard.uwp.cs]: That is going to be a long line, we have a style rule to make sure that it doesn't exceed 120 characters. Maybe make `Windows.ApplicationModel.DataTransfer` a using statement.

**Comment by jamesmontemagno** [comment]
We now have checks in place for StyleCop + Unit Tests + Samples ready to go.

Are you able to ensure everything pulls up from master and validates against StyleCop. Additionally can you add Unit Tests, Tests for Device Runners, and Samples.

**Comment by F1nZeR** [comment]
Should be ok now.

**Comment by Redth** [comment]
build

**Comment by Redth** [comment]
build

**Comment by Redth** [comment]
build

**Comment by Redth** [comment]
build

**Comment by Redth** [comment]
@F1nZeR finally got device tests running for Android/iOS... Looks like yours are failing on iOS due to not being on the UI Thread.  This suggests you will need to add some code to your iOS implementation to ensure the code runs on UI thread.  If you rebase on master, we have `PlatformBeginInvokeOnMainThread(Action)` available to use.

**Comment by F1nZeR** [comment]
@Redth probably the story is same as here: https://github.com/xamarin/Caboodle/blob/725c25604f1e050609bfcf33519c91316a137808/DeviceTests/Caboodle.DeviceTests.Shared/DeviceInfo_Tests.cs#L72

As you can see DeviceInfo API is also UI-thread dependent.
We should decide where UI-thread access logic should be: on dev side or on our side?

**Comment by Redth** [comment]
@F1nZeR you're right.  We'll talk about this issue this morning and update this thread when we have a consensus.  I'm pretty sure we'll end up doing the UI thread work in the library, otherwise it's potentially 1) inconsistent across platforms, and 2) more difficult to implement for the users.

**Comment by Redth** [comment]
@F1nZeR we're going to make sure the library does all the UIThread proxying work for the user.  We'll get Device Info stuff updated (thanks for pointing that out).

In the meantime if you could updating your implementation to call things within `Platform.BeingInvokeOnMainThread` that would be great!  I'm guessing you'll need to introduce TaskCompletonSource<T> in some places to achieve this.  Let me know if you have any questions.

**Comment by F1nZeR** [comment]
build

**Comment by F1nZeR** [comment]
@Redth is it possible to run build for the branch?

**Comment by Redth** [comment]
build

**Comment by F1nZeR** [comment]
build

**Comment by Redth** [comment]
approve

**Comment by Redth** [comment]
Looks like the `Set_Clipboard_Values` device test failed on iOS because it was not on a UI Thread when called.

File: Caboodle/Clipboard/Clipboard.ios.cs [hunk 8]
```diff
@@ -0,0 +1,16 @@
...
+using UIKit;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        public static Task SetTextAsync(string text) =>
+            Platform.InvokeOnMainThread(() => UIPasteboard.General.String = text);
+
+        public static bool HasText => UIPasteboard.General.HasStrings;
```
mattleibow: I just noticed that the test is failing:
https://jenkins.mono-project.com/job/Components-Caboodle-DeviceTests-PR-Mac/15/testReport/(root)/(empty)/DeviceTests_Clipboard_Tests_Set_Clipboard_Values_text____text___/

It appears iOS does not support this on background threads. Instead of trying to make this async and doing extra work, I think it might be easier just to remove this property and just return null from `GetTestAsync` like UWP. And do the same for Android.

jamesmontemagno: This is not correct, we should not invoke on main thread here, just set the String.

We should ONLY do an InvokeOnMainThread when we have events and they come from a background. Always assume they call the API from UIThread, else they need to do it.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Clipboard
+    {
+        static ClipboardManager clipboardManager;
```
jamesmontemagno: I am now kind of leaning towards not having this around being static and just having it part of the method. I think it can just query the current one. This will enable us to not have the object around in memory.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,32 @@
...
+            get
+            {
+                if (clipboardManager == null || clipboardManager.Handle == IntPtr.Zero)
+                    clipboardManager = (ClipboardManager)Application.Context.GetSystemService(Context.ClipboardService);
+
+                return clipboardManager;
+            }
+        }
+
+        public static Task SetTextAsync(string text) =>
```
jamesmontemagno: We should not be invoking on main thread here.

Caboodle/Clipboard/Clipboard.android.cs
```diff
@@ -0,0 +1,32 @@
...
+            Platform.InvokeOnMainThread(() => ClipboardManager.PrimaryClip = ClipData.NewPlainText("Text", text));
+
+        public static bool HasText => ClipboardManager.HasPrimaryClip;
+
+        public static Task<string> GetTextAsync() => HasText
```
jamesmontemagno: Same here, don't invoke on main thread.

mattleibow:
@F1nZeR Thanks for all your hard work, this was a small api, but I think the number of changes and comments exceeded the actual line count. LOL.

I merged it in and it looks good.

Issue #28, opened 2018-02-25
Geocoding API's


mattleibow:
A simple way to convert from a position to an address.

Geolocation API's: https://github.com/xamarin/Caboodle/issues/7

_NOTE: the `Position` type is used from #7 _
```csharp
static class Geocoding {
    // Requires a map key string on UWP
    // should throw an exception on UWP if this isn’t specified via the static MapKey property.
    static Task<IEnumerable<Address>> GetAddressesAsync(Position position);
    static Task<IEnumerable<Position>> GetPositionsForAddressAsync(string address);
}

// UWP-only partial
static class Geocoding {
    // UWP Only, and should only be in the UWP assembly
    // Users can call `Geolocation.MapKey = "blah"` from their UWP specific code (like where Inits are called).
    static string MapKey { get; set; }
}
```

```csharp
class Address{
    double Latitude { get; set; }
    double Longitude { get; set; }

    string AdminArea { get; set; }
    string CountryCode { get; set; }
    string CountryName { get; set; }
    string FeatureName { get; set; }
    string Locality { get; set; }
    string PostalCode { get; set; }
    string SubAdminArea { get; set; }
    string SubLocality { get; set; }
    string SubThoroughfare { get; set; }
    string Thoroughfare { get; set; }
}
```

jamesmontemagno:
Also should add:
```
Task<IEnumerable<Position>> GetPositionsForAddressAsync(string address, string mapKey = null);
```

jamesmontemagno:
I'll take this as I personally implemented it in my plugin

mattleibow:
This has been merged into master with https://github.com/xamarin/Caboodle/commit/9811d12b6d686b0f6f863d8c5e4eb9f796cc9327

Issue #17, opened 2018-02-24
Platform API's


Redth:
Some platforms will require some sort of initialization or otherwise platform specific calls.  This is the area they should be dealt with.

## Platform - Android Only
```csharp
public static class Platform
{
    internal static Activity CurrentActivity { get; }
    internal static Context CurrentContext { get; }

    // Should be called from the MainActivity in general
    public static void Init (Activity activity, Bundle bundle);

    // Should be called from the activity where permission requests occurred
    public static void OnRequestPermissionsResult (int requestCode, string[] permissions, Permission[] grantResults);
}
```

mattleibow:
Should this type be a static type? Or at least have a private constructor so that it cannot be instantiated?

mattleibow:
Something that we should consider is to keep all the platform stuff together. One example that comes to mind is #28. The `MapKey` exists there, should it exist here rather?

We could also do something like this:

```csharp
static class Platform {
    // ANDROID
    public static void Init (Activity activity, Bundle bundle);

    // UWP
    static class Geolocator {
        static string MapKey;
    }

    // IOS
    static class Email {
        // something for iOS only
    }
}
```

This way keeps all the Platform things in the `Platform` type, but grouped under a nested class that matches the API name. This may get messy, but keeps everything together.

mattleibow:
Another way would be to use a "settings" class:

```csharp
static class Platform {
    PlatformConfiguration  Configuration { get; set; }
    // OR
    void SetConfiguration(PlatformConfiguration config);
}

class PlatformConfiguration {
#if ANDROID
    // this could be used as part of the init...
    Application CurrentApplication { get; set; }
#endif

#if UWP
    static string MapKey { get; set; }
#endif

#if IOS
    // something for iOS only
#endif
}
```

Redth:
Here's another idea... What about a configuration file convention for these types of things?  Can we have a Caboodle.config type file here?

mattleibow:
That is a good idea, but we do need to support the case where the value is calculated or generated. Also, how are we could to build this in? As an embedded resource?

For an interim solution, we could create the configuration class - the type that will be used to deserialize the config. This way the user can configure the library from the platform projects as part of the `Init` (or a separate `SetConfig` method).

mattleibow:
@jamesmontemagno had an idea:

```csharp
static class Capabilities {
    static bool HasTelephony { get; }
}
```

Usage:

```csharp
if (Capabilities.HasTelephony) {
[...]

PR #29, opened 2018-02-25
Implementation of Geocoding #28

jamesmontemagno:
Geocoding APIs and helpers #28 

Also implemented "Types" for position and address with extensions for each project.


Caboodle/Types/Address.shared.cs
```diff
@@ -0,0 +1,54 @@
...
+            FeatureName = address.FeatureName;
+            PostalCode = address.PostalCode;
+            SubLocality = address.SubLocality;
+            Thoroughfare = address.Thoroughfare;
+            SubThoroughfare = address.SubThoroughfare;
+            SubAdminArea = address.SubAdminArea;
+            AdminArea = address.AdminArea;
+        }
+
+
```
F1nZeR: What about AddressLine property?
https://developer.android.com/reference/android/location/Address.html#getAddressLine(int)
https://developer.apple.com/documentation/corelocation/clplacemark/2890749-postaladdress

Caboodle/Types/Position.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public enum DistanceUnits
+    {
+        Kilometers,
+        Miles
+    }
+
+    public class Position
```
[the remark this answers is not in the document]
mattleibow: That is a very good idea.

mattleibow:
Before I review this, I think we should consider changing the name from`Position` to `GeoPosition`: https://github.com/xamarin/Caboodle/issues/7#issuecomment-368503485

Redth:
What about calling it `Location` instead?  I think there's less likely of naming collisions in the future using `Location` than `Position`.  We can always prefix some other class in the future that could possibly need to be called Location (eg: CursorLocation or whatever).

Caboodle/Types/Position.shared.cs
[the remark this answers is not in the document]
Redth: Perhaps let's use `Location` instead.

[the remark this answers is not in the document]
Redth: Alternatively, keep it as Position and worry about how to deal with collisions later.

mattleibow:
I agree.

F1nZeR:
GeoLocation maybe? :bowtie:

jamesmontemagno:
I think Location is correct as that is what other platforms call it.

jamesmontemagno:
So I have it set to "Location" instead of "Address" as this seems to fit better. A Location can have different things on it such as a GeoPoint, which is what Position used to be and better describes it. Right Not it just has a Lat/Long, but could have more information.

jamesmontemagno:
After further discussion we are going with Placemark and Location. Placemark as a Location. Follows closer to what iOS did for their API.

Caboodle/Geocoding/Geocoding.shared.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+	public partial class Geocoding
+    {
+        public static string MapKey { get; set; }
```
mattleibow: We may need to discuss the platform configuration first: #17

Caboodle/Geocoding/Geocoding.uwp.cs
```diff
@@ -0,0 +1,51 @@
...
+            var queryResults = await MapLocationFinder.FindLocationsAsync(address, null, 10);
+
+            if (queryResults?.Locations == null)
+                return null;
+
+            return queryResults.Locations.Select(l  => new Location
+			{
+				Latitude = l.Point.Position.Latitude,
+				Longitude = l.Point.Position.Longitude
+			});
```
mattleibow: Timestamp and make extension method.

Caboodle/Geocoding/Geocoding.uwp.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        static void SetMapKey()
```
mattleibow: If we do go the way of platform properties on each API, then we should call this something like: `EnsureMapKeyValid`.

Caboodle/Types/Location.shared.cs
```diff
@@ -0,0 +1,84 @@
...
+
+		public double Longitude { get; set; }
+
+
+		public static double CalculateDistance(Location locationStart, Location locationEnd, DistanceUnits units = DistanceUnits.Miles) =>
+			CalculateDistance(locationStart.Latitude, locationStart.Longitude, locationEnd.Latitude, locationEnd.Longitude, units);
+
+		public static double CalculateDistance(
+			double latitudeStart, double latitudeEnd,
+			double longitudeStart, double longitudeEnd, DistanceUnits units = DistanceUnits.Miles)
```
mattleibow: I recall a discussion about default parameters. They are nice to use but may bite us later. Maybe use overloads instead.

Caboodle/Types/Placemark.shared.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Caboodle
+{
+    public class Placemark
```
mattleibow: Could we maybe use the naming from UWP?
https://docs.microsoft.com/en-us/uwp/api/windows.services.maps.mapaddress

I find these properties very hard to understand. If I want the street number, I have to use the weird `SubThoroughfare` (which sounds like a highway or something) - UWP calls it `StreetNumber`.

Caboodle/Types/Placemark.shared.cs
```diff
@@ -0,0 +1,60 @@
...
+			PostalCode = placemark.PostalCode;
+			Locality = placemark.Locality;
+			SubLocality = placemark.SubLocality;
+			Thoroughfare = placemark.Thoroughfare;
+			SubThoroughfare = placemark.SubThoroughfare;
+			SubAdminArea = placemark.SubAdminArea;
+			AdminArea = placemark.AdminArea;
+		}
+
+		public Location Location { get; set; }
```
mattleibow: I like this as opposed to two properties. +1

Caboodle/Types/PlacemarkExtensions.android.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Caboodle
+{
+	public static partial class PlacemarkExtensions
+	{
+        private static readonly DateTime epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
mattleibow: This may be useful for other APIs, maybe we could move this to an internal property on `Platform` or another `Utils` type. In fact, the `Utils` type can be used to store members that are internal, but used across APIs - such as date conversion, maybe some "cached" math operations (eg: 2*PI or something)

Caboodle/Geocoding/Geocoding.android.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            using (var geocoder = new Geocoder(Application.Context))
+            {
+                var addressList = await geocoder.GetFromLocationNameAsync(address, 10);
+
+                return addressList?.Select(p => new Location
+				{
+                    Latitude = p.Latitude,
+                    Longitude = p.Longitude
+                });
```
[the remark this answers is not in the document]
mattleibow: I see there is already a `ToLocation` extension method. We could do something like this:
```csharp
return addressList?.Select(PlacemarkExtensions.ToLocation);
```

Caboodle/Geocoding/Geocoding.uwp.cs
jamesmontemagno: Probably ValidateMapKey? It is internal

Caboodle/Types/Location.shared.cs
jamesmontemagno: no worries, yeah can remove. I am cools

jamesmontemagno:
@mattleibow let me know what you think now. Love the recommendations :)

jamesmontemagno:
Yeah, I think we can move that later.

jamesmontemagno:
build

Caboodle/Caboodle.csproj
```diff
@@ -39,6 +39,10 @@
     </UpgradeBackupLocation>
     <OldToolsVersion>2.0</OldToolsVersion>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard1.0|AnyCPU'">
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
```
mattleibow: Is this needed? It looks suspicious.

Caboodle/Types/Placemark.shared.cs
mattleibow: Anyone have opinions on this?

Caboodle/Caboodle.csproj
jamesmontemagno: Yeah, soooooo this was an issue because I couldn't debug the samples without it! I think it could be simplified down.

jamesmontemagno:
@mattleibow fixed it up. How does it look now? Placemark is what iOS goes with so @Redth and I agreed on it.

PR #33, opened 2018-02-26
Add extension methods for preferences

jamesmontemagno:
Will allow:

```
public bool IsActive
{
   get => settings.Get();
   set => settings.Set(value);
}
```


jamesmontemagno:
So remove defaults. Correct it looks weird being combined with the multiple sets. I don't like that. Maybe we can do it so it is name different....

mattleibow:
build

jamesmontemagno:
I'mma close... don't think we need these TBH

Issue #25, opened 2018-02-25
PhoneDialer API's


mattleibow:
A simple API to open the default dialer, preloaded with a number and name.

```csharp
static class PhoneDialer {
    static bool IsSupported { get; }
    // should throw a NotSupportedOnDeviceException if the device is not capable.
    static void Open(string number);
}
```

For the discussion around exceptions, see #19

F1nZeR:
As for me it can be simplified:
```csharp
static class PhoneDialer
{
    static bool CanOpenDialer { get; }
    static void OpenDialer(string number);
    static void OpenDialer(string number, string name);
}
```

Just a few moments to clarify:
- there is no need in `Task *Async()`, because it seems to be sync everywhere;
- via this API all we want to do is to open phone dialer, so why do we have `Call` method? `OpenDial()` seems more natural;
- should we take into account CountryIso, or let the user decide?
- two methods `OpenDial(...)` can be simplified to a single `OpenDial(string number, string name = null)` or we can left these two methods?

mattleibow:
Thanks for your feedback. I am really agreeing with you here.

With regards to the methods, we would like to avoid the default parameters as this may make it harder to change in future.

jamesmontemagno:
I am thinking that in the Platform we can have "capabilities" instead of each API having a "can do this"

static class Capabilities
```
bool HasTelephony { get; }
```

```
static class PhoneDialer
{
    static void Open(string number);
}
```

I don't know if we need a name option as I don't think that is handled gracefully.

We should also avoid duplicating words if it is in the class name.

mattleibow:
I think we should move this (Capabilities) discussion to #17 - I have a longer comment there

mattleibow:
I removed the name parameter as it is just supported on UWP right now.

However, if this is a cool feature, we could probably just add it to the UWP version:

```csharp
class Dialer {
#if UWP
    void Open(string number, string name);
#endif
    void Open(string number);
}
```

This way we get a nice feature, but do not add it to the cross-platform API where it is not supported.
Thoughts?

Redth:
I think eventually we may add this sort of thing in, but let's stick to things that work everywhere for the most part for now.

PR #37, opened 2018-02-27
GH-25: Add PhoneDialer API

F1nZeR:
#25 

Will be updated according to related discussions.

### PR Checklist ###

- [ ] Consolidate commits as makes sense
- [x] Has Samples


F1nZeR:
I've added temporary exception classes.

Caboodle/PhoneDialer/PhoneDialer.android.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Microsoft.Caboodle
+{
+	public static partial class PhoneDialer
+	{
+		public static bool IsSupported
+		{
+			get
+			{
+				var packageManager = Application.Context.PackageManager;
```
jamesmontemagno: My thoughts here are to also check telephony is actually on the device.

Caboodle/PhoneDialer/PhoneDialer.ios.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.Caboodle
+{
+	public static partial class PhoneDialer
+	{
+		public static bool IsSupported
+		{
+			get
+			{
+				var nsUrl = CreateNsUrl("0000000000");
```
jamesmontemagno: shouldn't this append with: tel:

Caboodle/PhoneDialer/PhoneDialer.ios.cs
```diff
@@ -0,0 +1,51 @@
...
+				var canCall = UIApplication.SharedApplication.CanOpenUrl(nsUrl);
+				
+				if (canCall)
+				{
+					using (var netInfo = new CTTelephonyNetworkInfo())
+					{
+						var mnc = netInfo.SubscriberCellularProvider?.MobileNetworkCode;
+						return !string.IsNullOrEmpty(mnc) && mnc != "65535"; // 65535 stands for NoNetwordProvider
```
jamesmontemagno: don't need to check against null here.

Caboodle/PhoneDialer/PhoneDialer.ios.cs
```diff
@@ -0,0 +1,51 @@
...
+				}
+				
+				return false;
+			}
+		}
+
+		public static void Open(string number)
+		{
+			if (string.IsNullOrWhiteSpace(number))
+			{
```
jamesmontemagno: @mattleibow are we standardizing on need or not needing { around a single line if?

Caboodle/PhoneDialer/PhoneDialer.ios.cs
```diff
@@ -0,0 +1,51 @@
...
+			if (!IsSupported)
+			{
+				throw new CapabilityNotSupportedException();
+			}
+			
+			var nsUrl = CreateNsUrl(number);
+			UIApplication.SharedApplication.OpenUrl(nsUrl);
+		}
+
+		private NSUrl CreateNsUrl(string number)
```
jamesmontemagno: Should use expression bodied method here.

Caboodle/PhoneDialer/PhoneDialer.uwp.cs
```diff
@@ -0,0 +1,31 @@
...
+using Windows.ApplicationModel.Calls;
+
+namespace Microsoft.Caboodle
+{
+	public static partial class PhoneDialer
+	{
+		public static bool IsSupported =>
+			 Windows.Foundation.Metadata.ApiInformation.IsTypePresent("Windows.ApplicationModel.Calls.PhoneCallManager");
+
+		public static void Open(string number)
```
jamesmontemagno: Expression bodied Method here

Caboodle/PhoneDialer/PhoneDialer.ios.cs
```diff
@@ -0,0 +1,51 @@
...
+		public static bool IsSupported
+		{
+			get
+			{
+				var nsUrl = CreateNsUrl("0000000000");
+				var canCall = UIApplication.SharedApplication.CanOpenUrl(nsUrl);
+				
+				if (canCall)
+				{
+					using (var netInfo = new CTTelephonyNetworkInfo())
```
jamesmontemagno: What all is this check doing?

jamesmontemagno:
The other question I have is what should the user pass in here as it looks like ANdroid does special formatting....

Do they need +1 for instance? does 555.555.5555 work?

Caboodle/PhoneDialer/PhoneDialer.ios.cs
mattleibow: @jamesmontemagno I prefer always using braces for the simple reason that it avoids confusion due to the grouping. Also it makes changes easier - if we ever have to add another line, then we don't have to also add braces. But, this is a very opinionated thing. 

If we are following the .net foundation guides, then they say that we don't need it. So I guess we don't need them.
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

mattleibow: Then again, we don't really follow most of them.

Caboodle/PhoneDialer/PhoneDialer.ios.cs
F1nZeR: All these properties might be null.
https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/1616900-subscribercellularprovider
https://developer.apple.com/documentation/coretelephony/ctcarrier/1620324-mobilenetworkcode

Caboodle/PhoneDialer/PhoneDialer.ios.cs
F1nZeR: we are checking for SIM-card & network availability

mattleibow:
@F1nZeR to save you time merging and resolving conflicts, I did it here: https://github.com/F1nZeR/Caboodle/pull/1

NOTE: I left your files as tabs so that it (hopefully) won't break any local changes you are working on.

jamesmontemagno:
We now have checks in place for StyleCop + Unit Tests + Samples ready to go.

Are you able to ensure everything pulls up from master and validates against StyleCop. Additionally can you add Unit Tests, Tests for Device Runners, and Samples.

jamesmontemagno:
I think the only thing we are trying to figure out is if we want "IsSupported" or some enum there.

jamesmontemagno:
Pulling into a branch so I can work on it

PR #40, opened 2018-02-28
Added the testing framework

mattleibow:
Added the test projects and a few tests. This PR also has a large chunk of changes - mostly to the project files:

 - Added a basic Preferences sample to the sample app
 - Added the test framework and apps
 - Removed template readmes and other files
 - Changed the copyrights to Microsoft
 - The solution now correctly builds everything for AnyCPU

At this point, the tests and samples compile and run. But there is a bug in the Preferences API that will cause the samples to crash and the tests to fail. I decided not to put that fix in this PR as this is more for the samples and tests projects, not the actual code itself.

Based on the PR: https://github.com/xamarin/Caboodle/pull/32


mattleibow:
build

mattleibow:
build

PR #41, opened 2018-02-28
Fixed bugs in the preferences

mattleibow:
- Android/iOS: if the default value was `null`, then the `case string s` would not be true
 - UWP: don't try using the shared name if it was null

### DIFF
--- a/Caboodle/Preferences/Preferences.android.cs
+++ b/Caboodle/Preferences/Preferences.android.cs
@@ -88,9 +88,6 @@ namespace Microsoft.Caboodle
 				{
 					switch (defaultValue)
 					{
-						case string s:
-							value = sharedPreferences.GetString(key, s);
-							break;
 						case int i:
 							value = sharedPreferences.GetInt(key, i);
 							break;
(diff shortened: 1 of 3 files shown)


Caboodle/Preferences/Preferences.android.cs
```diff
@@ -121,6 +118,15 @@ T Get<T>(string key, T defaultValue)
 						case float f:
 							value = sharedPreferences.GetFloat(key, f);
 							break;
+						case string s:
+							// the case when the string is not null
+							value = sharedPreferences.GetString(key, s);
+							break;
+						default:
+							// the case when the string is null
+							if (typeof(T) == typeof(string))
```
jamesmontemagno: If we pass in null for a string we should throw an exception, as they aren't supported.

mattleibow: This is just the get. The user may what to get a string, and return null if there is no value.

jamesmontemagno: ahhh gotcha

Issue #14, opened 2018-02-24
Filesystem Helper API's


Redth:
Some simple and helpful abstractions for common Filesystem problems

Possible functionality:
 - Find Paths
   - Cache
   - Sync'd/Roaming
   - Data
   - User / Document

mattleibow:
Would this be something that would exist in a nested class under the Platform API:

```csharp
public class Platform {
    public class Directories {
        public static string AppData { get; }
    }
}
```

Redth:
Yeah my first thought was stick it in Platform as well.  Can we think of any other interesting bits we might be able to provide related to filesystem?

mattleibow:
I removed `Roaming` as this is not really supported the same way across devices. For example, on iOS the user must have an iCloud account and agree to use it.

This concept of synchronised data should be a separate one. This API is for local-only - even though it is backed up, it is still just for this device.

sparkie108:
The design should be validated against UWP, for example [ApplicationData.Current.LocalCacheFolder](https://docs.microsoft.com/en-us/uwp/api/Windows.Storage.ApplicationData#Windows_Storage_ApplicationData_LocalCacheFolder) returns a StorageFolder object not a string.

jamesmontemagno:
Correct, it would just be the .Path at the end :)

sparkie108:
if only it was that easy.  If you try the following code:
 ```
var tempDirPath = ApplicationData.Current.LocalCacheFolder.Path;
var tempFilePath = Path.Combine(tempDirPath, "test.txt");
File.WriteAllText(tempDirPath, "Hello World!");
```
The write throws a System.UnauthorizedAccessException.  The only way I could get writing temp files to work was using the StorageFolder/StorageFile classes.

**Comment by mattleibow** [comment]
Thanks for bringing this up, I will investigate.

**Comment by sparkie108** [comment]
Just been retesting this and noticed a typo in the quick test I did yesterday (oops)
```
var tempDirPath = ApplicationData.Current.LocalCacheFolder.Path;
var tempFilePath = Path.Combine(tempDirPath, "test.txt");
File.WriteAllText(tempFilePath, "Hello World!");
```
and it now works!  Looks like they fixed it in the .NET Std 2.0 update (which is what you are targeting).
So the proposed design should work fine.

PR #42, opened 2018-02-28
GH-14: Add FileSystem

mattleibow:
Adding the FileSystem APIs - #14 

**Checklist**

 - [x] API implementation
 - [x] Tests
 - [x] Samples
 - [x] Documentation

```csharp
static class FileSystem {
    // Returns a path to a directory that can be used to temporarily store data.
    // Potentially cleared by the user (eg: on Android in the device settings) or by the OS when space runs low.
    static string CacheDirectory { get; }

    // Returns a path to a directory that contains app data.
    // This directory may not be visible to the user.
    static string AppDataDirectory { get; }

    // Returns a path to a directory that contains user-visible data.
    // This directory may be visible/editable to the user.
    static string UserDataDirectory { get; }

    // Opens a file Stream to a file embedded in the app package.
    //     Android: Asset
    //     iOS: NSBundle.MainBundle
    //     UWP: Package.InstalledLocation
    static Task<Stream> OpenAppPackageFileAsync(string filename);
}
```

Implementation details:

Cache -> Not backed up and can be deleted | iOS -> CacheDir | Android -> CacheDir | UWP -> LocalCache Folder
AppData -> Backed up  | iOS -> Application Support | Android -> FileDir | UWP -> LocalFolder
UserData -> Backed up | iOS -> Documents | Android -> FileDir | UWP -> LocalFolder


jamesmontemagno:
@praeclarum can you give iOS an overview?

Caboodle/FileSystem/FileSystem.android.cs
```diff
@@ -0,0 +1,39 @@
...
+    public partial class FileSystem
+    {
+        private static string cache;
+        private static string appData;
+        private static string userData;
+
+        public static string CacheDirectory
+            => cache ?? (cache = Application.Context.CacheDir.AbsolutePath);
+
+        public static string AppDataDirectory
```
jamesmontemagno: FilesDir will get backed up automatically.

jamesmontemagno: Which is good :)

Caboodle/FileSystem/FileSystem.ios.cs
```diff
@@ -0,0 +1,43 @@
...
+    public static partial class FileSystem
+    {
+        private static string cache;
+        private static string appData;
+        private static string userData;
+
+        public static string CacheDirectory
+            => cache ?? (cache = GetDirectory(NSSearchPathDirectory.CachesDirectory));
+
+        public static string AppDataDirectory
```
jamesmontemagno: I guess should this be library or this one...? https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html  

This is the top-level directory for any files that are not user data files. You typically put files in one of several standard subdirectories. iOS apps commonly use the Application Support and Caches subdirectories; however, you can create custom subdirectories.

Use the Library subdirectories for any files you don’t want exposed to the user. Your app should not use these directories for user data files.

On Android we return the root.... aka Library

Caboodle/FileSystem/FileSystem.uwp.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static partial class FileSystem
+    {
+        private static string cache;
+        private static string appData;
+        private static string userData;
+
+        public static string CacheDirectory
+            => cache ?? (cache = ApplicationData.Current.LocalCacheFolder.Path);
+
+        public static string AppDataDirectory
```
jamesmontemagno: Should one of these be roaming? https://docs.microsoft.com/en-us/uwp/api/windows.storage.applicationdata#Examples

jamesmontemagno: So I think that "AppData" should just be "LocalFolder" as that is backed up which is the same as the other apps.

Caboodle/FileSystem/FileSystem.uwp.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Windows.ApplicationModel;
+using Windows.Storage;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class FileSystem
+    {
+        private static string cache;
```
jamesmontemagno: I think we can remove all these so they aren't cached IMHO

Caboodle/FileSystem/FileSystem.ios.cs
```diff
@@ -0,0 +1,39 @@
...
+        public static string UserDataDirectory
+            => GetDirectory(NSSearchPathDirectory.DocumentDirectory);
+
+        public static Task<Stream> OpenAppPackageFileAsync(string filename)
+        {
+            if (filename == null)
+                throw new ArgumentNullException(nameof(filename));
+
+            filename = filename.Replace("\\", "/");
+            var file = Path.Combine(NSBundle.MainBundle.BundlePath, filename);
```
praeclarum: I prefer to use `NSBundle.ResourcePath` and `PathForResource` since that is the more consistent API between iOS and Mac. Bundles on iOS just happened to put their resources in the `BundlePath` but they are not there in Mac so the above code would fail.

https://developer.xamarin.com/api/property/MonoTouch.Foundation.NSBundle.ResourcePath/

https://developer.xamarin.com/api/member/MonoTouch.Foundation.NSBundle.PathForResource/p/System.String/System.String/

But since you are abstracting the API, it doesn't really matter - you can just write different versions for iOS and Mac.

Caboodle/FileSystem/FileSystem.ios.cs
```diff
@@ -0,0 +1,39 @@
...
+            return Task.FromResult((Stream)File.OpenRead(file));
+        }
+
+        static string GetDirectory(NSSearchPathDirectory directory)
+        {
+            var dirs = NSSearchPath.GetDirectories(directory, NSSearchPathDomain.User);
+            if (dirs == null || dirs.Length == 0)
+            {
+                throw new ArgumentException(nameof(directory));
```
praeclarum: This isn't right - the accessor properties will throw if there are missing directories. Properties should not throw.

That said, in normal use this won't fail - but you can create some crazy machines where it will.

jamesmontemagno: I do: NSSearchPath.GetDirectories(NSSearchPathDirectory.CachesDirectory, NSSearchPathDomain.User)[0];

Caboodle/FileSystem/FileSystem.android.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public static string UserDataDirectory
+            => userData ?? (userData = Application.Context.FilesDir.AbsolutePath);
+
+        public static Task<Stream> OpenAppPackageFileAsync(string filename)
+        {
+            if (filename == null)
+                throw new ArgumentNullException(nameof(filename));
+
+            filename = filename.Replace("\\", "/");
```
Redth: This is probably ok, but should we use `Path.DirectorySeparatorChar` instead?

Caboodle/FileSystem/FileSystem.ios.cs
Redth: @praeclarum what would you like the property to do in that case? return null?

Caboodle/FileSystem/FileSystem.uwp.cs
```diff
@@ -0,0 +1,29 @@
...
+    public static partial class FileSystem
+    {
+        public static string CacheDirectory
+            => ApplicationData.Current.LocalCacheFolder.Path;
+
+        public static string AppDataDirectory
+            => ApplicationData.Current.LocalFolder.Path;
+
+        public static string UserDataDirectory
+            => ApplicationData.Current.LocalFolder.Path;
```
Redth: Is this going to cause any chaos that `AppDataDirectory` and `UserDataDirectory` are the same path?  Like if a user was to `Directory.Delete(AppDataDirectory)` without understanding that also wipes `UserDataDirectory` ?

Caboodle/FileSystem/FileSystem.ios.cs
praeclarum: Yeah that's up to whatever contract @jamesmontemagno wants to sign with the users. ;-) Perhaps breaking the properties don't throw rule is fine here since the other implementations don't seem to handle error cases.

Redth: Presumably returning null would just delay the exception anyway?  If users aren't checking for null, their code is going to throw if they attempt to use it.

Issue #26, opened 2018-02-25
SMS API's


mattleibow:
A simple API to send an SMS.

```csharp
static class Sms {
    // whether or not the device can open a compose dialog
    static bool IsComposeSupported { get; }

    // whether or not the device can send in the background
    static bool IsSendSupported { get; }

    // open the compose window, throw if not supported
    static Task ComposeAsync(SmsMessage message);

    // send a message in the background, throw if not supported
    // this is supported only on Android and UWP
    static Task SendAsync(SmsMessage message);
}

class SmsMessage {
    SmsMessage(string body, string recipient);

    string Body { get; set; }
    string Recipient { get; set; }
}
```

For the discussion around exceptions, see #19

F1nZeR:
What about Multi-SIM card support? It is pretty common situation on Android platform. 
Should it be supported as a separate prop on droid side?

Geertvdc:
Hi @jamesmontemagno i'll try to see if i can implement this since it is up for grabs

jamesmontemagno:
I believe there is a pr

jamesmontemagno:
I believe there is a pr

Geertvdc:
oh yeah now i see. might wanna move the "up for grabs" then cause i thought i could pick this one up

mattleibow:
After some more discussion (which is always good) we have decided to try another approach - copying the style of emails: https://github.com/xamarin/Caboodle/issues/27

jamesmontemagno:
Looks perfect. Should it be "body" or "message", should be consistent.

mattleibow:
I used "body" as the type is the message. This is what email does.
I was thinking `var message = new SmsMessage()` and then `message.Message` is a bit weird.

jamesmontemagno:
ahhhh, yeah makes sense. I think the method's properties should match :)

jamesmontemagno:
oh wait
nm. ignore me

PR #65, opened 2018-03-08
GH-26: Add SMS API 

ravinderjangra:
### Description of Change ###

Added the SMS APIs from #26 .

### Bugs Fixed ###

- No bugs here :)

### API Changes ###

Added: 
 
- see #26 

### Behavioral Changes ###

New Features

### PR Checklist ###

- [ ] Consolidate commits as makes sense

### DIFF
--- a/Caboodle/Sms/Sms.android.cs
+++ b/Caboodle/Sms/Sms.android.cs
@@ -1,0 +1,33 @@
+﻿using Android.App;
+using Android.Content;
+using Android.Net;
+using Android.Telephony;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => true;
+
+        static bool GetCanSendSmsInBackground() => true;
+
+        public static void PlatformSendSms(string recipient, string message)
+        {
+            Uri smsUri;
+            if (!string.IsNullOrWhiteSpace(recipient))
+                smsUri = Uri.Parse("smsto:" + recipient);
+            else
+                smsUri = Uri.Parse("smsto:");
+
+            var smsIntent = new Intent(Intent.ActionSendto, smsUri);
+            smsIntent.PutExtra("sms_body", message);
+            Platform.CurrentContext.StartActivity(smsIntent);
+        }
+
+        public static void PlatformSendSmsInBackground(string recipient, string message)
+        {
+            var smsManager = SmsManager.Default;
+            smsManager.SendTextMessage(recipient, null, message, null, null);
+        }
+    }
+}
--- a/Caboodle/Sms/Sms.android.cs
+++ b/Caboodle/Sms/Sms.android.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Caboodle
 
         static bool GetCanSendSmsInBackground() => true;
 
-        public static void PlatformSendSms(string recipient, string message)
+        private static void PlatformSendSms(string recipient, string message)
         {
             Uri smsUri;
             if (!string.IsNullOrWhiteSpace(recipient))
@@ -24,7 +24,7 @@ namespace Microsoft.Caboodle
             Platform.CurrentContext.StartActivity(smsIntent);
         }
 
-        public static void PlatformSendSmsInBackground(string recipient, string message)
+        private static void PlatformSendSmsInBackground(string recipient, string message)
         {
             var smsManager = SmsManager.Default;
             smsManager.SendTextMessage(recipient, null, message, null, null);
--- a/Caboodle/Sms/Sms.ios.cs
+++ b/Caboodle/Sms/Sms.ios.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using MessageUI;
+using UIKit;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => MFMessageComposeViewController.CanSendText;
+
+        static bool GetCanSendSmsInBackground() => false;
+
+        public static void PlatformSendSms(string recipient, string message)
+        {
+            var smsController = new MFMessageComposeViewController() { Body = message };
+
+            if (!string.IsNullOrWhiteSpace(recipient))
+            {
+                var recipients = recipient.Split(';');
+                if (recipients.Length > 0)
+                    smsController.Recipients = recipients;
+            }
+
+            smsController.Finished += (sender, e) => smsController.DismissViewController(true, null);
+            UIApplication.SharedApplication.KeyWindow.RootViewController.PresentViewController(smsController, true, null);
+
+            ////smsController.Body = message;
+
+            ////EventHandler<MFMessageComposeResultEventArgs> handler = null;
+            ////handler = (sender, args) =>
+            ////   {
+            ////       smsController.Finished -= handler;
+
+            ////       if (!(sender is UIViewController uiViewController))
+            ////       {
+            ////           throw new ArgumentException("sender");
+            ////       }
+
+            ////       uiViewController.DismissViewController(true, () => { });
+            ////   };
+
+            ////smsController.Finished += handler;
+
+            ////smsController.PresentUsingRootViewController();
+        }
+
+        public static void PlatformSendSmsInBackground(string recipient, string message)
+        {
+            throw new PlatformNotSupportedException("Sending SMS in background not supported on iOS");
+        }
+    }
+}
--- a/Caboodle/Sms/Sms.ios.cs
+++ b/Caboodle/Sms/Sms.ios.cs
@@ -23,25 +23,6 @@ namespace Microsoft.Caboodle
 
             smsController.Finished += (sender, e) => smsController.DismissViewController(true, null);
             UIApplication.SharedApplication.KeyWindow.RootViewController.PresentViewController(smsController, true, null);
-
-            ////smsController.Body = message;
-
-            ////EventHandler<MFMessageComposeResultEventArgs> handler = null;
-            ////handler = (sender, args) =>
-            ////   {
-            ////       smsController.Finished -= handler;
-
-            ////       if (!(sender is UIViewController uiViewController))
-            ////       {
-            ////           throw new ArgumentException("sender");
-            ////       }
-
-            ////       uiViewController.DismissViewController(true, () => { });
-            ////   };
-
-            ////smsController.Finished += handler;
-
-            ////smsController.PresentUsingRootViewController();
         }
 
         public static void PlatformSendSmsInBackground(string recipient, string message)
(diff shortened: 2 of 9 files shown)


ravinderjangra:
Still working on it. But, Is it going in right direction?

Caboodle/Sms/Sms.android.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => true;
+
+        static bool GetCanSendSmsInBackground() => true;
+
+        public static void PlatformSendSms(string recipient, string message)
```
mattleibow: This should be private.

Caboodle/Sms/Sms.android.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            var smsIntent = new Intent(Intent.ActionSendto, smsUri);
+            smsIntent.PutExtra("sms_body", message);
+            Platform.CurrentContext.StartActivity(smsIntent);
+        }
+
+        public static void PlatformSendSmsInBackground(string recipient, string message)
+        {
+            var smsManager = SmsManager.Default;
+            smsManager.SendTextMessage(recipient, null, message, null, null);
```
mattleibow: This should be private.

Caboodle/Sms/Sms.ios.cs
```diff
@@ -0,0 +1,33 @@
...
+                var recipients = recipient.Split(';');
+                if (recipients.Length > 0)
+                    smsController.Recipients = recipients;
+            }
+
+            smsController.Finished += (sender, e) => smsController.DismissViewController(true, null);
+            UIApplication.SharedApplication.KeyWindow.RootViewController.PresentViewController(smsController, true, null);
+        }
+
+        public static void PlatformSendSmsInBackground(string recipient, string message)
```
mattleibow: This should be private.

Caboodle/Sms/Sms.ios.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => MFMessageComposeViewController.CanSendText;
+
+        static bool GetCanSendSmsInBackground() => false;
+
+        public static void PlatformSendSms(string recipient, string message)
```
mattleibow: This should be private.

Caboodle/Sms/Sms.netstandard.cs
```diff
@@ -0,0 +1,14 @@
...
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => throw new NotImplentedInReferenceAssembly();
+
+        static bool GetCanSendSmsInBackground() => throw new NotImplentedInReferenceAssembly();
+
+        public static void PlatformSendSms(string recipient, string message) =>
+            throw new NotImplentedInReferenceAssembly();
+
+        public static void PlatformSendSmsInBackground(string recipient, string message) => throw new NotImplentedInReferenceAssembly();
```
mattleibow: This should be private.

Caboodle/Sms/Sms.uwp.cs
```diff
@@ -0,0 +1,35 @@
...
+            var msg = new ChatMessage { Body = message };
+            if (!string.IsNullOrWhiteSpace(recipient))
+                msg.Recipients.Add(recipient);
+
+#pragma warning disable 4014
+            ChatMessageManager.ShowComposeSmsMessageAsync(msg);
+#pragma warning restore 4014
+        }
+
+        public static void PlatformSendSmsInBackground(string recipient, string message)
```
mattleibow: Should be private.

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        public static bool CanSendSms => GetCanSendSms();
```
mattleibow: I think this API is simple enough just to use the public API in each platform - no need to have a property and the methods. BUT... that is preference. I personally like your way as it helps with maintenance and development, but this is not my repository. 😄 @jamesmontemagno @Redth , what do you think?

Caboodle/Sms/Sms.uwp.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => true;
+
+        static bool GetCanSendSmsInBackground() => true;
+
+        public static void PlatformSendSms(string recipient, string message)
```
mattleibow: Should be private.

Caboodle/Sms/Sms.uwp.cs
mattleibow: Should be private.

Samples/Caboodle.Samples/View/SMSPage.xaml.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace Caboodle.Samples.View
+{
+    [XamlCompilation(XamlCompilationOptions.Compile)]
```
mattleibow: This is not needed here as it is already set for all pages in the App.xaml.cs.

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (recipient == null)
+                throw new ArgumentNullException(nameof(recipient));
+
+            if (message == null)
+                throw new ArgumentNullException(nameof(message));
+
+            if (CanSendSmsInBackground)
+            {
+                PlatformSendSmsInBackground(recipient, message);
+            }
```
mattleibow: I think this should throw if it can't send, not fail silently.

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,39 @@
...
+        public static void SendSms(string recipient = null, string message = null)
+        {
+            if (recipient == null)
+                throw new ArgumentNullException(nameof(recipient));
+
+            if (message == null)
+                throw new ArgumentNullException(nameof(message));
+
+            if (CanSendSms)
+                PlatformSendSms(recipient, message);
```
mattleibow: I think this should throw if it can't send, not fail silently.

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        public static bool CanSendSms => GetCanSendSms();
+
+        public static bool CanSendSmsInBackground => GetCanSendSmsInBackground();
+
+        public static void SendSms(string recipient = null, string message = null)
```
mattleibow: I think these two methods should be combined into one that takes an enum that specifies whether to send in background or not. Something along the lines of this:
```csharp
enum SmsSendType {
    Foreground,
    Background,
    // if the device supports background sending, the use that, otherwise use foreground
    PreferBackground
}
```

Caboodle/Sms/Sms.shared.cs
ravinderjangra: Already throwing on platform code.

Caboodle/Sms/Sms.shared.cs
ravinderjangra: Already throwing on platform code.

Caboodle/Sms/Sms.shared.cs
ravinderjangra: @mattleibow  I think this is simple and have self explanatory method names. But I can make changes if you want it that way,

Redth:
@ravinderjangra we have a spec written for this api in #26 so please make sure your public API surface matches that :)

Caboodle/Sms/Sms.shared.cs
mattleibow: What I mean is that this should not be guarded by the `if`. If I call this on iOS, it will just do nothing and I will never know.

Caboodle/Sms/Sms.android.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Android.App;
+using Android.Content;
+using Android.Net;
+using Android.Telephony;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        static bool GetCanSendSms() => true;
```
jamesmontemagno: Are we sure that every device can send an Sms? this seems like a false assumption. We should validate that there is an app installed that can handle the "smsto:" we should probably make a Platform internal static to validate Intents and the Uri.

Caboodle/Sms/Sms.shared.cs
jamesmontemagno: Agreed, I would remove

Caboodle/Sms/Sms.uwp.cs
```diff
@@ -0,0 +1,35 @@
...
+        static bool GetCanSendSmsInBackground() => true;
+
+        private static void PlatformSendSms(string recipient, string message)
+        {
+            var msg = new ChatMessage { Body = message };
+            if (!string.IsNullOrWhiteSpace(recipient))
+                msg.Recipients.Add(recipient);
+
+#pragma warning disable 4014
+            ChatMessageManager.ShowComposeSmsMessageAsync(msg);
```
jamesmontemagno: This is async... so really this should be a Task and the SendSms should be SendSmsAsync

Caboodle/Sms/Sms.uwp.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        private static void PlatformSendSmsInBackground(string recipient, string message)
+        {
+                var sendingMessage = new SmsTextMessage2
+                {
+                    Body = message,
+                    To = recipient
+                };
+
+                SmsDevice2.GetDefault().SendMessageAndGetResultAsync(sendingMessage).AsTask().Wait();
```
jamesmontemagno: Making the entire method a Task will remove this Wait() which can lock the UI thread.

Additionally GetDefault() could be null here so the CanSendInbackground should check for that.

jamesmontemagno:
Thanks for your PR, going through it we have certain design decisions and have created a new PR with our original design with our new findings. This PR was also too close to other open source projects in the plugin community and we only accept a contribution if it is granted form the original creator. Feel free to add comments on the other PR. Closing for now.

jamesmontemagno:
Talked with some of the coreclr team today about how they work with and optimize the PR workflow when contributing code from other repos. We will update this guideline to better handle this situation. Still kinda new to working with some of the PR open source official packages. But all in all want to thank you for your original PR as it set us on a good path.

PR #81, opened 2018-03-14
Make Geocoding testable from desktop unit test projects

kzu:
This showcases how a minimal change in the netstandard version of the
API can trivially allow for fully testable API invocations without
incurring in *any* of the run-time performance compromises of switching
the platform implementations to use either unnecessary interface/virtual
calls or even instance method call overhead ('cause of the null reference
check that's otherwise not necessary).

This is an effective compromise that covers the majority of the testing
scenarios (unit testing from view models down) while still preserving
the desired run-time characteristics of the project.

### Description of Change ###

Tweak *just* the netstandard version of an API to make it trivially 
testable from desktop unit tests.

### Bugs Fixed ###

The overwhelming feedback that un-testable APIs are not cool ;)

Also, my personal belief that forcing everyone to invent their own 
abstractions and achieving the same through unnecessary indirections  
that *do* impact ultimate run-time performance is bad for both 
our users (developers) as well as their customers (users of their apps).

### API Changes ###

Added: 
 
- IGeocoding (as an example only, it's not strictly required), *only* to the 
  netstandard Geocoding 
- Geocoding.Current, to allow replacing the implementation with a fake/mock

### Behavioral Changes ###

No behavioral changes whatsoever are introduced in the platform code. The 
netstandard code passes all existing unit tests too, since by default it will still 
throw the same exception as before, unless a test overrides the implementation.

### PR Checklist ###

- [x] Consolidate commits as makes sense

### DIFF
--- a/Caboodle.sln
+++ b/Caboodle.sln
@@ -36,6 +36,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Caboodle.DeviceTests.iOS", 
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Caboodle.DeviceTests.UWP", "DeviceTests\Caboodle.DeviceTests.UWP\Caboodle.DeviceTests.UWP.csproj", "{4BD0D88F-7E7A-4C3B-9E34-BF3717A8FF4B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Caboodle.Samples.Tests", "Samples\Caboodle.Samples.Tests\Caboodle.Samples.Tests.csproj", "{ADFA598A-957E-4D3F-BD79-E75F3400D52D}"
+EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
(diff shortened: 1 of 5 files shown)


jamesmontemagno:
What will happen here if someone was to create a device runner with xunit or nunit tests? I assume it will not use the mocks and still use our implementation?

kzu:
Yes, of course. In fact, they will not see the API at all in their device tests, so it will fail at compile time, since for devices, the Caboodle will *always* run at "full speed" with no abstractions whatesoever. At that point they will have to find a more creative workaround (like the suggested usage of something that does IL rewriting to allow faking static methods). 

It's a trade-off, but one we can more easily get buy in from @migueldeicaza I'd say ;)

kzu:
And I'm calling this pattern "test and switch", just so you know ;)

ghuntley:
Thank-you @kzu

kzu:
I'll take it as an "I think it's a good idea" @ghuntley ;)

jamesmontemagno:
Thanks for the PR kzu, we have analyzed it and like the idea, but not ready to implement it, but will keep it in mind for the future. We have documented our approach for developers in our README.

PR #85, opened 2018-03-14
Fix the docs generator

mattleibow:


Caboodle/mdoc.targets
```diff
@@ -86,7 +86,43 @@
       <MDocReferenceAssemblies>@(MDocReferenceAssembly -> '--lib=&quot;%(Identity)&quot;', ' ')</MDocReferenceAssemblies>
       <MDocAssemblies>@(MDocAssembly -> '&quot;%(Identity)&quot;', ' ')</MDocAssemblies>
     </PropertyGroup>
-    <Exec Command="$(_ManagedExeLauncher) &quot;$(MDocToolPath)&quot; update --preserve --out=&quot;$(MDocDocumentationDirectory)&quot; $(MDocReferenceAssemblies) $(MDocAssemblies)" />
+    <Exec Command="$(_ManagedExeLauncher) &quot;$(MDocToolPath)&quot; update --debug --preserve --out=&quot;$(MDocDocumentationDirectory)&quot; $(MDocReferenceAssemblies) &quot;%(MDocAssembly.Identity)&quot;" />
```
mattleibow: Here we run the tool once for each assembly, this fixes the case where mdoc doesn't understand that two assemblies might represent the same api

Caboodle/mdoc.targets
```diff
@@ -86,7 +86,43 @@
...
+              Encoding = new UTF8Encoding(),
+              Indent = true,
+              NewLineChars = "\n",
+              OmitXmlDeclaration = true,
+          };
+          using (var writer = XmlWriter.Create(file, settings)) {
+              xdoc.Save(writer);
+              writer.Flush();
+          }
+          File.AppendAllText(file, "\n");
```
mattleibow: A nice task that unifies the docs format. this is good for after editing in DocWriter or any other editor that may change XML formatting rules.

PR #91, opened 2018-03-15
GH-26: Add SMS API

mattleibow:
### Description of Change ###

Adding the SMS API as designed in #26.

### Bugs Fixed ###

- Related to issue #26

### API Changes ###

Adding the SMS API as designed in #26.

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Caboodle/Sms/Sms.ios.cs
+++ b/Caboodle/Sms/Sms.ios.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Threading.Tasks;
+using MessageUI;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        public static bool IsComposeSupported
+            => MFMessageComposeViewController.CanSendText;
+
+        public static Task ComposeAsync(SmsMessage message)
+        {
+            if (message == null)
+                throw new ArgumentNullException(nameof(message));
+            message.Validate();
+
+            // do this first so we can throw as early as possible
+            var controller = Platform.GetCurrentViewController();
+
+            // show the controller
+            var tcs = new TaskCompletionSource<bool>();
+            var messageController = new MFMessageComposeViewController
+            {
+                Body = message.Body,
+                Recipients = new[] { message.Recipient }
+            };
+            messageController.Finished += (sender, e) =>
+            {
+                messageController.DismissViewController(true, null);
+                tcs.SetResult(e.Result == MessageComposeResult.Sent);
+            };
+            controller.PresentViewController(messageController, true, null);
+
+            return tcs.Task;
+        }
+    }
+}
--- a/Caboodle/Sms/Sms.ios.cs
+++ b/Caboodle/Sms/Sms.ios.cs
@@ -1,30 +1,30 @@
-﻿using System;
-using System.Threading.Tasks;
+﻿using System.Threading.Tasks;
 using MessageUI;
 
 namespace Microsoft.Caboodle
 {
     public static partial class Sms
     {
-        public static bool IsComposeSupported
+        static bool IsComposeSupported
             => MFMessageComposeViewController.CanSendText;
 
         public static Task ComposeAsync(SmsMessage message)
         {
-            if (message == null)
-                throw new ArgumentNullException(nameof(message));
-            message.Validate();
+            if (!IsComposeSupported)
+                throw new FeatureNotSupportedException();
 
             // do this first so we can throw as early as possible
             var controller = Platform.GetCurrentViewController();
 
+            // create the controller
+            var messageController = new MFMessageComposeViewController();
+            if (!string.IsNullOrWhiteSpace(message?.Body))
+                messageController.Body = message.Body;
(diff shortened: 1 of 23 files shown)


Caboodle/Sms/Sms.netstandard.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class Sms
+    {
+        public static bool IsComposeSupported
```
Redth: I think we still need to decide if a bool and this naming is what we want to end up with.

mattleibow: Yeah. We may want to have some support level thing: Unsupported, Supported, SupportedWithBackgroundSend or something along those lines. This is more a feature that works when there is a SIM card of some sort.

jamesmontemagno: I think that probably each class should be small enough that a single Enum should represent what it needs...

So we could introduce "SmsBackground" for instance and that is a new API with a new Enum. I think that is almost better to be honest with you.

Caboodle/Sms/Sms.ios.cs
```diff
@@ -0,0 +1,38 @@
...
+            var controller = Platform.GetCurrentViewController();
+
+            // show the controller
+            var tcs = new TaskCompletionSource<bool>();
+            var messageController = new MFMessageComposeViewController
+            {
+                Body = message.Body,
+                Recipients = new[] { message.Recipient }
+            };
+            messageController.Finished += (sender, e) =>
```
jamesmontemagno: Does this get triggered no matter what? even if it is canceled? Is it better ot fire and forget?

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+    }
+
+    public class SmsMessage
+    {
+        public SmsMessage()
+        {
+        }
+
+        public SmsMessage(string body, string recipient)
```
jamesmontemagno: Should there be one with just the body? also what is a recipient? phone number?

Caboodle/Sms/Sms.shared.cs
```diff
@@ -0,0 +1,34 @@
...
+        public string Body { get; set; }
+
+        public string Recipient { get; set; }
+
+        internal void Validate()
+        {
+            if (string.IsNullOrWhiteSpace(Body))
+                throw new ArgumentException("SMS body must not be empty.", nameof(Body));
+
+            if (string.IsNullOrWhiteSpace(Recipient))
```
jamesmontemagno: I think it can be empty. I think that is fine.

DeviceTests/Caboodle.DeviceTests.Shared/Utils.cs
```diff
@@ -1,10 +1,14 @@
 ﻿using System;
 using System.Threading.Tasks;
+using Microsoft.Caboodle;
 
 namespace Caboodle.DeviceTests
 {
     public class Utils
     {
+        public static bool IsiOSSimulator
```
jamesmontemagno: Could this be a method that we pass in the device platform?

Caboodle/Sms/Sms.ios.cs
mattleibow: This appears to be the case. The docs do say we have to explicitly hide the controller in this event and we are provided a result (success, canceled, failed)

Caboodle/Sms/Sms.shared.cs
mattleibow: we could have overloads for each combo, or we could just let the devs use the properties.
The recipient... each platform just takes a string, so we could leave it as is and hope for the best :)

DeviceTests/Caboodle.DeviceTests.Shared/Utils.cs
mattleibow: This was more just for the tests, as the iOS sim was quite different

jamesmontemagno: ahhh gotcha <3 I see it is devicetests cool.

jamesmontemagno:
My only feedback here is to allow empty recipients in general. Also... does it hurt if the message is blank? Maybe they just want to launch the sms app?

Redth:
1. Let's have a `ComposeAsync()` (but please double check we can open sms compose on all platforms with no recipient/body.
2. Don't require recipient/body.
3. Make `IsComposeSupported` internal for now, and check it when we execute `ComposeAsync` and throw the same exception across platforms if it's unsupported at runtime.

mattleibow:
I think this is done.
I haven't really been able to properly test the Android part - I think my old Android device is a bit messed - I don't appear to have a messaging app... Or a SIM card for it. I could reset and use my SIM, but there are others that are Android users, I think? Is it possible for someone with a real device to test?

Issue #22, opened 2018-02-24
Share API


jamesmontemagno:
Ability to Share text via a share sheet.

Things to consider:

- Title
- Chooser Title
- Message
- Link
- Activities to Exclude on iOS?

Each of these work a bit different so we should look at this.

mattleibow:
I remember a discussion with someone on this. It appears that the activities on iOS are really things that are grouped by what the content type is. For example, when sharing text, we would want Facebook or Mail, but not Vimeo or Camera Roll. The same might be said for a picture, we want Facebook and Camera Roll, but not Vimeo or iBooks.

We might be able to hide the concept of iOS's activities behind a "content type" indicator.

PR #97, opened 2018-03-18
GH-22: Initial Support for Data Transfer aka Share

jamesmontemagno:
### Description of Change ###

Initial support for Share API aka Data Transfer. Leave API open to transferring of other types of data in the future.

### Bugs Fixed ###

- Related to issue #22

### API Changes ###

Added: 
 
- Task ShowShareUI(ShareTextRequest request)
- Task ShowShareUI(string text)
- Task ShowShareUI(string text, string title)

### PR Checklist ###

### DIFF
--- a/Caboodle/DataTransfer/DataTransfer.android.cs
+++ b/Caboodle/DataTransfer/DataTransfer.android.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Caboodle
 {
     public static partial class DataTransfer
     {
-        public static Task ShowShareUI(ShareTextRequest request)
+        public static Task RequestAsync(ShareTextRequest request)
         {
             var items = new List<string>();
             if (!string.IsNullOrWhiteSpace(request.Text))
--- a/Caboodle/DataTransfer/DataTransfer.android.cs
+++ b/Caboodle/DataTransfer/DataTransfer.android.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Android.Content;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class DataTransfer
+    {
+        public static Task ShowShareUI(ShareTextRequest request)
+        {
+            var items = new List<string>();
+            if (!string.IsNullOrWhiteSpace(request.Text))
+            {
+                items.Add(request.Text);
+            }
+
+            if (!string.IsNullOrWhiteSpace(request.Uri))
+            {
+                items.Add(request.Uri);
+            }
+
+            var intent = new Intent(Intent.ActionSend);
+            intent.SetType("text/plain");
+            intent.PutExtra(Intent.ExtraText, string.Join(Environment.NewLine, items));
+
+            if (!string.IsNullOrWhiteSpace(request.Subject))
+            {
+                intent.PutExtra(Intent.ExtraSubject, request.Subject);
+            }
+
+            var chooserIntent = Intent.CreateChooser(intent, request.Title ?? string.Empty);
+            chooserIntent.SetFlags(ActivityFlags.ClearTop);
+            chooserIntent.SetFlags(ActivityFlags.NewTask);
+            Platform.CurrentContext.StartActivity(chooserIntent);
+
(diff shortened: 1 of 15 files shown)


Caboodle/DataTransfer/DataTransfer.android.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Android.Content;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class DataTransfer
+    {
+        public static Task ShowShareUI(ShareTextRequest request)
```
mattleibow: Should this be `UI` or `Ui`?

Samples/Caboodle.Samples/Caboodle.Samples.csproj
```diff
@@ -21,6 +21,12 @@
     <ProjectReference Include="..\..\Caboodle\Caboodle.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="View\DataTransferPage.xaml">
+      <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
+    </EmbeddedResource>
+  </ItemGroup>
```
mattleibow: This should not be here

Caboodle/DataTransfer/DataTransfer.uwp.cs
```diff
@@ -0,0 +1,46 @@
...
+using Windows.ApplicationModel.DataTransfer;
+using Windows.Foundation;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class DataTransfer
+    {
+        static ShareTextRequest shareRequest;
+
+        public static Task ShowShareUI(ShareTextRequest request)
```
mattleibow: I think the method may be better named `RequestRansfer` rather than show share ui. Just my preference as I don't like the `UI` casing as well as we are talking transfer in the class name, and then switch to share in the method.

Caboodle/DataTransfer/DataTransfer.uwp.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Windows.ApplicationModel.DataTransfer;
+using Windows.Foundation;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class DataTransfer
+    {
+        static ShareTextRequest shareRequest;
```
mattleibow: Is this safe? Maybe rather than using a static field, just use an inline delegate?

Caboodle/DataTransfer/DataTransfer.android.cs
jamesmontemagno: I followed what UWP did... I think UI is correct.

jamesmontemagno: Because it is 2 words is my thought here.

Caboodle/DataTransfer/DataTransfer.uwp.cs
jamesmontemagno: I think just "Request" would be good..

DataTransfer.Request("text")

jamesmontemagno: and should be REquestAsync

Caboodle/DataTransfer/DataTransfer.uwp.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.Caboodle
+{
+    public static partial class DataTransfer
+    {
+        public static Task RequestAsync(ShareTextRequest request)
+        {
+            var dataTransferManager = DataTransferManager.GetForCurrentView();
+
+            dataTransferManager.DataRequested += new TypedEventHandler<DataTransferManager, DataRequestedEventArgs>(ShareTextHandler);
```
Redth: What's the lifecycle of the event handler here?

jamesmontemagno: ahhh good cactch, need to remove it after it comes back.

Issue #66, opened 2018-03-09
Storage Secure API


moljac:
Storage Secure (Secured) API

| Platform                           | Implementation                                         |
|------------------------|------------------------------------------|
| Android                          | KeyStore                                                      |
| iOS                                  | KeyChain                                                     |
| UWP                               |  StorageFile + DataProtectionProvider      |

1.   `Task<string> LoadAsync(string filename)`
2.  `Task SaveAsync(string filename, string content)`

moljac:
Branch:
https://github.com/xamarin/Caboodle/tree/master_based_storagesecure

mattleibow:
A diff: https://github.com/xamarin/Caboodle/compare/master_based_storagesecure

PR #102, opened 2018-03-21
GH-66: SecureStorage Implementation

Redth:
### Description of Change ###

Adds Secure Storage implementation with simple `GetAsync(string key)` and `SetAsync(string key, string data)` methods.

#### iOS: ####
 - KeyChain

#### Android: ####
- KeyStore to store encryption keys, and a subfolder of FilesDir to store encrypted data itself
- On < API 23 a random AES symmetric key is generated, and encrypted with an RSA asymmetric key and saved to disk since KeyStore doesn't support symmetric key storage on these api levels.

#### UWP: ####
 - DataProtectionProvider is used to encrypt data stored in files in a subfolder of ApplicationData.Current.LocalFolder

### API Changes ###

Added: 

- `SecureStorage.GetAsync(string key)`
- `SecureStorage.SetAsync(string key, string data)`

### PR Checklist ###


CodeStyles.targets
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Caboodle.ruleset</CodeAnalysisRuleSet>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
jamesmontemagno: Did you mean to do this?

Samples/Caboodle.Samples/View/StorageSecurePage.xaml.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace Caboodle.Samples.View
+{
+    [XamlCompilation(XamlCompilationOptions.Compile)]
```
jamesmontemagno: Can remove this and cleanup usings

Caboodle/SecureStorage/SecureStorage.ios.cs
```diff
@@ -0,0 +1,97 @@
...
+            var kc = new KeyChain();
+            kc.SetValueForKey(data, key);
+
+            return Task.CompletedTask;
+        }
+    }
+
+    class KeyChain
+    {
+        private static SecRecord ExistingRecordForKey(string key)
```
jamesmontemagno: could remove private from here...

Caboodle/SecureStorage/SecureStorage.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+#endif
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+    public static partial class SecureStorage
+    {
+        // md5(Caboodle.SecureStorage)
+        const string localDirName = "d2f74ffc823f1d9f015ee8edf1463498";
```
Redth: On most platforms we'll need to store data in files.  How do we feel about this?

Caboodle/SecureStorage/SecureStorage.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public static Task<string> GetAsync(string key)
+            => PlatformGetAsync(Md5Hash(key));
```
Redth: In order to get a valid filename I'm just hashing the given key (so it doesn't matter what the key contains).  Thoughts?

mattleibow:
Updating the PR with results of various discussions:
 - writing to files (on UWP/Android) may result in them being deleted if the user does not know which ones are theirs and which are ours
 - It might be better/consistent if all platforms had an approach of storing this in the preferences as a key/value pair

Caboodle/Caboodle.csproj
```diff
@@ -57,6 +57,9 @@
   <ItemGroup Condition=" $(TargetFramework.StartsWith('Xamarin.iOS')) ">
     <Compile Include="**\*.ios.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="AssemblyInfo.shared.cs" />
```
jamesmontemagno: Can remove this. I hate that it keeps adding these things in.

Samples/Caboodle.Samples/View/StorageSecurePage.xaml
```diff
@@ -0,0 +1,26 @@
...
+             >
+    <ContentPage.Content>
+
+        <StackLayout>
+            <Label Text="Quickly and easily securely store data." FontAttributes="Bold" Margin="12" />
+
+            <ScrollView>
+                <StackLayout Padding="12,0,12,12" Spacing="6">
+                    <Label Text="Filename where to save:" />
+                    <Entry Placeholder="Filename..." Text="{Binding FileName}" />
```
mattleibow: Is this a filename or just a key?

DeviceTests/Caboodle.DeviceTests.Shared/SecureStorage_Tests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [InlineData("funny*&$%@!._/\\chars", "data3")]
+        public async Task Saves_And_Loads(string key, string data)
+        {
+            await SecureStorage.SetAsync(key, data);
+
+            var c = await SecureStorage.GetAsync(key);
+
+            Assert.Equal(data, c);
+        }
+#endif
```
mattleibow: Maybe rather than duplicating the whole block, just wrap the `SecureStorage.AlwaysUseAsymmetricKeyStorage = emulatePreApi23; ` line. It doesn't matter that the code may run twice on the other platforms.

mattleibow: Maybe rather than duplicating the whole block, just wrap the `SecureStorage.AlwaysUseAsymmetricKeyStorage = emulatePreApi23; ` line. It doesn't matter that the code may run twice on the other platforms.

Samples/Caboodle.Samples/View/StorageSecurePage.xaml
mattleibow: Is this a filename or just a key?

DeviceTests/Caboodle.DeviceTests.Shared/SecureStorage_Tests.cs
```diff
@@ -30,6 +30,12 @@ public class SecureStorage_Tests
         [InlineData("funny*&$%@!._/\\chars", "data3")]
         public async Task Saves_And_Loads(string key, string data)
         {
+#if __IOS__
+            // Don't run this test on Simulator on iOS
+            if (ObjCRuntime.Runtime.Arch != ObjCRuntime.Arch.DEVICE)
+                return;
+#endif
+
```
mattleibow: Rather use `Platform.DeviceType` instead of this code.

mattleibow: I think I also created a `Utils.IsiOSSimulator` for just this case as well.

DeviceTests/Caboodle.DeviceTests.Shared/SecureStorage_Tests.cs
mattleibow: I think we should just relax that rule and make it a single function. This "unused parameter" will come up a fair bit - rather just leave it as a Message instead of Warning.

Issue #18, opened 2018-02-24
Permissions API's


Redth:
Check and Request permissions.

Internally we need to ensure we have API’s for a couple of things:
- `internal bool EnsurePermissionDeclared (string permission)` checks if *Info.plist* on iOS, or *AndroidManifest.xml*, or *x* on UWP has the permission declared/defined and throws an exception with a meaningful error message including a link to documentation of how to ensure permissions are defined in the appropriate file for the given platform at runtime.

## Permissions
```sharp
public static class Permissions 
{
    public static Task<PermissionStatus> CheckPermissionAsync (string permission);
    public static Task<IDictionary<string, PermissionStatus>> CheckPermissionsAsync (params string[] permissions);

    public static Task<PermissionStatus> RequestPermissionAsync (string permission);
    public static Task<IDictionary<string, PermissionStatus>> RequestPermissionsAsync (params string[] permissions);
}
```

## PermissionStatus (enum)
```sharp
public enum PermissionStatus {
    Unknown,
    Denied,
    Disabled,
    Restricted,
    Granted,
}
```

Redth:
I think we also need some helper properties/static readonly fields that return the correct string permission for a given platform:

This is not how I'd want the actual implementation to be, but to illustrate the idea: 

```csharp
public static class PermissionType
{
#if __ANDROID__
    public static readonly Camera = "android.permission.CAMERA";
#elif __IOS__
    public static readonly Camera = "Camera";
#elif __UWP__
    public static readonly Camera = "Camera";
#endif
}
```

So that the user can call the api's more sanely splat:

```csharp
await Permissions.RequestPermissionAsync(PermissionType.Camera);
```

Redth:
I may be changing my mind on this a bit now.  Looking at iOS and UWP, they both have different code paths depending on which permission has been requested, unlike Android which is just a string to identify the permission.

I'm starting to think an enum was already the best choice here.

PR #117, opened 2018-03-27
GH-18: Permissions Internal implementation

Redth:
### Description of Change ###

Some API's require checking permissions.  This adds some (currently) internal API's for checking permissions and requesting them.

Currently this only supports permission types which we need for other API's (`Battery`, `NetworkAccess1, and `LocationWhileInUse`).  We can add more types as we encounter the need for them.

### API Changes ###

Added: 

```csharp
internal static class Permissions {
    // Throws a relevant PermissionException if permissions aren't declared in:
    //   AndroidManifest.xml / Info.plist / AppxManifest.xml
    internal static Task EnsureDeclaredAsync(PermissionType permission);

    // Checks the permission status
    internal static Task<PermissionStatus> CheckStatusAsync(PermissionType permission);

    // Requests permission
    internal static Task<PermissionStatus> RequestAsync(PermissionType permission);
}
```

### PR Checklist ###

### DIFF
--- a/Caboodle/Permissions/PermissionEnums.shared.cs
+++ b/Caboodle/Permissions/PermissionEnums.shared.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
--- a/Caboodle/Permissions/PermissionEnums.shared.cs
+++ b/Caboodle/Permissions/PermissionEnums.shared.cs
@@ -1,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Caboodle
+{
+    internal enum PermissionStatus
+    {
+        // Denied by user
+        Denied,
+
+        // Feature is disabled on device
+        Disabled,
+
+        // Granted by user
+        Granted,
+
+        // Restricted (only iOS)
+        Restricted,
+
+        // Permission is in an unknown state
+        Unknown
+    }
+
+    internal enum PermissionType
+    {
+        Unknown,
+        Battery,
+        LocationWhenInUse,
+        NetworkState,
+    }
+}
--- a/Caboodle/Permissions/Permissions.android.cs
+++ b/Caboodle/Permissions/Permissions.android.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
--- a/Caboodle/Permissions/Permissions.android.cs
+++ b/Caboodle/Permissions/Permissions.android.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Caboodle
 
         static void PlatformEnsureDeclared(PermissionType permission)
         {
-            var androidPermissions = permission.ToAndroidPermissions();
+            var androidPermissions = permission.ToAndroidPermissions(onlyRuntimePermissions: false);
 
             // No actual android permissions required here, just return
             if (androidPermissions == null || !androidPermissions.Any())
@@ -43,7 +43,8 @@ namespace Microsoft.Caboodle
 
             // If there are no android permissions for the given permission type
             // just return granted since we have none to ask for
-            var androidPermissions = permission.ToAndroidPermissions();
+            var androidPermissions = permission.ToAndroidPermissions(onlyRuntimePermissions: true);
+
             if (androidPermissions == null || !androidPermissions.Any())
                 return Task.FromResult(PermissionStatus.Granted);
 
@@ -99,7 +100,7 @@ namespace Microsoft.Caboodle
             if (!doRequest)
                 return await tcs.Task;
 
-            var androidPermissions = permission.ToAndroidPermissions().ToArray();
+            var androidPermissions = permission.ToAndroidPermissions(onlyRuntimePermissions: true).ToArray();
 
             ActivityCompat.RequestPermissions(Platform.CurrentActivity, androidPermissions, requestCode);
 
@@ -133,19 +134,32 @@ namespace Microsoft.Caboodle
 
     internal static class PermissionTypeExtensions
     {
-        internal static string[] ToAndroidPermissions(this PermissionType permissionType)
+        internal static IEnumerable<string> ToAndroidPermissions(this PermissionType permissionType, bool onlyRuntimePermissions)
         {
+            var permissions = new List<(string permission, bool runtimePermission)>();
+
             switch (permissionType)
             {
                 case PermissionType.Battery:
-                    return new[] { Manifest.Permission.BatteryStats };
+                    permissions.Add((Manifest.Permission.BatteryStats, false));
+                    break;
                 case PermissionType.LocationWhenInUse:
-                    return new[] { Manifest.Permission.AccessFineLocation, Manifest.Permission.AccessCoarseLocation };
+                    permissions.Add((Manifest.Permission.AccessFineLocation, true));
+                    permissions.Add((Manifest.Permission.AccessCoarseLocation, true));
+                    break;
                 case PermissionType.NetworkState:
-                    return new[] { Manifest.Permission.AccessNetworkState };
+                    permissions.Add((Manifest.Permission.AccessNetworkState, false));
+                    break;
             }
 
-            return null;
+            if (onlyRuntimePermissions)
+            {
+                return permissions
+                    .Where(p => p.runtimePermission)
+                    .Select(p => p.permission);
+            }
+
+            return permissions.Select(p => p.permission);
         }
     }
 }
--- a/Caboodle/Permissions/Permissions.android.cs
+++ b/Caboodle/Permissions/Permissions.android.cs
@@ -47,12 +47,12 @@ namespace Microsoft.Caboodle
             if (androidPermissions == null || !androidPermissions.Any())
                 return Task.FromResult(PermissionStatus.Granted);
 
-            var hasApiM = Platform.HasApiLevel(Android.OS.BuildVersionCodes.M);
             var context = Platform.CurrentContext;
+            var targetsMOrHigher = context.ApplicationInfo.TargetSdkVersion >= Android.OS.BuildVersionCodes.M;
 
             foreach (var ap in androidPermissions)
             {
-                if (hasApiM)
+                if (targetsMOrHigher)
                 {
                     if (context.CheckSelfPermission(ap) != Permission.Granted)
                         return Task.FromResult(PermissionStatus.Denied);
--- a/Caboodle/Permissions/Permissions.android.cs
+++ b/Caboodle/Permissions/Permissions.android.cs
@@ -13,8 +13,8 @@ namespace Microsoft.Caboodle
     {
         static readonly object locker = new object();
 
-        static Dictionary<PermissionType, Tuple<int, TaskCompletionSource<PermissionStatus>>> requests =
-            new Dictionary<PermissionType, Tuple<int, TaskCompletionSource<PermissionStatus>>>();
+        static Dictionary<PermissionType, (int requestCode, TaskCompletionSource<PermissionStatus> tcs)> requests =
+            new Dictionary<PermissionType, (int, TaskCompletionSource<PermissionStatus>)>();
 
         static void PlatformEnsureDeclared(PermissionType permission)
         {
@@ -81,7 +81,7 @@ namespace Microsoft.Caboodle
             {
                 if (requests.ContainsKey(permission))
                 {
-                    tcs = requests[permission].Item2;
+                    tcs = requests[permission].tcs;
                     doRequest = false;
                 }
                 else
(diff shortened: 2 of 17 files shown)


Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,102 @@
...
+using Android;
+using Android.Content.PM;
+using Android.Support.V4.App;
+using Android.Support.V4.Content;
+
+namespace Microsoft.Caboodle
+{
+    internal static partial class Permissions
+    {
+        static TaskCompletionSource<PermissionStatus> tcs;
```
Redth: Don 't love this being a static... It really means we can only ever have one request going at a time, which should be the case, but who knows...

I'm thinking maybe a `Dictionary<int, TaskCompletionSource<PermissionStatus>>` is a better idea (where `int` is the permission request code that we keep incrementing with each call).

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            if (androidPermissions == null || !androidPermissions.Any())
+                return Task.FromResult(PermissionStatus.Unknown);
+
+            var hasApiM = Platform.HasApiLevel(Android.OS.BuildVersionCodes.M);
+            var context = Platform.CurrentContext;
+
+            foreach (var ap in androidPermissions)
+            {
+                if (hasApiM)
```
Redth: @jamesmontemagno is this truly necessary? Shouldn't `PermissionChecker` handle this for us regardless of API level?

jamesmontemagno: You actually need to check if it is TARGETING... not running on M+

```
	var targetsMOrHigher = context.ApplicationInfo.TargetSdkVersion >= Android.OS.BuildVersionCodes.M;

			foreach (var name in names)
			{
				if (targetsMOrHigher)
				{
					if (ContextCompat.CheckSelfPermission(context, name) != Android.Content.PM.Permission.Granted)
						return Task.FromResult(PermissionStatus.Denied);
				}
				else
				{
					if(PermissionChecker.CheckSelfPermission(context, name)  != PermissionChecker.PermissionGranted)
						return Task.FromResult(PermissionStatus.Denied);
				}
            }
```

Caboodle/Permissions/Permissions.android.cs
jamesmontemagno: yeah, that seems like a better idea

Caboodle/Permissions/PermissionEnums.shared.cs
```diff
@@ -0,0 +1,32 @@
...
+        // Restricted (only iOS)
+        Restricted,
+
+        // Permission is in an unknown state
+        Unknown
+    }
+
+    internal enum PermissionType
+    {
+        Unknown,
```
jamesmontemagno: Should we have an unknown?

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -11,7 +11,10 @@ namespace Microsoft.Caboodle
 {
     internal static partial class Permissions
     {
-        static TaskCompletionSource<PermissionStatus> tcs;
+        static readonly object locker = new object();
+
+        static Dictionary<PermissionType, Tuple<int, TaskCompletionSource<PermissionStatus>>> requests =
```
jamesmontemagno: Should we be using the new Tuple Types?

Caboodle/Permissions/Permissions.android.cs
Redth: Oops, missed that... However my question remains, why do we need to change code paths when support libraries should do this dirty work for us already?

jamesmontemagno: It doesn't :( trust me, it sucks.

Caboodle/Permissions/Permissions.android.cs
Redth: This requires a nuget package right? but just for build time stuff?

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,151 @@
...
+                    tcs = requests[permission].tcs;
+                    doRequest = false;
+                }
+                else
+                {
+                    tcs = new TaskCompletionSource<PermissionStatus>();
+
+                    // Get new request code and wrap it around for next use if it's going to reach max
+                    if (++requestCode >= int.MaxValue)
+                        requestCode = 1;
```
mattleibow: What is the scope of `requestCode`? It appears to be created in this method, so it will always be `0` - see line 79.

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,151 @@
...
+using Android.Support.V4.Content;
+
+namespace Microsoft.Caboodle
+{
+    internal static partial class Permissions
+    {
+        static readonly object locker = new object();
+
+        static Dictionary<PermissionType, (int requestCode, TaskCompletionSource<PermissionStatus> tcs)> requests =
+            new Dictionary<PermissionType, (int, TaskCompletionSource<PermissionStatus>)>();
```
mattleibow: Maybe make this a `using PermissionsDictionary = Sys.Coll.Gen.Dictionary<PermissionType, (int requestCode, TaskCompletionSource<PermissionStatus> tcs)>;` to make it easier to read as this is a very long type.
Maybe?

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,151 @@
...
+            // No actual android permissions required here, just return
+            if (androidPermissions == null || !androidPermissions.Any())
+                return;
+
+            var context = Platform.CurrentContext;
+
+            foreach (var ap in androidPermissions)
+            {
+                var packageInfo = context.PackageManager.GetPackageInfo(context.PackageName, PackageInfoFlags.Permissions);
+                var requestedPermissions = packageInfo?.RequestedPermissions;
```
mattleibow: Maybe move this out of the loop?

Caboodle/Permissions/Permissions.android.cs
```diff
@@ -0,0 +1,165 @@
...
+            foreach (var ap in androidPermissions)
+            {
+                if (targetsMOrHigher)
+                {
+                    if (context.CheckSelfPermission(ap) != Permission.Granted)
+                        return Task.FromResult(PermissionStatus.Denied);
+                }
+                else
+                {
+                    if (PermissionChecker.CheckSelfPermission(context, ap) != PermissionChecker.PermissionGranted)
```
mattleibow: Is there a benefit to using the native method vs the support method? Won't the support fall back to the native method if it exists?

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                case PermissionType.LocationWhenInUse:
+                    EnsureLocationPermissionDeclared(permission);
+                    break;
+            }
+        }
+
+        static Task<PermissionStatus> PlatformCheckStatusAsync(PermissionType permission)
+        {
+            EnsureLocationPermissionDeclared(permission);
```
mattleibow: Should this be in the switch?

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,122 @@
...
+                case PermissionType.LocationWhenInUse:
+                    return Task.FromResult(GetLocationStatus());
+            }
+
+            return Task.FromResult(PermissionStatus.Granted);
+        }
+
+        static Task<PermissionStatus> PlatformRequestAsync(PermissionType permission)
+        {
+            EnsureLocationPermissionDeclared(permission);
```
mattleibow: Should this be in the switch?

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,122 @@
...
+            switch (permission)
+            {
+                case PermissionType.LocationWhenInUse:
+                    return RequestLocationAsync();
+            }
+
+            return Task.FromResult(PermissionStatus.Granted);
+        }
+
+        static void EnsureLocationPermissionDeclared(PermissionType permission)
```
mattleibow: I think this is mis-named... This is not just for location.

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            if (!UIDevice.CurrentDevice.CheckSystemVersion(8, 0))
+                return Task.FromResult(PermissionStatus.Unknown);
+
+            var locationManager = new CLLocationManager();
+
+            EventHandler<CLAuthorizationChangedEventArgs> authCallback = null;
+            var tcs = new TaskCompletionSource<PermissionStatus>();
+
+            authCallback = (sender, e) =>
```
mattleibow: Is it possible to use an inner function here so as to avoid this scary set-to-null-and-then-create-delegate?

mattleibow: Also, it will not require a delegate to be created each time...

Caboodle/Permissions/Permissions.android.cs
Redth: @jamesmontemagno says there are dragons here and to do this.

Caboodle/Permissions/Permissions.android.cs
Redth: 🤦‍♂️

Caboodle/Permissions/PermissionEnums.shared.cs
Redth: There's a few spots in iOS and one in UWP where we return this....

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    internal static partial class Permissions
+    {
+        static void PlatformEnsureDeclared(PermissionType permission)
+        {
+            var info = NSBundle.MainBundle.InfoDictionary;
+
+            if (permission == PermissionType.LocationWhenInUse)
+            {
+                if (!info.ContainsKey(new NSString("NSLocationWhenInUseUsageDescription")))
```
mattleibow: Is there a constant for this at all?

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,111 @@
...
+                    throw new PermissionException("On iOS 8.0 and higher you must set either `NSLocationWhenInUseUsageDescription` or `NSLocationAlwaysUsageDescription` in your Info.plist file to enable Authorization Requests for Location updates!");
```
mattleibow: This exception talks about iOS 8.0 - what happens on 7.0? Should there be another check somewhere or condition?

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -0,0 +1,111 @@
...
+        static Task<PermissionStatus> RequestLocationAsync()
+        {
+            if (!UIDevice.CurrentDevice.CheckSystemVersion(8, 0))
+                return Task.FromResult(PermissionStatus.Unknown);
+
+            var locationManager = new CLLocationManager();
+
+            var tcs = new TaskCompletionSource<PermissionStatus>();
+
+            void AuthCallback(object sender, CLAuthorizationChangedEventArgs e)
```
mattleibow: These nested/inner functions ❤️  and 😈

DeviceTests/Caboodle.DeviceTests.Shared/Permissions_Tests.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Caboodle;
+using Xunit;
+
+#if __ANDROID__
+[assembly: Android.App.UsesPermission(Android.Manifest.Permission.BatteryStats)]
+#endif
```
mattleibow: Does this have to be here? Isn't it in the manifest?

Caboodle/Permissions/Permissions.ios.cs
Redth: Fixed this, it shouldn't bother checking info.plist in < 8.0

DeviceTests/Caboodle.DeviceTests.Shared/Permissions_Tests.cs
Redth: This makes sure it is in the manifest... I like doing it attributes especially for a unit test which requires it :)

mattleibow: No, I mean isn't it already in the manifest? Maybe we should remove it there - or maybe it is fine for both locations?

Redth: You know, it _is_ already there, which I thought it wasn't earlier but I believe that was due to me trying to check for the BATTERY_STATS permission as if it required user input and it was returning false because it didn't.  In any case I think it's fine in both places.

Caboodle/Permissions/Permissions.ios.cs
Redth: I can't find one....

PR #123, opened 2018-03-28
[Proposal] Get the Latest Store Version Number

mattleibow:
Just a diff of what could potentially be a feature.

### DIFF
--- a/Caboodle/AppInfo/AppInfo.android.cs
+++ b/Caboodle/AppInfo/AppInfo.android.cs
@@ -1,14 +1,10 @@
 ﻿using System;
 using System.Globalization;
+using System.Net.Http;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
-using Android.App;
-using Android.Content;
 using Android.Content.PM;
-using Android.Content.Res;
-using Android.OS;
-using Android.Provider;
-using Android.Runtime;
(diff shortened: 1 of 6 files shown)


Caboodle/AppInfo/AppInfo.android.cs
```diff
@@ -43,5 +39,24 @@ static string GetBuild()
                 return info.VersionCode.ToString(CultureInfo.InvariantCulture);
             }
         }
+
+        static async Task<string> PlatformGetLatestVersionStringAsync()
+        {
+            var version = string.Empty;
+            var url = $"https://play.google.com/store/apps/details?id={GetPackageName()}";
```
mattleibow: Android is a bit hacky as there is no official API, so we just ask the website. This may return "Multiple Versions" for apps that have multiple APKs

jamesmontemagno:
I fear this API a bit to be honest with you. Only because Google doesn't offer an official API. Additionally, this is then locking it into google play. What about amazon or samsung's app store? I just don't think it is a good idea.

mattleibow:
Closing this PR as it is not something that can be realistically supported.

jamesmontemagno:
I also thought about my "store review" plugin, but on Android it is tricky as it requires google play and makes assumptions :(

PR #124, opened 2018-03-28
[Proposal] Add an API to Track App Versions

mattleibow:
This is inspired by https://github.com/colbylwilliams/VersionTrackingPlugin
This way there is no need for complex code, just use the Preferences and AppInfo

### PR Checklist ###


Caboodle/VersionTracking/VersionTracking.shared.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        const string versionTrailKey = "VersionTracking.Trail";
+        const string versionsKey = "VersionTracking.Versions";
+        const string buildsKey = "VersionTracking.Builds";
+
+        static readonly Preferences preferences = new Preferences($"{AppInfo.PackageName}.caboodle");
+
+        static Dictionary<string, List<string>> versionTrail;
+
+        public static void Track()
```
jamesmontemagno: Should this really be required? Couldn't we just make a static constructor that gets called when anyone accesses a method?

I could see people calling IsFirstLaunchEver all the time...

jamesmontemagno:
Can you add to readme :) added PR requirements

Issue #7, opened 2018-02-24
Geolocation API's


Redth:
The Geolocation API is meant to be extremely simple at this point.

Geocoding API's: https://github.com/xamarin/Caboodle/issues/28

```csharp
static class Geolocation {
    // Android and iOS have a concept of a last known position that should return very quickly compared to firing up the GPS and waiting for a position.
    // This should return null if there is no known last position.
    // UWP has something interesting about consent-less location:
    // https://docs.microsoft.com/en-us/uwp/api/Windows.Devices.Geolocation.Geolocator#Windows_Devices_Geolocation_Geolocator_AllowFallbackToConsentlessPositions
    static Task<Position> GetLastKnownPositionAsync();

    // We should throw different exceptions depending on the error: For example a DeviceFeatureUnavailableException or something similar if the device has no GPS abilities.
    static Task<Position> GetPositionAsync(GeolocationAccuracy desiredAccuracy, TimeSpan timeout, CancellationToken token = default(CancellationToken));
}
```
```csharp
class Position {
    Position();
    Position(double latitude, double longitude);
    Position(double latitude, double longitude, DateTimeOffset timestamp);
    Position(Position position);

    DateTimeOffset Timestamp { get; set; }
    double Latitude { get; set; }
    double Longitude { get; set; }

    static double CalculateDistance(Position positionStart, Position positionEnd);
    static double CalculateDistance(
        double latitudeStart, double latitudeEnd, 
        double longitudeStart, double longitudeEnd);
}
```

```csharp
enum GeolocationAccuracy {
    Lowest,
    Low,
    Medium,
    High,
    Best
}
```

The accuracy enum values should map to the following table of platform specific accuracy choices:

|            | **iOS**          | Expected | **Android**                   | Expected | **UWP** | Expected  |
| ---------- | ---------------- | -------- | ----------------------------- | -------- | ------- | --------- |
| **Lowest** | ThreeKilometers  | 3000     | ACCURACY_LOW, POWER_LOW       | 500      | 3000    | 1000-5000 |
| **Low**    | Kilometer        | 1000     | ACCURACY_LOW, POWER_MEDIUM    | 500      | 1000    | 300-3000  |
| **Medium** | HundredMeters    | 100      | ACCURACY_MEDIUM, POWER_MEDIUM | 100-500  | 100     | 30-500    |
| **High**   | NearestTenMeters | 10       | ACCURACY_HIGH, POWER_MEDIUM   | 0-100    | High    | <=10      |
[...]

PR #132, opened 2018-03-29
GH-7: Geolocation API's

Redth:
### Description of Change ###

Adds Geolocation API's

### API Changes ###

```csharp
public static Task<Location> GetLastKnownLocationAsync();
public static Task<Location> GetLocationAsync();
public static Task<Location> GetLocationAsync(GeolocationRequest request);
public static Task<Location> GetLocationAsync(GeolocationRequest request, CancellationToken cancelToken);
```

```csharp
public enum GeolocationAccuracy
{
    // iOS:     ThreeKilometers         (3000m)
    // Android: ACCURACY_LOW, POWER_LOW (500m)
    // UWP:     3000                    (1000-5000m)
    Lowest,

    // iOS:     Kilometer               (1000m)
    // Android: ACCURACY_LOW, POWER_MED (500m)
    // UWP:     1000                    (300-3000m)
    Low,

    // iOS:     HundredMeters           (100m)
    // Android: ACCURACY_MED, POWER_MED (100-500m)
    // UWP:     100                     (30-500m)
    Medium,

    // iOS:     NearestTenMeters        (10m)
    // Android: ACCURACY_HI, POWER_MED  (0-100m)
    // UWP:     High                    (<=10m)
    High,

    // iOS:     Best                    (0m)
    // Android: ACCURACY_HI, POWER_HI   (0-100m)
    // UWP:     High                    (<=10m)
    Best
}
```

```csharp
public class GeolocationRequest
{
    public GeolocationRequest();
    public GeolocationRequest(GeolocationAccuracy accuracy);
    public GeolocationRequest(GeolocationAccuracy accuracy, TimeSpan timeout);
    
    public TimeSpan? Timeout { get; set; }
    public GeolocationAccuracy DesiredAccuracy { get; set; }
}
```

### PR Checklist ###


Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -9,6 +9,10 @@ namespace Microsoft.Caboodle
 {
     internal static partial class Permissions
     {
+        static readonly object locationLocker = new object();
+        static CLLocationManager locationManager;
+        static TaskCompletionSource<PermissionStatus> locationTcs;
```
mattleibow: This needs a set of eyes here 👀

Caboodle/Types/Shared/Utils.shared.cs
```diff
@@ -31,5 +32,18 @@ internal static string Md5Hash(string input)
             return hash.ToString();
 #endif
         }
+
+        internal static CancellationToken TimeoutToken(CancellationToken cancellationToken, TimeSpan timeout)
```
mattleibow: A util function that will auto-cancel a token after a timeout

Caboodle/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,113 @@
...
+                Accuracy = location.HorizontalAccuracy,
+                TimestampUtc = ToDate(location.Timestamp)
+            };
+        }
+
+        static async Task<Location> PlatformLocationAsync(GeolocationRequest request, CancellationToken cancellationToken)
+        {
+            await Permissions.RequireAsync(PermissionType.LocationWhenInUse).ConfigureAwait(false);
+
+            var manager = new CLLocationManager();
```
mattleibow: This location manager is different to the one used in permissions. Need to review if this gets GC'ed unexpectedly.

Caboodle/Permissions/Permissions.ios.cs
mattleibow: I am locking and making sure that there is only a single location manager for all permissions, but is this necessary?

mattleibow: I also saw a constructor for `TaskCompletionSource` that takes `object state`... I have never used that, but maybe there is some way to use that to keep the listener alive... will investigate, but comments will help.

Caboodle/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            var tcs = new TaskCompletionSource<CLLocation>(manager);
```
mattleibow: Keep a reference to the manager for the duration of the task... this needs 👀

Caboodle/Permissions/Permissions.ios.cs
```diff
@@ -91,25 +95,23 @@ static Task<PermissionStatus> RequestLocationAsync()
             if (!UIDevice.CurrentDevice.CheckSystemVersion(8, 0))
                 return Task.FromResult(PermissionStatus.Unknown);
 
-            var locationManager = new CLLocationManager();
+            var manager = new CLLocationManager();
+
+            var tcs = new TaskCompletionSource<PermissionStatus>(manager);
```
mattleibow: Keep a reference to the manager for the duration of the task... this needs 👀

mattleibow:
With regards to a fail-fast case, this is a bit harder to implement. To be realistic, almost all devices nowadays have a GPS or some basic location system. In most cases there are 2 alternatives: no permission and disabled.

UWP is fun with a more async model. The locator has to init and then returns a status. Also, when the GPS is off, the `RequestAccessAsync` returns `Denied`. In addition, we have to first get permission and then request a location in order to get the status.

iOS appears to always have a GPS, and they have a property for detecting whether it enabled or not.

Android appears to be like UWP in that you have to get permission in order to see what GPS/location devices are available.

mattleibow:
Just looking at the implementations, we have to handle the cases of the location devices being disabled or permissions revoked during a request.

Redth:
Honestly this is more reason to consider play services location since it helps deal with getting the user to enable location. Not to mention it's a much better API to deal with. I'm not excited about the dependency it would add and we would still need to fall back on the old api but something to consider!

jamesmontemagno:
I have a version 5 of the geolocator plugin, it does help a bit on Android that is sure, but yeah some tricky parts with google play as a dependency :( 

What I did on Android was detect if it had any providers available to determine if it was active or not. The gotcha is that you need to have permissions set to query providers :)


iOS is easy peasy:
```
 public bool IsGeolocationEnabled
        {
            get
            {         
                var status = CLLocationManager.Status;
                return CLLocationManager.LocationServicesEnabled;
            }
        }
```

UWP is a bundle of terrible to be honest with you see here: https://github.com/jamesmontemagno/GeolocatorPlugin/issues/144

jamesmontemagno:
I am wondering if it is best just to assume always on... make the request and throw the exception if it fails... we are exception first.. the question is does that work properly.

On iOS/Android we can check permissions first... UWP is a pain in the butt a bit.

Caboodle/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Microsoft.Caboodle
+{
+    public static partial class Geolocation
+    {
+        internal static bool IsSupported
+            => CLLocationManager.LocationServicesEnabled;
+
+        static async Task<Location> PlatformLastKnownLocationAsync()
+        {
+            await Permissions.RequireAsync(PermissionType.LocationWhenInUse).ConfigureAwait(false);
```
jamesmontemagno: Can we remove the configureawait for now throughout. I am still worried about that stuff.

Caboodle/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            var manager = new CLLocationManager();
+            var location = manager.Location;
+
+            return new Location
```
jamesmontemagno: Can you make this an extension method to convert like i do for geocoding. It may already be there actually.

Caboodle/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,113 @@
...
+                manager.StopUpdatingLocation();
+                tcs.TrySetResult(null);
+            }
+        }
+
+        static DateTimeOffset ToDate(NSDate timestamp)
+        {
+            try
+            {
+                return new DateTimeOffset((DateTime)timestamp);
```
jamesmontemagno: We should put these in platform helpers

Caboodle/Geolocation/Geolocation.uwp.cs
```diff
@@ -0,0 +1,60 @@
...
+                DesiredAccuracy = PositionAccuracy.Default,
+            };
+            geolocator.AllowFallbackToConsentlessPositions();
+
+            var location = await geolocator.GetGeopositionAsync().AsTask().ConfigureAwait(false);
+
+            if (location?.Coordinate == null)
+                return null;
+
+            return new Location
```
jamesmontemagno: Same here for extension methods

Caboodle/Platform/Platform.android.cs
```diff
@@ -84,6 +85,9 @@ public static void BeginInvokeOnMainThread(Action action)
 
         internal static ClipboardManager ClipboardManager =>
             Application.Context.GetSystemService(Context.ClipboardService) as ClipboardManager;
+
+        internal static LocationManager LocationManager =>
+            (LocationManager)Application.Context.GetSystemService(Context.LocationService);
```
jamesmontemagno: can we use "as" for null checks along the way.

Samples/Caboodle.Samples.iOS/Info.plist
```diff
@@ -53,5 +53,7 @@
 	<string>Caboodle.Samples</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
+	<key>NSLocationWhenInUseUsageDescription</key>
+    <string>Access to your location is required for cool things to happen!</string>
```
jamesmontemagno: tab this puppy over

Redth:
Yeah I say we just check for hardware and not try and query enabled status here. I think it's not expected to have a UI asking for permission when checking if it's supported. Also we don't really expose the is supported yet anyway so there really is no code path that wouldn't check for permission anyway. 

Tl;dr let's check for hardware only.

DeviceTests/DeviceTests.Shared/Geolocation_Tests.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Xamarin.Essentials;
+using Xunit;
+
+namespace DeviceTests
+{
+    public class Geolocation_Tests
+    {
+        /*
```
mattleibow: The test are commented out as all platforms require user approval - maybe not the old Androids.

mattleibow:
I went with option 3.

Xamarin.Essentials/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            await Permissions.RequireAsync(PermissionType.LocationWhenInUse);
+
+            var tcs = new TaskCompletionSource<CLLocation>();
+
+            // the location manager requires an active run loop
+            // so just use the main loop
+            Platform.BeginInvokeOnMainThread(() =>
+            {
+                var manager = new CLLocationManager();
```
mattleibow: We create and start the location manger on the main thread:
 - because we need an active run loop
 - it is non-blocking
 - it does not affect the actual call threading system

Xamarin.Essentials/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,95 @@
...
+
+                void Cancel()
+                {
+                    manager.StopUpdatingLocation();
+                    tcs.TrySetResult(null);
+                }
+            });
+
+            // still wait and return on the background thread
+            var clLocation = await tcs.Task;
```
mattleibow: We await the location manager:
 - we still block the current thread until the main thread gets back to us
 - we continue on the background thread

mattleibow:
build

DeviceTests/DeviceTests.Android/Properties/AndroidManifest.xml
```diff
@@ -1,7 +1,15 @@
...
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0.1.0" package="com.xamarin.essentials.devicetests" android:installLocation="auto">
 	<uses-sdk android:minSdkVersion="19" android:targetSdkVersion="26" />
+	<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+	<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 	<uses-permission android:name="android.permission.BATTERY_STATS" />
+	<uses-permission android:name="android.permission.CAMERA" />
+	<uses-permission android:name="android.permission.FLASHLIGHT" />
+	<uses-permission android:name="android.permission.INTERNET" />
+	<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```
jamesmontemagno: Do we need Read/Write for anything?

Xamarin.Essentials/Geolocation/Geolocation.android.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                RemoveUpdates();
+                tcs.TrySetResult(null);
+            }
+
+            void RemoveUpdates()
+            {
+                try
+                {
+                    locationManager.RemoveUpdates(listener);
```
jamesmontemagno: Should we really ignore this?

Xamarin.Essentials/Geolocation/Geolocation.android.cs
```diff
@@ -0,0 +1,200 @@
...
+                    criteria.PowerRequirement = Power.High;
+                    break;
+                case GeolocationAccuracy.Best:
+                    criteria.Accuracy = Accuracy.Fine;
+                    criteria.HorizontalAccuracy = Accuracy.Fine;
+                    criteria.PowerRequirement = Power.High;
+                    break;
+            }
+
+            return locationManager.GetBestProvider(criteria, true) ?? locationManager.GetProviders(true).First();
```
jamesmontemagno: Should this be FirstorDefault?

Xamarin.Essentials/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static partial class Geolocation
+    {
+        internal static bool IsSupported
+            => CLLocationManager.LocationServicesEnabled;
+
+        static async Task<Location> PlatformLastKnownLocationAsync()
+        {
+            await Permissions.RequireAsync(PermissionType.LocationWhenInUse);
+
+            var manager = new CLLocationManager();
```
jamesmontemagno: Should we have this on the main thread or doesn't matter??

Xamarin.Essentials/Geolocation/Geolocation.ios.cs
```diff
@@ -0,0 +1,92 @@
...
+            public Action<CLLocation> LocationHandler { get; set; }
+
+            public override void LocationsUpdated(CLLocationManager manager, CLLocation[] locations)
+            {
+                if (wasRaised)
+                    return;
+
+                wasRaised = true;
+
+                var location = locations.LastOrDefault();
```
jamesmontemagno: Are we assuming these are ordered by priority? Accuracy?

mattleibow: According to the docs:
> The objects in the array are organized in the order in which they occurred. Therefore, the most recent location update is at the end of the array. 
> https://developer.apple.com/documentation/corelocation/cllocationmanagerdelegate/1423615-locationmanager

Xamarin.Essentials/Geolocation/Geolocation.ios.cs
mattleibow: It does not appear to matter.

Issue #52, opened 2018-03-01
Sensors API Magnetometer


moljac:
Sensors API Magnetometer indicating the devices orientation relative to Earth's magnetic field.

API Level:
* iOS: [CMMotionManager](https://developer.apple.com/documentation/coremotion/cmaccelerometerdata) -> Create 1 Instance only
* Android: Sensor Listeners
* UWP: [Windows.Device.Sensors](https://docs.microsoft.com/en-us/uwp/api/windows.devices.sensors.magnetometerreading)

API:
``` csharp
bool IsListening {get;}
void Start(SensorSpeed speed, Action<MagnetometerData> handler);
void Stop();

struct MagnetometerData
{
   bool MagneticFieldX{ get; }
  bool MagneticFieldY{get;}
bool MagneticFieldZ{get;}
}
```

Issue #51, opened 2018-03-01
Sensors API Gyroscope


moljac:
Sensors API Gyroscope to the instantaneous rotation around the device's three primary axes.

API Level:
* iOS: [CMMotionManager](https://developer.apple.com/documentation/coremotion/cmaccelerometerdata) -> Create 1 Instance only
* Android: Sensor Listeners
* UWP: [Windows.Device.Sensors](https://docs.microsoft.com/en-us/windows/uwp/devices-sensors/use-the-gyrometer)

API:
``` csharp
bool IsListening {get;}
void Start(SensorSpeed speed, Action<GyrometerData> handler);
void Stop();

struct GyrometerData
{
   bool AngularVelocityX { get; }
  bool AngularVelocityY {get;}
bool AngularVelocityX {get;}
}
```

Issue #50, opened 2018-03-01
Sensors API Accelerometer


moljac:
Sensors API Accelerometer to indicate the instantaneous acceleration of the device in three dimensional space.

API Level:
* iOS: [CMMotionManager](https://developer.apple.com/documentation/coremotion/cmaccelerometerdata) -> Create 1 Instance only
* Android: Sensor Listeners
* UWP: [Windows.Device.Sensors](https://docs.microsoft.com/en-us/windows/uwp/devices-sensors/use-the-accelerometer)

API:
``` csharp
bool IsListening {get;}
void Start(SensorSpeed speed, Action<AccelerometerData> handler);
void Stop();

struct AccelerometerData
{
   bool AccelerationX { get; }
  bool AccelerationY {get;}
bool AccelerationZ {get;}
}
```

PR #140, opened 2018-03-29
GH-52 Add Magnetometer/Gyro/Accelerometer code and samples

jamesmontemagno:
### Description of Change ###

Add Magnetometer

### Bugs Fixed ###

- Related to issue #52, #51, #50

### API Changes ###

Added: 
 
- bool IsMonitoring { get; set; } 
- void Start(SensorSpeed speed);
- void Stop();

- event MagnetometerChangedEventHandler<MagnetometerData> ReadingChanged;

- class MagnetometerData

### PR Checklist ###


Caboodle/Magnetometer/Magnetometer.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+        internal static void PreMonitorValidation()
+        {
+            if (!IsSupported)
+            {
+                throw new FeatureNotSupportedException();
+            }
+
+            if (IsMonitoring)
+            {
+                throw new InvalidOperationException("Magnetometer is already being monitored. Please stop to start a new session.");
```
mattleibow: I am just looking at the API and I see _why_ this may be a throw, but then why do we use delegates? What was the reason for the choice over events? .NET is very event-y so from my side of UWP/WinForms, the delegate approach is a bit foreign.

The only possibility for delegates is that you can have multiple event streams with different speeds. Is this something that the platforms actually support? Even so, is this something we want to encourage?

What is wrong with:

```csharp
static class Magnetometer {
    static event EventHandler<Data> Updated;
    static void Start(EventSpeed speed);
    static void Stop();
}
```

mattleibow: Another thing that I "feel" with events vs delegates is that delegates are more of a demand: "do this here, or under this condition". Events are more of: "when X happens I want to be notified so I can do this"

jamesmontemagno: Let's discuss today after the call :)

mattleibow: Moving this to a new issue to contain the discussions: https://github.com/xamarin/Caboodle/issues/142

Caboodle/Magnetometer/Magnetometer.ios.cs
```diff
@@ -0,0 +1,53 @@
...
+                    manager.MagnetometerUpdateInterval = NormalInterval;
+                    break;
+                case SensorSpeed.Ui:
+                    manager.MagnetometerUpdateInterval = UiInterval;
+                    break;
+            }
+
+            manager.StartMagnetometerUpdates(NSOperationQueue.CurrentQueue, DataUpdated);
+
+            void DataUpdated(CMMagnetometerData data, NSError error)
```
mattleibow: Maybe move this out into the class to be consistent - unless it has to be here?

Caboodle/Magnetometer/Magnetometer.uwp.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                case SensorSpeed.Fastest:
+                    interval = FastestInterval;
+                    break;
+                case SensorSpeed.Game:
+                    interval = GameInterval;
+                    break;
+            }
+
+            sensor.ReportInterval = sensor.MinimumReportInterval >= interval ? interval : sensor.MinimumReportInterval;
```
mattleibow: Just checking if this is the right way: `min >= desired ? desired : min`

Caboodle/Magnetometer/Magnetometer.uwp.cs
```diff
@@ -0,0 +1,54 @@
...
+            var reading = e.Reading;
+            var data = new MagnetometerData(reading.MagneticFieldX, reading.MagneticFieldY, reading.MagneticFieldZ);
+            OnChanged(data);
+        }
+
+        internal static void PlatformStop()
+        {
+            Platform.BeginInvokeOnMainThread(() =>
+            {
+                DefaultMagnetometer.ReadingChanged -= DataUpdated;
```
mattleibow: I see that this is on the UI thread... is there a specific reason that this is only the detach? Should it also be the attach, or is the main thread for events not necessary at all?

Caboodle/Magnetometer/Magnetometer.uwp.cs
mattleibow: Is min a _high_ number, or a low number?

mattleibow: Ah, I see when I reversed it... maybe flip it around to `min < interval ? min : interval` as it just appears weird.

mattleibow: No wait... that also looks weird.

jamesmontemagno: #resolved

jamesmontemagno: https://docs.microsoft.com/en-us/windows/uwp/devices-sensors/use-the-compass

Caboodle/Magnetometer/Magnetometer.uwp.cs
jamesmontemagno: #resolved

Caboodle/Magnetometer/Magnetometer.ios.cs
jamesmontemagno: #resolved

Caboodle/Accelerometer/Accelerometer.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            if (IsMonitoring)
+            {
+                return;
+            }
+
+            UseSyncContext = sensorSpeed == SensorSpeed.Normal || sensorSpeed == SensorSpeed.Ui;
+            PlatformStart(sensorSpeed);
+            isMonitoring = true;
```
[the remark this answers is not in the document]
jamesmontemagno: #resolved

PR #144, opened 2018-03-30
GH-143 Change compass to event based.

jamesmontemagno:
### Description of Change ###

Makes API for compass event based.

### Bugs Fixed ###

- Related to issue #143

### PR Checklist ###

### DIFF
--- a/Caboodle/Compass/Compass.android.cs
+++ b/Caboodle/Compass/Compass.android.cs
@@ -50,6 +50,10 @@ namespace Microsoft.Caboodle
             Platform.SensorManager.UnregisterListener(listener, magnetometer);
             listener.Dispose();
             listener = null;
+            magnetometer?.Dispose();
+            magnetometer = null;
+            accelerometer?.Dispose();
+            accelerometer = null;
         }
     }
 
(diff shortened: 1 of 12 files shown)


Caboodle/Compass/Compass.android.cs
```diff
@@ -13,10 +11,13 @@ public static partial class Compass
             Platform.SensorManager?.GetDefaultSensor(SensorType.Accelerometer) != null &&
             Platform.SensorManager?.GetDefaultSensor(SensorType.MagneticField) != null;
 
-        internal static void PlatformStart(SensorSpeed sensorSpeed, Action<CompassData> handler)
+        static SensorListener listener;
+        static Sensor magnetometer;
+        static Sensor accelerometer;
```
Redth: Can we dispose/null these out when they are no longer needed?

jamesmontemagno: so they are shared system wide and if you dispose them out then they will effect other listeners from my testing :(

jamesmontemagno: Well actually.... maybe if we only allow one at a time.... let me see here. Will need to see if you can do both compass and others at the same time.

Redth: I guess it could get ugly, but could we track/count the # of listeners we have and dispose if 0?

mattleibow: What if we use a weak reference? This way we leave it up the Androids?

jamesmontemagno: #resolved

Issue #145, opened 2018-04-02
Issues Installing NuGet with Forms Templates


jamesmontemagno:
### Description

Issue comes out of support libraries. Forms has 25.4.0.2 and we had 26.X. NuGet doesn't like this as we are too specific with our NuGets.

### Steps to Reproduce

1.  File New Forms
2.  Add NuGet
3. Unable to install

### Expected Behavior

Smooth Installation

### Actual Behavior

NuGet error due to dependency chain.

PR #146, opened 2018-04-02
GH-145 Adjust Package References for Android

jamesmontemagno:
### Description of Change ###

Match Xamarin.Forms for Android NuGet Libraries.
Can no longer use fancy Android 8 Features, so if def them for now.

### Bugs Fixed ###

- Related to issue #145 

### API Changes ###

Changed:

 - Vibrate will always fall back to classic.

### PR Checklist ###

### DIFF
--- a/Caboodle/Caboodle.csproj
+++ b/Caboodle/Caboodle.csproj
@@ -2,8 +2,8 @@
   <PropertyGroup>
     <!--Work around so the conditions work below-->
     <TargetFrameworks></TargetFrameworks>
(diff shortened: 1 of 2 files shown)


jamesmontemagno:
We thought about both 7.1 and 8.0, but they would still technically both need 25.X support libraries for compat reasons with Forms.

@Redth thoughts on doing both 7.1 and 8.0 for API surface?

Redth:
technically we could compile against 8.0 still and ship a nuget with 7.1 with the 8.0 compiled dll as long as we did runtime checks around any 8.0 api's used.   I'm guessing the `/t:Pack` won't allow this very easily...

I do not think we should ship multi targeted builds for both 7.1 and 8.0 however.

Issue #136, opened 2018-03-29
Audit Linker Safe


jamesmontemagno:
Validate that we are linker safe and that any unused API gets linked out correctly.

PR #147, opened 2018-04-02
GH-136 Make linker safe!

jamesmontemagno:
### Description of Change ###

Ensure linker safe! Enabled LinkerSafe attributes on iOS/Android. Also updated screenmetrics with a preserve so it doesn't get removed.

### Bugs Fixed ###

- Related to issue #136 

### API Changes ###

Made ScreenMetrics a struct to align with other types that we have inside of a class.

Added preserve to it so it works.

### Behavioral Changes ###

Added linker settings for sample projects for better testing.

### PR Checklist ###


Caboodle/DeviceInfo/DeviceInfo.shared.cs
```diff
@@ -98,8 +98,22 @@ public enum DeviceType
         Virtual
     }
 
-    public class ScreenMetrics
+#if __ANDROID__
+    [Android.Runtime.Preserve(AllMembers = true)]
+#elif __IOS__
+    [Foundation.Preserve(AllMembers = true)]
+#endif
+    public struct ScreenMetrics
```
mattleibow: Do we have any idea why this type gets removed if the device info is being used?

jamesmontemagno: I don't to be honest with you. It is kind of odd because other ones are just fine. I would have contact the linker team i guess. However not too worried about a single struct.

jamesmontemagno:
build

Caboodle/DeviceInfo/DeviceInfo.shared.cs
mattleibow: Maybe it is because it is a struct... the other types are classes.

jamesmontemagno: it was a class originally, i changed to struct :)

mattleibow: :sob:

Issue #156, opened 2018-04-09
[iOS] NullRef with sensors on background thread


mattleibow:
When on a background thread (such as during the tests), the sensors throw an exception.
The reason for this is that `NSOperationQueue.CurrentQueue` is `null`.

jamesmontemagno:
Yeah, looks like if called from outside caller it will be null: https://developer.apple.com/documentation/foundation/nsoperationqueue/1413097-currentqueue

PR #157, opened 2018-04-09
GH-156 Create new queue if null queue

jamesmontemagno:
via:  https://stackoverflow.com/questions/33642590/nsoperationqueue-currentqueue-not-working

### Description of Change ###

If the current queue is null because it is being called from an outside context then create a new one :)

### Bugs Fixed ###

- Related to issue #156

### PR Checklist ###

- [x] Updated documentation ([see walkthrough](https://github.com/xamarin/Xamarin.Essentials/wiki/Documenting-your-code-with-mdoc))

### DIFF
--- a/Xamarin.Essentials/Platform/Platform.ios.cs
+++ b/Xamarin.Essentials/Platform/Platform.ios.cs
@@ -58,7 +58,7 @@ namespace Xamarin.Essentials
         internal static CMMotionManager MotionManager =>
             motionManager ?? (motionManager = new CMMotionManager());
 
-        internal static NSOperationQueue CurrentQueue =>
+        internal static NSOperationQueue GetCurrentQueue() =>
             NSOperationQueue.CurrentQueue ?? new NSOperationQueue();
     }
 }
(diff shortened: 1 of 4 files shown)


Xamarin.Essentials/Platform/Platform.ios.cs
```diff
@@ -57,5 +57,8 @@ internal static UIViewController GetCurrentViewController(bool throwIfNull = tru
 
         internal static CMMotionManager MotionManager =>
             motionManager ?? (motionManager = new CMMotionManager());
+
+        internal static NSOperationQueue CurrentQueue =>
+            NSOperationQueue.CurrentQueue ?? new NSOperationQueue();
```
mattleibow: We just have to remember (or maybe make a note) that this is going to fail:

```csharp
var myThing = DoSomething(Platform.CurrentQueue);

if (myThing.Queue == Platform.CurrentQueue) {
    // in my mind
} else {
    // say whaa?!?
}
```

mattleibow: Maybe we should change this from a property to a method like `Platform.GetQueue()` or something that indicates that it does not return the same object in some cases.

mattleibow: But, this is internal so a comment may be just fine as well.

jamesmontemagno: yeah, i like get queue so it doesn't think it is always the same. Good call.

PR #158, opened 2018-04-09
Update private preferences for version tracking & secure

jamesmontemagno:
### Description of Change ###

During re-name it add xamarin.essentials. I think "essentials" is good.

### PR Checklist ###

- [x] Updated documentation ([see walkthrough](https://github.com/xamarin/Xamarin.Essentials/wiki/Documenting-your-code-with-mdoc))


mattleibow:
I am thinking this might be better served with some fields somewhere, like `Utils.PrivatePreferencesSharedName` or something... maybe in `Preferences` or in `Platform`. This way we don't ever have to fix this code in two places again.ever 😭 😿

Samples/Samples/View/BrowserPage.xaml
```diff
@@ -17,9 +17,9 @@
             <Label Text="Browser Launch Type:"/>
             <Picker HorizontalOptions="FillAndExpand"
                     ItemsSource="{Binding BrowserLaunchTypes}" 
-                    SelectedIndex="{Binding BrowserType, Mode=TwoWay}" />
+                    SelectedItem="{Binding BrowserType, Mode=TwoWay}" />
```
mattleibow: Isn't the items source `BrowserLaunchTypes` is a list of strings? You appear to be mapping it to the `BrowserType` property which is an int, and you are using the `SelectedItem` property - won't this be mapping a string to an int?

Xamarin.Essentials/Preferences/Preferences.shared.cs
```diff
@@ -2,6 +2,9 @@
 {
     public static partial class Preferences
     {
+        internal static string PrivatePreferencesSharedName =>
+            $"{AppInfo.PackageName}.xamarinessentials";
```
mattleibow: NICE!

PR #161, opened 2018-04-09
Using the new mdoc frameworks

mattleibow:
Updating the way we generate docs


docs/en/frameworks.xml
```diff
@@ -0,0 +1,7 @@
+<Frameworks>
+  <Framework Name="Xamarin.Essentials.Android" Source="monoandroid71" />
+  <Framework Name="Xamarin.Essentials.NetStandard1.0" Source="netstandard1.0" />
+  <Framework Name="Xamarin.Essentials.NetStandard2.0" Source="netstandard2.0" />
+  <Framework Name="Xamarin.Essentials.UWP" Source="uap10.0.16299" />
+  <Framework Name="Xamarin.Essentials.iOS" Source="xamarin.ios10" />
```
mattleibow: These are the frameworks that I have added, but there are still some points for discussion:
 - do we want the .NET Standard 1.0 to be visible to the public? (this is really a compat for PCL thing)
 - those `Name="..."`, do we want those or do we want some changes:
    - do we want to include version numbers `"Xamarin.Essentials.UWP.10.0.16299"`
 - those `Source="..."`, do we want the build versions?
    - I just used the target frameworks as it was easy
    - how do we map what the UI has to this framework? If there is a mapping outside the repo, should we drop the version bits? (as soon as Xamarin.Forms updates, we will move to target monoandroid8.0 and we don't want to keep having to talk to the docs team for this)
    - we could potentially do anything here...

docs/packages.docs.config
```diff
@@ -1,4 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="mdoc" version="5.6.1"  />
-</packages>
```
mattleibow: I removed this file, is it still needed?

docs/en/frameworks.xml
jamesmontemagno: I think it is fine that is net standard 1.0, they are the same API :)

I am not sure about the others... @joelmartinez @conceptdev ?

conceptdev: @joelmartinez may know more - it might just create a version menu on the [docs api site](https://docs.microsoft.com/en-us/dotnet/api/)? i don't immediately know of any issues

docs/packages.docs.config
joelmartinez: @mattleibow correct, it is no longer needed ... since you're pulling mdoc through msbuild :)

docs/frameworks.xml
```diff
@@ -0,0 +1,6 @@
+<Frameworks>
+  <Framework Name="Xamarin.Essentials" Source="netstandard2.0" />
+  <Framework Name="Xamarin.Essentials.Android" Source="monoandroid71" />
+  <Framework Name="Xamarin.Essentials.iOS" Source="xamarin.ios10" />
+  <Framework Name="Xamarin.Essentials.UWP" Source="uap10.0.16299" />
+</Frameworks>
```
mattleibow: This is the new frameworks.xml file. How does this relate to the UI, if at all?

Issue #457, opened 2018-08-08
SecureStorage.ios.cs throws improper Exceptions.


robv8r:
Bug report best practices: [Submitting Issues](SUBMITTING-ISSUES.md) <-- 404 File Not Found :-)

### Description

The source code for [SecureStorage.ios.cs](https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/SecureStorage/SecureStorage.ios.cs) throws `System.Exception` instead of a more appropriate exception.

The first is in [SetValueForKey](https://github.com/xamarin/Essentials/blob/0.9.1-preview/Xamarin.Essentials/SecureStorage/SecureStorage.ios.cs#L101).  The second is in [RemoveRecord](https://github.com/xamarin/Essentials/blob/0.9.1-preview/Xamarin.Essentials/SecureStorage/SecureStorage.ios.cs#L144)

This is not a proper use of Exceptions.  A developer should neither throw nor catch a `System.Exception` per [Microsoft's Best Practices](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types).

### Expected Behavior

A more appropriate exception that can be handled at runtime.  For example, a `System.ArgumentOutOfRangeException` if the value could not be removed because the key didn't exist.

### Actual Behavior

If a value cannot be set or removed, a generic `System.Exception` is thrown.

### Basic Information

- Version with issue: 0.9.1
- Last known good version: N/A
- IDE: Visual Studio 2017 Enterprise 
- Platform Target Frameworks: 
  - iOS:  All versions

jamesmontemagno:
Will assign to @Redth to cleanup this exception.

I would need to see which of these are correct or if it should be a custom exception.

Mrnikbobjeff:
No exception in the [best standard docs](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types) fits our scenario. We should use [ExternalException](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.externalexception(v=vs.110).aspx) as it fits our use case exactly. External calls from our application which returns an error code.

robv8r:
In this case, always throwing an `ExternalException` may not be descriptive enough for developers.

[...]

PR #500, opened 2018-09-04
GH-457 Add new exception for SecureStorage

jamesmontemagno:
### Description of Change ###

Adds a new SecureStorageException to be thrown by iOS to provide more details.

### Bugs Fixed ###

- Related to issue #457 

### API Changes ###

Added: 
 
- SecureStorage 
- enum SecureStorageErrorCode

### Behavioral Changes ###

### PR Checklist ###

- [ x Has samples (if omitted, state reason in description)
- [x]Rebased on top of master at time of PR

### DIFF
--- a/Xamarin.Essentials/Types/SecureStorageErrorCode.ios.cs
+++ b/Xamarin.Essentials/Types/SecureStorageErrorCode.ios.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using Security;
+﻿using Security;
 
 namespace Xamarin.Essentials
 {


jamesmontemagno:
Thoughts @Redth ?

Redth:
I really think if we're going to do this, we should see how it maps to the other platforms.  I'm also wondering if having all of these error codes translated to an Essentials enum really adds that much more value than just writing out the SecStatusCode's .ToString() on iOS in the exception message?

I think the next step is to see what errors come out of android and UWP and how they would map to these enums, otherwise I don't think it makes a lot of sense to have an enum that's 'cross platform' but really only used in the case of iOS.

mattleibow:
It might just be fine to use `Exception` and pass the native exception as the inner exception - that is what it is there for. ([inner exception docs](https://docs.microsoft.com/en-us/dotnet/api/system.exception.innerexception))

We can also create a new "generic" exception to throw so that we don't ever throw the base `Exception`.

Thoughts?

mattleibow:
OK, so that was a partially useless comment - this is just an error code.
But, the same thing applies, we can either create a special `SecureStorageException`, or we can create a generic `XamarinEssentialsException` that is used for all the non-specific errors.

In either case, we can probably just create the exception with the message: `$"Error doing something: {errorCode.GetStatusDescription()} Error Code: {errorCode}."` (there appears to be an extension method `GetStatusDescription`)

Issue #515, opened 2018-09-21
ProGuard enabled crashes Android app


JaccoDieleman:
### Description
After consuming the Xamarin.Essentials NuGet package my Android app crashes during start up.

### Steps to Reproduce
The problem can be reproduced with the Samples.Android project in this repository.
To reproduce, check "Enable ProGuard" in the Release configuration and try starting the app in release mode.
### Expected Behavior
App starts

### Actual Behavior
App crashes

### Basic Information

- Version with issue: 0.10.0-preview
- IDE: Visual Studio Professional for Windows 15.7.6
- Platform Target Frameworks:
  - Android:  8.1

EmilAlipiev:
you should added output window with stacktrace information. I am also using proguard and full linking but no problem so far. you may include the apis you are accessing. it could be related to one of them as well.

JaccoDieleman:
@EmilAlipiev I'm trying to use GeoLocation. The app however crashes before it tries to use GeoLocation, it crashes while the splashScreen is being shown. The SampleApp from this repo uses all apis and has the same behavior.

I don't get a stacktrace in Visual Studio. However, this is what I picked from the device log:
```Time	Device Name	Type	PID	Tag	Message
09-24 09:19:52.634	LGE Nexus 5X	Error	23888	AndroidRuntime	Caused by: java.lang.ClassNotFoundException: Didn't find class "android.support.v7.widget.FitWindowsFrameLayout" on path: DexPathList[[zip file "/data/app/com.easysystems.easyexpense-XygxF7LxfTqJsFStVPWJYg==/base.apk"],nativeLibraryDirectories=[/data/app/com.easysystems.easyexpense-XygxF7LxfTqJsFStVPWJYg==/lib/arm, /data/app/com.easysystems.easyexpense-XygxF7LxfTqJsFStVPWJYg==/base.apk!/lib/armeabi-v7a, /system/lib, /vendor/lib]]
	at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:125)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
	at android.view.LayoutInflater.createView(LayoutInflater.java:606)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:790)
	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730)
[...]

PR #530, opened 2018-10-01
Add progaurd file for reference

jamesmontemagno:
### Description of Change ###

Reference progaurd file.

### Bugs Fixed ###

- Related to issue #515


conceptdev:
I have a feeling this is not a valid tag `<see cref="https://developer.xamarin.com/api/type/Security.SecAccessible/" />` probably it should just be `<see cref="T:Security.SecAccessible" />` or some variation. It wasn't done in this PR -- but I guess on master prior to the branch? @lobrien does that sound right?

Issue #448, opened 2018-08-06
AEADBadTagException when getting a secure setting


johnnyasantoss:
### Description
The error occurs when trying to get a secure setting.
I don't have exactly a reproducible environment, but what happen to me is that I was a few day without build this app that I'm developing and now It throws AEADBadTagException (I'm developing it solo, so there where no modifications between this two builds - from the same branch, same rev)

### Basic Information

- Version with issue: 0.8.0-preview
- IDE: Visual Studio 15.7.5 & Rider 2018.1.3
- Platform Target Frameworks:
  - Android: 7.1 API 25
- Android Support Library Version: 27.0.2.1

Stacktrace: 
```log
 Javax.Crypto.AEADBadTagException: Exception of type 'Javax.Crypto.AEADBadTagException' was thrown. ---> Java.Lang.Exception: Signature/MAC verification failed
   --- End of inner exception stack trace ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <f32579baafc1404fa37ba3ec1abdc0bd>:0 
  at Java.Interop.JniEnvironment+InstanceMethods.CallNonvirtualObjectMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x00089] in <7802aa64ad574c33adca332a3fa9706a>:0 
  at Java.Interop.JniPeerMembers+JniInstanceMethods.InvokeNonvirtualObjectMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters) [0x0001f] in <7802aa64ad574c33adca332a3fa9706a>:0 
  at Javax.Crypto.Cipher.DoFinal (System.Byte[] input, System.Int32 inputOffset, System.Int32 inputLen) [0x00052] in <263adecfa58f4c449f1ff56156d886fd>:0 
  at Xamarin.Essentials.AndroidKeyStore.Decrypt (System.Byte[] data) [0x0005f] in <39c36f0831c64fc491c61f5f8d5bbdb5>:0 
  at Xamarin.Essentials.SecureStorage.PlatformGetAsync (System.String key) [0x0004f] in <39c36f0831c64fc491c61f5f8d5bbdb5>:0 
[...]

PR #540, opened 2018-10-05
Add sample of backup rools, and handl AEADBadTagException

jamesmontemagno:
### Description of Change ###

Test out custom xml for not backing up specific data to google as sample. Also handle the AEADBadTagException and just remove the key and return null.

### Bugs Fixed ###

- Related to issue #448

### API Changes ###

None

### Behavioral Changes ###

Before an exception would be thrown, but now it will be handled correctly and remove the key.

### PR Checklist ###

### DIFF
--- a/Samples/Samples.Android/Properties/AndroidManifest.xml
+++ b/Samples/Samples.Android/Properties/AndroidManifest.xml
@@ -12,5 +12,11 @@
   <uses-feature android:name="android.hardware.location" android:required="false" />
   <uses-feature android:name="android.hardware.location.gps" android:required="false" />
   <uses-feature android:name="android.hardware.location.network" android:required="false" />
-  <application android:label="@string/app_name" android:icon="@mipmap/ic_launcher" android:roundIcon="@mipmap/ic_launcher_round" android:theme="@style/MainTheme"></application>
+  <application android:label="@string/app_name" 
+               android:icon="@mipmap/ic_launcher" 
+               android:roundIcon="@mipmap/ic_launcher_round" 
+               android:theme="@style/MainTheme"
+               android:fullBackupContent="@xml/my_backup_rules">
+    
+  </application>
 </manifest>
--- a/Samples/Samples.Android/Resources/xml/my_backup_rules.xml
+++ b/Samples/Samples.Android/Resources/xml/my_backup_rules.xml
@@ -1,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <include domain="sharedpref" path="."/>
+    <exclude domain="sharedpref" path="${applicationId}.xamarinessentials.xml"/>
+</full-backup-content>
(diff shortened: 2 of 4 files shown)


Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
```diff
@@ -26,7 +26,15 @@ static Task<string> PlatformGetAsync(string key)
...
                 var encData = Convert.FromBase64String(encStr);
                 var ks = new AndroidKeyStore(context, Alias, AlwaysUseAsymmetricKeyStorage);
-                decryptedData = ks.Decrypt(encData);
+                try
+                {
+                    decryptedData = ks.Decrypt(encData);
+                }
+                catch (AEADBadTagException)
+                {
+                    System.Diagnostics.Debug.WriteLine($"Unable to decrpt key {key} likely due to uninstall, removing old key.");
```
mattleibow: This is OK here, but since we compile in release mode, I am thinking that if we want other devs to see this, we may have to use the android log. This probably goes for the other platforms - we should use the platform logging if we need the developer to see these errors and things.

jamesmontemagno: Yeah, my thought was that we probably don't want to leak the key's name in release mode.

Samples/Samples.Android/Resources/xml/my_backup_rules.xml
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <include domain="sharedpref" path="."/>
+    <exclude domain="sharedpref" path="${applicationId}.xamarinessentials.xml"/>
```
mattleibow: If we merge this into the actual library, my concern is that we may conflict with another resource - especially since we have to specify this on the `<application>` element in the manifest. But... we can add this to the library and then let the user know that for better things they should use our resource - if they have their own, then they should add the line to theirs for better security.

Samples/Samples.Android/Properties/AndroidManifest.xml
```diff
@@ -12,5 +12,11 @@
   <uses-feature android:name="android.hardware.location" android:required="false" />
   <uses-feature android:name="android.hardware.location.gps" android:required="false" />
   <uses-feature android:name="android.hardware.location.network" android:required="false" />
-  <application android:label="@string/app_name" android:icon="@mipmap/ic_launcher" android:roundIcon="@mipmap/ic_launcher_round" android:theme="@style/MainTheme"></application>
+  <application android:label="@string/app_name" 
+               android:icon="@mipmap/ic_launcher" 
+               android:roundIcon="@mipmap/ic_launcher_round" 
+               android:theme="@style/MainTheme"
+               android:fullBackupContent="@xml/my_backup_rules">
```
mattleibow: Can we have our rules and the user rules together? This appears to be an exclusive resource, so we may have to just let the user know - see my other comment.

Samples/Samples.Android/Resources/xml/my_backup_rules.xml
jamesmontemagno: This is just for the sample app so I can test it. We wouldn't put it into the actual app as I don't think that is a good idea. 

We would just document it.

PR #558, opened 2018-10-20
Catch FeatureNotSupportedException when navigating back for Barometer

manutdkid77:
### Bugs Fixed ###

If barometer was not supported when navigating back from the Barometer page in the Samples app, FeatureNotSupportedException was thrown and navigation was not possible.


jamesmontemagno:
build

Issue #568, opened 2018-11-01
[api-fix] Browser


Redth:
- `Browser.OpenAsync` should return `Task<bool>`
  - Update documentation with the fact that this method doesn't return after the browser is closed, but after it is launched, and that the result is whether the opening was successful or not. 
  - iOS: Use Async methods where appropriate in platform in implementation, including using one that returns a `bool`.
- `BrowserLaunchMode.SystemPreferred` should become `0` value.

PR #589, opened 2018-11-02
GH-568: [api-fix] Browser

Redth:
See https://github.com/xamarin/Essentials/issues/568

### DIFF
--- a/Xamarin.Essentials/Browser/Browser.shared.cs
+++ b/Xamarin.Essentials/Browser/Browser.shared.cs
@@ -21,7 +21,7 @@ namespace Xamarin.Essentials
         public static Task OpenAsync(Uri uri) =>
           OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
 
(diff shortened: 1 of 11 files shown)


Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -37,7 +37,7 @@ internal static Uri EscapeUri(Uri uri)
 
     public enum BrowserLaunchMode
     {
+        SystemPreferred = 0,
         External,
```
jamesmontemagno: Should we put = 1 here?

mattleibow: I have no real strong opinion here, but maybe for consistency?

Redth: Not against putting the 1 as well but I do like having 0 so it's obvious where it starts...

I think having 0 is sufficient. But again not against adding 1,2 etc to the others

Issue #579, opened 2018-11-01
[api-fix] Platform


Redth:
On netstandard, `Platform` should be marked internal

PR #592, opened 2018-11-02
GH-579: [api-fix] Platform

mattleibow:
See #579

### DIFF
--- a/Xamarin.Essentials/Platform/Platform.shared.cs
+++ b/Xamarin.Essentials/Platform/Platform.shared.cs
@@ -1,6 +1,8 @@
 ﻿namespace Xamarin.Essentials
 {
+#if !NETSTANDARD
     public static partial class Platform
     {
     }
+#endif
 }


Xamarin.Essentials/Platform/Platform.shared.cs
```diff
@@ -1,6 +1,8 @@
 ﻿namespace Xamarin.Essentials
 {
+#if !NETSTANDARD
     public static partial class Platform
```
jamesmontemagno: Could we make it internal? Or would this break bait and switch?

mattleibow: I can't do internal since the style rules clash :(
 - if partial, then it requires the visibility
 - if internal, then remove because it is the default

This will bite us :) We can change the rules, but one will have to go. If we have to pick one, I say we can drip the partial requirement as this is usually the smallest chance of error.

PR #611, opened 2018-11-05
GH-510: If there is no main window, then there is no UI thread

### LINKED ISSUE #510: Evaluate throwing if Window is null on UWP
Should we be throwing here? https://github.com/xamarin/Essentials/blob/f56c8baf787bbc14246308fb90ed2ba85b6831ba/Xamarin.Essentials/MainThread/MainThread.uwp.cs#L16

 - [x] Check if there is a main window, if not run on current thread
 - [x] Audit usage of any Platform main thread usage and change to shared

mattleibow:
If there is no main window, then there is no UI thread.

Closes #510

### DIFF
--- a/Xamarin.Essentials/MainThread/MainThread.uwp.cs
+++ b/Xamarin.Essentials/MainThread/MainThread.uwp.cs
@@ -6,8 +6,19 @@ namespace Xamarin.Essentials
 {
     public static partial class MainThread
     {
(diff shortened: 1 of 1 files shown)


jamesmontemagno:
Looks good, we always use "BeginInvokeOnMainThread" so this is good to go!

jamesmontemagno:
build

PR #416, opened 2018-07-26
GH-130 & GH-129: Android support for safe shareable file URI’s

Redth:
On later versions of Android, you have to wrap streams of data you want to share outside your app (between apps) in a stream through a content provider.  Android Support providers a general use FileProvider we can use for this.  This commit basically adds support for getting all the right AndroidManifest declarations for the custom file provider based on the android support provider, so that we expose an internal method which gets a URI safe for sharing outside of the app.

### Bugs Fixed ###

- Fixes #130
- Fixes #129 
- Fixes #425

### API Changes ###

List all API changes here (or just put None), example:

Added: 

```
public partial static Platform
{
    internal static AndroidUri GetShareableFileUri(string filename);
}
 
// This type needs to be exposed public afaik to be used, so it's added in the .Types namespace to be a bit more hidden from public use
public class FileProvider { }
```

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Has samples (if omitted, state reason in description)
- [x] Rebased on top of master at time of PR
- [x] Changes adhere to coding standard
- [x] Updated documentation ([see walkthrough](https://github.com/xamarin/Essentials/wiki/Documenting-your-code-with-mdoc))

Issue #130, opened 2018-03-29
FileProvider for Android


Redth:
Doing things like attaching files to emails and sharing files with DataTransfer is difficult on android since one must get a URI from a file provider in newer versions of android.

We should add some functionality to help get a proper URI from an existing filename the developer passes to us which we then transfer into a file provider friendly filename.

# What's needed

## XML Resource

You must specify the paths you intend to use in an xml resource file.  Most users would want to use one or both of these paths:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<paths>
  <external-path path "caboodle" name="caboodle/" />

  <!-- NOT NEEDED: -->
  <!-- <external-path path="Android/data/com.some.package/" name="files_root" /> -->
  <!-- <external-path path="." name="external_storage_root" /> -->
</paths>
```

**EDIT:** We probably only need to do something like `<external-path path "caboodle" name="caboodle" />` or similar... We don't need external storage and we don't need the package name in the path it appears.. so this is good.

We can just add this as a resource in the library itself.

## Provider declaration in AndroidManifest.xml

The AndroidManifest.xml file needs this declared:
```xml
<provider
  android:name="android.support.v4.content.FileProvider" 
  android:authorities="${applicationId}.fileProvider"
  android:grantUriPermissions="true"
  android:exported="false">
    <meta-data 
      android:name="android.support.FILE_PROVIDER_PATHS"
      android:resource="@xml/file_paths" />
</provider>
```

I don't believe we can accomplish this currently with C# attributes (there doesn't seem to be an `[assembly: Provider(..)]` attribute.

Also note the `@xml/file_paths` resource name needs to be known for this declaration.

We could get the user to add this manually (again, not nice). Or we can use a custom build task.

## Content URI
Finally we get the actual content uri:

```csharp
var contentUri = Android.Support.V4.Content.FileProvider.GetUriForFile(context, "com.some.package", new Java.IO.File(filename));
```

We can use this uri for things like starting an intent chooser:

```csharp
var intent = new Android.Content.Intent(Android.Content.Intent.ActionSend);
intent.PutExtra(Android.Content.Intent.ExtraStream, contentUri);
intent.SetFlags(Android.Content.ActivityFlags.GrantReadUriPermission);

var intentChooser = Android.Content.Intent.CreateChooser(intent, "Title");

context.StartActivity(intentChooser);
```

> VS bug [#732192](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/732192)

[...]

Issue #129, opened 2018-03-29
Add Attachments to Email


mattleibow:
This needs to be investigated as there are some limitations and additional steps on each platform.

> VS bug [#735679](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735679)

mfeingol:
This would be fantastic. I'm looking to replace platform-specific code in my app with Xamarin.Essentials, and this is very important functionality for me.

transis2:
Here we have the implementation for iOS, Android, UWP and Tizen

https://github.com/cjlotz/Xamarin.Plugins/blob/master/Messaging/Plugin.Messaging.iOSUnified/EmailTask.cs

https://github.com/cjlotz/Xamarin.Plugins/blob/master/Messaging/Plugin.Messaging.Android/EmailTask.cs

https://github.com/cjlotz/Xamarin.Plugins/blob/master/Messaging/Plugin.Messaging.UWP/EmailTask.cs

https://github.com/cjlotz/Xamarin.Plugins/blob/master/Messaging/Plugin.Messaging.Tizen/EmailTask.cs

mfeingol:
Will this ever be merged?

jamesmontemagno:
@mfeingol we are actively working on locking down 1.0 apis and will be including more features after that. There is no full PR for this besides just are first investigation into the file provider.

ephraimf:
Is an IEmailTask.cs file available?
I don't have a definition for the fields in the interface or for the interface itself.

minaairsupport:
this important feature to include as soon as possible  , I can do it for you and make pull request if you will accept

ephraimf:
Will there be support for multiple attachments?

mattleibow:
@ephraimf Yes. There should be support for an arbitrary number of arbitrary files. Unless the platform specifically does not support it - but I haven't found that.

robsoft:
Any likely eta on this, please? I need to get something into a demo in the next couple of days, trying to decide if I should just use Xam.Plugins.Messaging for the time being (and not sure what impact adding that might have on my Xamarin.Essentials stuff already in the app). If it was to drop in the next few days, I'd kick myself!

amirvenus:
Would be nice to have it implemented as I am currently using DI for sending an email that has attachment(s).

choirunnisa:
When can we expect to see th attachment functionality?

ephraimf:
Thank you very much.

[...]

PR #508, opened 2018-09-16
Screenshot API

pictos:
### Description of Change ###

Adds new feature for screenshot.

### Proposal ###

- Related to issue #125 

### API Changes ###

Added:

- Task<string> CaptureAsync(ScreenOutputType type, string fileName)

- Task<byte[]> GetImageBytesAsync(ScreenOutputType type)

- enum ScreenOutputType

### PR Checklist ###

- [ x ] Has samples (if omitted, state reason in description)
- [ x ] Rebased on top of master at time of PR
- [ x ] Changes adhere to coding standard
- [ x ] Updated documentation ([see walkthrough]

PR #613, opened 2018-11-07
Add a Media Picker

mattleibow:
Adding a media picker...

With the new file type, there may be cause to evaluate the file / attachment types in #129, #130 and #416


Samples/Samples/ViewModel/MediaPickerViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+                    return;
+                }
+
+                // save the file into local storage
+                var newFile = Path.Combine(FileSystem.CacheDirectory, photo.FileName);
+                using (var stream = await photo.OpenReadAsync())
+                using (var newStream = File.OpenWrite(newFile))
+                {
+                    await stream.CopyToAsync(newStream);
+                }
```
mattleibow: This is the not-too-heavy code to copy the picked image to a local folder for viewing. We could add helpers, such as `MediaFile.CopyAsync(Stream)` and `MediaFile.CopyAsync(string)`.

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
```diff
@@ -0,0 +1,121 @@
...
+            PickVideo = Base + 3,
+            TakeVideo = Base + 4,
+
+            Min = PickPhoto,
+            Max = TakeVideo,
+        }
+
+        static MediaPicker()
+        {
+            Platform.ActivityResult += OnActivityResult;
```
mattleibow: We could subscribe to this event for some other API

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
```diff
@@ -0,0 +1,121 @@
...
+            // make sure we have permission and an activity
+            await Permissions.RequireAsync(PermissionType.ExternalStorage);
+
+            var intent = new Intent(Intent.ActionPick);
+            intent.SetType("image/*");
+
+            try
+            {
+                // launch the picker intent via the intermediate activity
+                var data = await IntermediateActivity.StartAsync(intent, (int)RequestCode.PickPhoto);
```
mattleibow: We can use a `Task` to get the result of an activity. This should be OK as the task will complete when the picker activity completes. We may have to investigate what happens if the activity is killed and then the task resumes. There may be some mono/managed code that will run as part of the completing async machine.

Xamarin.Essentials/Platform/Platform.android.cs
```diff
@@ -152,4 +159,95 @@ void Application.IActivityLifecycleCallbacks.OnActivityStopped(Activity activity
         {
         }
     }
+
+    [Activity(ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
+    class IntermediateActivity : Activity
```
mattleibow: This intermediate activity is totally generic and can be used for any sort of external activity. Not only that, it will always ensure that the tasks will complete. The only exception is if the ENTIRE app gets killed - in this case, nothing matters as all managed state is lost and the tasks will have been cleared.

Xamarin.Essentials/MediaPicker/MediaPicker.uwp.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.Storage.Pickers;
+
+namespace Xamarin.Essentials
+{
+    public static partial class MediaPicker
+    {
+        static readonly string[] defaultVideoFileTypes = new[] { ".mp4", ".wmv", ".avi" };
+        static readonly string[] defaultImageFileTypes = new[] { ".jpeg", ".jpg", ".png", ".gif", ".bmp" };
```
mattleibow: We may need to evaluate this and see if we want more or makle it customizable. TIFF anyone?

Xamarin.Essentials/MediaPicker/MediaPicker.uwp.cs
```diff
@@ -0,0 +1,55 @@
...
+            picker.ViewMode = PickerViewMode.Thumbnail;
+
+            // show the picker
+            var result = await picker.PickSingleFileAsync();
+
+            // cancelled
+            if (result == null)
+            {
+                OnMediaPicked(new MediaPickedEventArgs(true));
+                return null;
```
mattleibow: We want to (optionally) fire the global event and then resure for the task value.

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
mattleibow: No events!

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (requestCode == (int)RequestCode.TakePhoto || requestCode == (int)RequestCode.TakeVideo)
+            {
+            }
+
+            // something went wrong
+            return null;
+        }
+    }
+
+    public partial class MediaFile
```
Mrnikbobjeff: I would love for the MediaFile class to be abstract or provide virtual overloads for reading as byte array or stream. I started working on the play audio API, where we would also need a media file. There we would have the user pass a MediaFile, which would necessarily have to be a file on the disk. If the class were abstract or allow overrideing AsStream etc. the user could create his own in memory mediafile with e.g. a byte array as backing store and not a File. Also, Perhaps include an enum in here to signify media type, we will have at least 3 different types very soon :)

Mrnikbobjeff:
Is there an ETA for this feature? I am working on 3 branches which would benefit from a common MediaFile class, the less work I would have to redo the better.

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
Redth: I still think it might be worth having the Intermediate Activity have the _option_ to start a foreground service for the life of itself so that Android absolutely will _not_ kill the activity.  The foreground service needs a persistent notification which could just have the current app's name.  The notification will go away as soon as the intermediate activity is finished.  We could gate this behaviour with a platform specific flag (`MediaPicker.KeepAliveWithForegroundService`) or something.

Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
```diff
@@ -0,0 +1,71 @@
...
+                this.tcs = tcs;
+            }
+
+            public override void FinishedPickingMedia(UIImagePickerController picker, NSDictionary info)
+            {
+                // TODO: investgate iOS 11 and it's suggestion of
+                //       using `(PHAsset)info[UIImagePickerController.PHAsset];`
+
+                var url = (NSUrl)info[UIImagePickerController.ReferenceUrl];
+                tcs.SetResult(new MediaFile(url?.AbsoluteString));
```
Mrnikbobjeff: Can url ever be null? If so, this will throw unconditionally as we check for null in the mediafile constructor

Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            // permission is not required on iOS 11 for the picker
+            if (!Platform.HasOSVersion(11, 0))
+                await Permissions.RequestAsync(PermissionType.Photos);
+
+            var vc = Platform.GetCurrentViewController();
+
+            var picker = new UIImagePickerController();
+            picker.SourceType = UIImagePickerControllerSourceType.PhotoLibrary;
+            picker.MediaTypes = new string[] { UTType.Image };
```
Mrnikbobjeff: Perhaps make this a static readonly to avoid constant array allocation, as done in uwp

Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            await vc.PresentViewControllerAsync(picker, true);
+
+            return await tcs.Task;
+        }
+
+        class PhotoPickerDelegate : UIImagePickerControllerDelegate
+        {
+            TaskCompletionSource<MediaFile> tcs;
```
Mrnikbobjeff: This can be made readonly

1Cor125:
I see some progress on the dev/add-media-picker branch. There is also a dependency on the new FilePicker feature. Are these still being worked on? Could these features use another contributor? I need to implement a media picker for iOS, Android and UWP anyway so I would prefer to contribute to something in Xamarin.Essentials than to roll my own and end up moving to Essentials later. Is there a target date for 1.6?

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
```diff
@@ -0,0 +1,99 @@
...
+using Android.Content;
+using Android.Content.PM;
+using Android.Provider;
+
+namespace Xamarin.Essentials
+{
+    public static partial class MediaPicker
+    {
+        const int requestCodeMediaPicker = 12348;
+        const int requestCodeMediaCapture = 12349;
```
mattleibow, line 17: At some point we need to make sure we don't have any clashes of codes in our library.

Xamarin.Essentials/MediaPicker/MediaPicker.shared.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Xamarin.Essentials
+{
+    public static partial class MediaPicker
+    {
+        public static bool IsPhotoCaptureAvailable
+            => PlatformIsPhotoCaptureAvailable;
```
mattleibow, line 10: Is this just photo? Or photo and video? Maybe should be `IsMediaCaptureAvailable`?

Xamarin.Essentials/MediaPicker/MediaPicker.tizen.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Tizen.Applications;
+
+namespace Xamarin.Essentials
+{
+    public static partial class MediaPicker
+    {
+        static bool PlatformIsPhotoCaptureAvailable
+               => false;
```
mattleibow, line 13: Always false?

Xamarin.Essentials/FileSystem/FileSystem.android.cs
```diff
@@ -29,6 +31,14 @@ static Task<Stream> PlatformOpenAppPackageFileAsync(string filename)
...
             }
         }
+
+        internal static string GetAbsolutePath(this global::Android.Net.Uri uri)
+        {
+            using (var file = new Java.IO.File(uri.ToString()))
+            {
+                return file.AbsolutePath;
+            }
+        }
```
mattleibow, line 41: This doesn't look needed anymore?

Xamarin.Essentials/MediaPicker/MediaPicker.android.cs
Redth: Yeah I actually had the thought of moving this into a simple Platform internal API where we can just go `NextRequestCode()` and it manages them.  Maybe should do that quick.

Xamarin.Essentials/MediaPicker/MediaPicker.shared.cs
Redth: Switched this to `IsCaptureAvailable`

Xamarin.Essentials/MediaPicker/MediaPicker.tizen.cs
Redth: This is fixed now.

Xamarin.Essentials/FileSystem/FileSystem.android.cs
Redth: Removed!

mattleibow:
I am going to try merge the `FilePickerResult` and `MediaPickerResult` types into a single `FileResult` so that we can reduce duplication. Also, when the #508 gets merged, we can use the same type.

There is the `ReadOnlyFile` type, but there ight be a case where the picker result is not readonly... This may need investigation...

Issue #480, opened 2018-08-19
Connectivity - Add Wi-Fi signal strength


YZahringer:
## Description of Feature
Get the signal strenght for connected Wi-Fi network. 

### Specific platform APIs
- [iOS](https://developer.apple.com/documentation/networkextension/nehotspotnetwork/1618923-signalstrength)
- [Android](https://developer.android.com/reference/android/net/wifi/WifiManager#calculateSignalLevel(int,%20int))
- [UWP](https://docs.microsoft.com/en-us/uwp/api/windows.devices.wifi.wifiavailablenetwork.signalbars#Windows_Devices_WiFi_WiFiAvailableNetwork_SignalBars)

> VS bug [#733079](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/733079)

Redth:
Someone needs to figure out how we normalize this and provide an API spec for it.

Given that UWP returns Signal Bars we probably need to be pretty simple in our API.  I'm guessing this is some value between like 1-5 only, which is probably good enough for what most people would _use_ this API for (to detect a weak or really good signal).  The UWP docs don't really say what the expected value ranges are for this so we need to figure that out.

If someone wants to draft a spec and figure out how we normalize this across the platforms we can continue the discussion!

Mrnikbobjeff:
Normalisation is quite forard. Android offers inbuild normalization which "should be used any time signal strength is shown." iOS offers a range between 0 and 1, which is also super easy to adapt to whatever degree of detail we choose. The UWP doc is pretty unspecific, I only managed to reproduce values between one and three. I'd thus suggest to implement this via

public enum SignalStrength
{
  Unknown,
  Weak,
  Fair,
  Strong
}

where we simply use the UWP values as provided (1-3 bars), on android we use
public static int calculateSignalLevel (int rssi, Int numLevels)
which "Calculates the level of the signal. This should be used any time a signal is being shown."
on iOS we divide the possible range by 3 and have the corresponding enum value.
If this would be accebtable I would be willing to implement this and have it assigned to me.

pictos:
@Redth  , this API can be very simple ... We can expose an int property for the force signal, in the range 0 to 5. 

```csharp
 public static int SignalStrenght => PlataformSignal();
```

### Android
Need permission of ACCESS_WIFI_STATE

### iOS
I did the implementation, but I don't have a MAC so if any one can test I will be grateful.

### UWP
I test in PC, and don't need extra configuration.


[...]

PR #616, opened 2018-11-10
[WIP] Implement Wifi signal strength

Mrnikbobjeff:
### Description of Change ###
Implements #480 
### API Changes ###
Added: 
 
-  SignalStrength Connectivity.WifiSignalStrength{ get; } 

public enum SignalStrength
{
  Unknown,
  None, 
  Weak,
  Fair,
  Strong
}

### PR Checklist ###

- [] Has tests (if omitted, state reason in description)


Mrnikbobjeff:
Samples are available, I will further test this code tomorrow. It is still not quite complete and thus should not yet be reviewed.

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -21,6 +22,20 @@ static void StartListeners()
             Platform.AppContext.RegisterReceiver(conectivityReceiver, new IntentFilter(ConnectivityManager.ConnectivityAction));
         }
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            var wifiManager = Platform.WifiManager;
```
jamesmontemagno: Need to wrap this in a try/catch and need to ensure that we have the WiFi permission set else it will throw. See other example in here.

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -21,6 +22,20 @@ static void StartListeners()
...
+            if (wifiManager is null)
+                return SignalStrength.Unknown;
+
+            var info = wifiManager?.ConnectionInfo;
+            if (info is null)
```
jamesmontemagno: We prefer to use == for null checks.

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -21,6 +22,20 @@ static void StartListeners()
             Platform.AppContext.RegisterReceiver(conectivityReceiver, new IntentFilter(ConnectivityManager.ConnectivityAction));
         }
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            var wifiManager = Platform.WifiManager;
+            if (wifiManager is null)
```
jamesmontemagno: This is redundant as you do wifiManager?. below so you can remove this

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
```diff
@@ -13,6 +14,11 @@ static void StartListeners()
             listener.ReachabilityChanged += OnConnectivityChanged;
         }
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            return (SignalStrength)(int)(new NEHotspotNetwork().SignalStrength * 3) + 1;
```
jamesmontemagno: I don't think a case like this is good, i would prefer to have a switch case. Same with Android above.

Xamarin.Essentials/Connectivity/Connectivity.shared.cs
```diff
@@ -65,6 +67,15 @@ static void OnConnectivityChanged(ConnectivityChangedEventArgs e)
         }
     }
 
+    public enum SignalStrength
+    {
```
jamesmontemagno: We set the values on each to = 0, = 1

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: Acutally I would rather remove the ?. as this check is necessary to return unknown value if we do not have a wifi Manager instance available if that were acceptable

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: okay :)

Xamarin.Essentials/Connectivity/Connectivity.shared.cs
Mrnikbobjeff: done

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
Mrnikbobjeff: done

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: done

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
```diff
@@ -13,6 +14,22 @@ static void StartListeners()
...
+            switch (signalstrength)
+            {
+                case 1:
+                    return SignalStrength.Weak;
+                case 2:
+                    return SignalStrength.Fair;
+                case 3:
+                    return SignalStrength.Strong;
+                default:
+                    throw new InvalidOperationException("Exhaustive switch statement hit default value.");
```
mattleibow: Maybe rather than throwing, since this comes from the platform and we have no control, just return unknown?

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -21,6 +22,38 @@ static void StartListeners()
...
+                switch (WifiManager.CalculateSignalLevel(info.Rssi, 3))
+                {
+                    case 1:
+                        return SignalStrength.Weak;
+                    case 2:
+                        return SignalStrength.Fair;
+                    case 3:
+                        return SignalStrength.Strong;
+                    default:
+                        throw new InvalidOperationException("Exhaustive switch statement hit default value.");
```
mattleibow: Maybe rather than throwing, since this comes from the platform and we have no control, just return unknown?

Xamarin.Essentials/Connectivity/Connectivity.shared.cs
```diff
@@ -9,12 +9,14 @@ public static partial class Connectivity
...
 
         // a cache so that events aren't fired unnecessarily
-        // this is mainly an issue on Android, but we can stiil do this everywhere
+        // this is mainly an issue on Android, but we can still do this everywhere
         static NetworkAccess currentAccess;
         static List<ConnectionProfile> currentProfiles;
 
         public static NetworkAccess NetworkAccess => PlatformNetworkAccess;
 
+        public static SignalStrength WiFiSignalStrength => PlatformSignalStrength();
```
mattleibow: Should we be caching this as well to avoid unnecessary events?

Xamarin.Essentials/Connectivity/Connectivity.uwp.cs
```diff
@@ -15,6 +16,18 @@ public static partial class Connectivity
         static void StopListeners() =>
              NetworkInformation.NetworkStatusChanged -= NetworkStatusChanged;
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            var signalStrength = NetworkInformation.GetConnectionProfiles()
+                .Where(profile => profile != null && profile.IsWlanConnectionProfile)
+                .Select(profile => profile.GetSignalBars())
+                .Where(signalBar => signalBar.HasValue)
+                .OrderBy(strength => strength).FirstOrDefault();
```
mattleibow: Total styling preference here, but the fact that each line is a separate method, except for the FirstOrDefault is freaking me out a bit :) OCD guy here.

Xamarin.Essentials/Connectivity/Connectivity.uwp.cs
```diff
@@ -15,6 +16,18 @@ public static partial class Connectivity
...
+
+            // Adding 1 to return value because 0 is Unknown and 1 is None. If we have a connection we receive at least 1 in return value
+            return signalStrength.HasValue ? (SignalStrength)signalStrength.Value + 1 : SignalStrength.Unknown;
```
mattleibow: According to the [docs](https://docs.microsoft.com/en-us/uwp/api/windows.networking.connectivity.connectionprofile.getsignalbars#Windows_Networking_Connectivity_ConnectionProfile_GetSignalBars) the range returned from the platform is **0 - 5**, we need to handle those cases since the enum is basically 1-4.

I am thinking: 0 = none, 1 = weak, 2-3 = fair, 4-5 = strong

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
```diff
@@ -13,6 +14,22 @@ static void StartListeners()
             listener.ReachabilityChanged += OnConnectivityChanged;
         }
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            var signalstrength = (int)new NEHotspotNetwork().SignalStrength * 3;
```
mattleibow: According to the docs, the range is **0.0 - 1.0**. This formula is mostly ok, but really looses the potential for 0.99 to be marked as strong because you cast to int.

I am thinking that maybe just do: 0 = none, 0 - 0.3 = weak, 0.3 - 0.7 = fair and 0.7+ = strong.

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -21,6 +22,38 @@ static void StartListeners()
...
+            {
+                var wifiManager = Platform.WifiManager;
+                if (wifiManager == null)
+                    return SignalStrength.Unknown;
+
+                var info = wifiManager.ConnectionInfo;
+                if (info == null)
+                    return SignalStrength.None;
+
+                switch (WifiManager.CalculateSignalLevel(info.Rssi, 3))
```
mattleibow: Looking at the docs, this function appears to return 0 - num-1, so in this case 0 - 2 and will never be strong.

I am thinking that we should do 5 levels: 0 = none, 1 = weak, 2 = fair, 3-4 = strong
I am going for 3-4 as strong as I don't think it is ever the case where the signal strength is at max, so will technically be 0.99 or something. Need to test that theory though.

docs/en/Xamarin.Essentials/SignalStrength.xml
```diff
@@ -0,0 +1,103 @@
...
+      <AssemblyInfo>
+        <AssemblyName>Xamarin.Essentials</AssemblyName>
+        <AssemblyVersion>1.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>Xamarin.Essentials.SignalStrength</ReturnType>
+      </ReturnValue>
+      <MemberValue>2</MemberValue>
+      <Docs>
+        <summary>Bad signal strength.</summary>
```
mattleibow: Maybe we should use the word "Weak" here as bad sounds like there is a error. This just means that the device is really far away or there is interference.

mattleibow:
I see you said that you are still working on this PR and should not be reviewed 😉 But we are all eager beavers here 😁 

I hope our comments are helpful. Thanks for taking the time to work on this!

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
Mrnikbobjeff: Here I disagree. If the platform spec changes our results become worthless and the exception helps us find it. If our logic changes the switch statement should be adapted to be exhaustive. This exception should never be hit ever on iOS

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: I changed it on android as well as uwp to write a debug message and return unknown if it encounters an invalid value in the range provided

Xamarin.Essentials/Connectivity/Connectivity.shared.cs
Mrnikbobjeff: Which events could be triggered by this? I am mostly worried about iOS because I allocate a new NEHotspotNetwork on every call.

Xamarin.Essentials/Connectivity/Connectivity.uwp.cs
Mrnikbobjeff: Understandable, changed it :)

Xamarin.Essentials/Connectivity/Connectivity.uwp.cs
Mrnikbobjeff: Thank you for taking the time to look this up, as it was still a WIP i did not get to sift through all the docs and was not able to test very much 👍 I changed it to your spec.

docs/en/Xamarin.Essentials/SignalStrength.xml
Mrnikbobjeff: I agree with you, but I really dislike docs where the explanation for a value uses the thing it describes itself. But for this it should be self evident. Changed the docs :)

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: Yeah I am also not sure about when this will be max. I'll test this first, but for the time being I used your partitioning.

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
Mrnikbobjeff: Yeah, I did not put the Math.Round call in before the int cast. I agree with your splitting of the range, but as we are working with double values I changed everything below 0.01 to be None. This should be fine as well and lets us keep our nice math :)

Mrnikbobjeff:
> I see you said that you are still working on this PR and should not be reviewed 😉 But we are all eager beavers here 😁
> 
> I hope our comments are helpful. Thanks for taking the time to work on this!

Thank you for taking time to already look at this :) Yes, they were very helpful thank you :)

Xamarin.Essentials/Connectivity/Connectivity.android.cs
Mrnikbobjeff: Finished testing, currently all available values are reachable if you have the equivalent signal strength on android

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
Mrnikbobjeff: Well the question would be whether we would rather return senseless values on an api change or throw an Exception. I believe this to be a stable api though, so this should be dead code´.

Mrnikbobjeff:
I got around to testing android and uwp where this works just fine for me. Since I recently switched jobs I did not set up a developer account for iOS yet. If anyone would like to verify that it works that would be great. Otherwise I would love further discussion of the unresolved comments :)

DeviceTests/DeviceTests.Android/Properties/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0.1.0" package="com.xamarin.essentials.devicetests" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="19" android:targetSdkVersion="27" />
+	<uses-sdk android:minSdkVersion="16" android:targetSdkVersion="27" />
```
Redth: Please revert this change to:

```suggestion
	<uses-sdk android:minSdkVersion="19" android:targetSdkVersion="27" />
```

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
```diff
@@ -13,6 +14,26 @@ static void StartListeners()
...
+            {
+                case 0:
+                case 1:
+                    return SignalStrength.Weak;
+                case 2:
+                    return SignalStrength.Fair;
+                case 3:
+                    return SignalStrength.Strong;
+                default:
+                    throw new InvalidOperationException("Exhaustive switch statement hit default value.");
```
Redth: I agree with @mattleibow we should return unknown here.

```suggestion
                    return SignalStrength.Unknown;
```

Xamarin.Essentials/Connectivity/Connectivity.shared.cs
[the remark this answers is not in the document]
Mrnikbobjeff: Well Wi-Fi would not work, and that is the [official spelling](https://en.wikipedia.org/wiki/Wi-Fi). According to the [abbrevation docs](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)) we should probably use Wifi as suggested.

Redth:
```suggestion
        public static SignalStrength WifiSignalStrength => PlatformSignalStrength();
```

docs/en/FrameworksIndex/xamarin-essentials-android.xml
```diff
@@ -117,17 +117,13 @@
       <Member Id="E:Xamarin.Essentials.Connectivity.ConnectivityChanged" />
       <Member Id="P:Xamarin.Essentials.Connectivity.NetworkAccess" />
       <Member Id="P:Xamarin.Essentials.Connectivity.Profiles" />
+      <Member Id="P:Xamarin.Essentials.Connectivity.WiFiSignalStrength" />
```
Redth:
```suggestion
      <Member Id="P:Xamarin.Essentials.Connectivity.WifiSignalStrength" />
```

docs/en/FrameworksIndex/xamarin-essentials-ios.xml
```diff
@@ -118,17 +118,13 @@
       <Member Id="E:Xamarin.Essentials.Connectivity.ConnectivityChanged" />
       <Member Id="P:Xamarin.Essentials.Connectivity.NetworkAccess" />
       <Member Id="P:Xamarin.Essentials.Connectivity.Profiles" />
+      <Member Id="P:Xamarin.Essentials.Connectivity.WiFiSignalStrength" />
```
Redth:
```suggestion
      <Member Id="P:Xamarin.Essentials.Connectivity.WifiSignalStrength" />
```

docs/en/FrameworksIndex/xamarin-essentials-uwp.xml
```diff
@@ -117,17 +117,13 @@
       <Member Id="E:Xamarin.Essentials.Connectivity.ConnectivityChanged" />
       <Member Id="P:Xamarin.Essentials.Connectivity.NetworkAccess" />
       <Member Id="P:Xamarin.Essentials.Connectivity.Profiles" />
+      <Member Id="P:Xamarin.Essentials.Connectivity.WiFiSignalStrength" />
```
Redth:
```suggestion
      <Member Id="P:Xamarin.Essentials.Connectivity.WifiSignalStrength" />
```

docs/en/FrameworksIndex/xamarin-essentials.xml
```diff
@@ -117,17 +117,13 @@
       <Member Id="E:Xamarin.Essentials.Connectivity.ConnectivityChanged" />
       <Member Id="P:Xamarin.Essentials.Connectivity.NetworkAccess" />
       <Member Id="P:Xamarin.Essentials.Connectivity.Profiles" />
+      <Member Id="P:Xamarin.Essentials.Connectivity.WiFiSignalStrength" />
```
Redth:
```suggestion
      <Member Id="P:Xamarin.Essentials.Connectivity.WifiSignalStrength" />
```

docs/en/Xamarin.Essentials/Connectivity.xml
```diff
@@ -79,5 +79,22 @@
         </remarks>
       </Docs>
     </Member>
+    <Member MemberName="WiFiSignalStrength">
```
Redth:
```suggestion
    <Member MemberName="WifiSignalStrength">
```

docs/en/Xamarin.Essentials/Connectivity.xml
```diff
@@ -79,5 +79,22 @@
...
+      <MemberSignature Language="C#" Value="public static Xamarin.Essentials.SignalStrength WiFiSignalStrength { get; }" />
```
Redth:
```suggestion
      <MemberSignature Language="C#" Value="public static Xamarin.Essentials.SignalStrength WifiSignalStrength { get; }" />
```

docs/en/Xamarin.Essentials/Connectivity.xml
```diff
@@ -79,5 +79,22 @@
...
+      <MemberSignature Language="ILAsm" Value=".property valuetype Xamarin.Essentials.SignalStrength WiFiSignalStrength" />
```
Redth:
```suggestion
      <MemberSignature Language="ILAsm" Value=".property valuetype Xamarin.Essentials.SignalStrength WifiSignalStrength" />
```

docs/en/Xamarin.Essentials/Connectivity.xml
```diff
@@ -79,5 +79,22 @@
...
+      <MemberSignature Language="DocId" Value="P:Xamarin.Essentials.Connectivity.WiFiSignalStrength" />
```
Redth:
```suggestion
      <MemberSignature Language="DocId" Value="P:Xamarin.Essentials.Connectivity.WifiSignalStrength" />
```

docs/en/Xamarin.Essentials/Launcher.xml
```diff
@@ -98,6 +98,7 @@
       </Parameters>
       <Docs>
         <param name="uri">uri scheme</param>
+
```
Redth:
```suggestion
```

Samples/Samples/ViewModel/ConnectivityViewModel.cs
```diff
@@ -4,23 +4,13 @@ namespace Samples.ViewModel
...
-        {
-            get
-            {
-                var profiles = string.Empty;
-                foreach (var p in Connectivity.Profiles)
-                    profiles += "\n" + p.ToString();
-                return profiles;
-            }
-        }
+        public string SignalStrength => Connectivity.WiFiSignalStrength.ToString();
```
Redth:
```suggestion
        public string SignalStrength => Connectivity.WifiSignalStrength.ToString();
```

mattleibow:
I see @Redth discovered the suggestions feature ;)

Xamarin.Essentials/Connectivity/Connectivity.ios.cs
```diff
@@ -13,6 +14,26 @@ static void StartListeners()
             listener.ReachabilityChanged += OnConnectivityChanged;
         }
 
+        static SignalStrength PlatformSignalStrength()
+        {
+            var doubleSignalStrength = new NEHotspotNetwork().SignalStrength;
```
mattleibow: This allocation. I think it should be OK, but maybe wrap it in a using to dispose sooner?

Mrnikbobjeff: Yeah, I was worried about this as well. I was unable to decide whether to take the permanent memory hit and just lazy initialize a static NEHotspotNetwork which just sticks around forever once SignalStrength is used or allocate a new object on every call. This probably depends on the usage scenario as well, as if this were called in a rather tight loop or rather frequently would just add unnecessary memory pressure. If we decide to not cache the value we should definitely dispose the object ASAP to let i be destroyed on iOS.

Issue #454, opened 2018-08-07
Device Information : Storage Info


jrahma:
Hi,

Please add an option in Device to get the Device total space and free space.

Thanks,
Jassim

> VS bug [#735665](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735665)

mattleibow:
I have not really looks at the implementation for this, but we need to make sure that each platform can do this first.

 - Can we distinguish between SD card (external, removable storage), external/non-system storage (embedded, but not the OS) and internal storage (the OS partition)? I know Android has multiple locations that may differ with permissions or other criteria.
 - Can we access the space used, free space, total space?

craylward:
This would be very useful!

https://forums.xamarin.com/discussion/42727/working-with-disk-space-info-in-forms

pictos:
@mattleibow it's possible. I made a simple test using dependency service and it's work fine in Android and UWP, as I don't haven't a Mac i don't testing in iOS, but is implemented. If you want i can implement this in Essentials.
Implementation:
https://github.com/pictos/Dependency

Redth:
Would love someone from the community to make an API spec for this and figure out how to deal with the points @mattleibow pointed out (how to handle different storage devices and/or os partitions, etc).

pictos:
@Redth i made a example implementation, using dependency service, to get free storage in device... I answer the questions in next post

https://github.com/pictos/Dependency (the project)

pictos:
_Can we distinguish between SD card (external, removable storage), external/non-system storage (embedded, but not the OS) and internal storage (the OS partition)? I know Android has multiple locations that may differ with permissions or other criteria._

Yes, we can distinguish the external and internal storage, using the path of folder. In each plataform is like that:

**Android**

```csharp
var fullExternalStorage = Android.OS.Environment.ExternalStorageDirectory.TotalSpace; 
var freeExternalStorage = Android.OS.Environment.ExternalStorageDirectory.UsableSpace;

var fullInternalStorage = Android.OS.Environment.RootDirectory.TotalSpace;
var freeInternalStorage = Android.OS.Environment.RootDirectory.UsableSpace;
```

in Android we have "Android.OS.Environment.IsExternalStorageEmulated" and "Android.OS.Environment.IsExternalStorageRemovable".

With multiple locations we can use the StatFs, when we use a path.

(https://github.com/pictos/Dependency/blob/master/Dependency/Dependency.Android/Services/StorageAndroid.cs)

[...]

PR #617, opened 2018-11-12
GH-454 Implement File System Information

Mrnikbobjeff:
### Description of Change ###
Implemented Storage information #454 
Still WIP, docs are generated but not filled out yet, iOS was not tested yet.
### API Changes ###

Added: 
 
    public struct StorageInfo
    {
        public ulong Capacity { get; }

        public ulong FreeBytes { get; }

        public ulong UsedBytes { get; }

        public StorageType Type { get; }
    }

    public enum StorageType
    {
        Unknown = 0,
        Internal = 1,
        External = 2
    }

Changed:

 - `Task<List<StorageInfo>> DeviceInfo.GetStorageInformationAsync()`

### Behavioral Changes ###
Once finalised users will have to update their uwp manifest if they want to acesss external drives with permissions.

### PR Checklist ###


Mrnikbobjeff:
What I noticed while testing is that on UWP disk readers and empty sd card readers are also displayed but have a capacity and free space of zero and are read-only. We should probably remove these devices as there is no way to control them from UWP. Wanted more input there though

Mrnikbobjeff:
Any update on this?

newky2k:
@Mrnikbobjeff  can you please take a look at the conflicts on this PR so i can review it and get it merged.

Thanks(and sorry for the delay)

Mrnikbobjeff:
@newky2k Sure, I will get to it this week. Could you perhaps also answer the other unanswered question in this pr about disc readers etc? Would like to do both in one go

mattleibow:
@Mrnikbobjeff I would suggest rather resolving the merge conflicts and then we can merge this. The issue with the empty readers can be resolved later.

But, I am also thinking that rather than fixing that just for UWP, put the logic to remove the zero size  storage in the shared code for all platforms.

Samples/Samples/ViewModel/DeviceInfoViewModel.cs
```diff
@@ -28,12 +36,14 @@ public ScreenMetrics ScreenMetrics
...
 
-        public override void OnAppearing()
+        public async override void OnAppearing()
         {
             base.OnAppearing();
 
             DeviceDisplay.ScreenMetricsChanged += OnScreenMetricsChanged;
             ScreenMetrics = DeviceDisplay.ScreenMetrics;
+            StorageInformation = await DeviceInfo.GetStorageInformationAsync();
+            OnPropertyChanged(nameof(StorageInformation));
```
VladislavAntonyuk: Unnecessary call, it is called in SetProperty

Xamarin.Essentials/Types/StorageExtensions.shared.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static class StorageExtensions
+    {
+        public static double ToKilobytes(this ulong bytes) => bytes / 1024d;
+
+        public static double ToMegabytes(this ulong bytes) => bytes.ToKilobytes() / 1024d;
+
+        public static double ToGigabytes(this ulong bytes) => bytes.ToKilobytes() / 1024d / 1024d;
+
+        public static double FreePercentage(this StorageInfo info) => (double)info.FreeBytes / info.Capacity;
```
VladislavAntonyuk: I expect percentage to be 0-100 rather than 0-1

Xamarin.Essentials/DeviceInfo/DeviceInfo.ios.tvos.watchos.cs
```diff
@@ -62,6 +65,16 @@ static DeviceIdiom GetIdiom()
 #endif
         }
 
+        static Task<List<StorageInfo>> PlatformGetStorageInformation()
+        {
+            var attributes = NSFileManager.DefaultManager.GetFileSystemAttributes(Environment.GetFolderPath(Environment.SpecialFolder.Personal));
+            return Task.FromResult(
+                new List<StorageInfo>
+                {
+                    new StorageInfo(attributes.Size, attributes.FreeSize, attributes.Size - attributes.FreeSize, StorageType.Internal)
```
mattleibow: Does iOS not have any external storage options?

Xamarin.Essentials/DeviceInfo/DeviceInfo.uwp.cs
```diff
@@ -68,5 +75,23 @@ static DeviceType GetDeviceType()
...
+            {
+                var folderProps = await folder.Properties.RetrievePropertiesAsync(properties);
+                var fCapacity = (ulong)folderProps[capacity];
+                var fFree = (ulong)folderProps[freeSpace];
+                storageInfos.Add(new StorageInfo(fCapacity, fFree, fCapacity - fFree, StorageType.External));
+            }
+            var localProps = await ApplicationData.Current.LocalFolder.Properties.RetrievePropertiesAsync(properties);
+            var localCapacity = (ulong)localProps[capacity];
+            var localFree = (ulong)localProps[freeSpace];
+            storageInfos.Add(new StorageInfo(localCapacity, localFree, localCapacity - localFree, StorageType.External));
```
mattleibow: Should this be internal?

```suggestion
            storageInfos.Add(new StorageInfo(localCapacity, localFree, localCapacity - localFree, StorageType.Internal));
```

Xamarin.Essentials/DeviceInfo/DeviceInfo.ios.tvos.watchos.cs
mattleibow: Does iOS not have any external storage options?

Xamarin.Essentials/Types/StorageExtensions.shared.cs
```diff
@@ -0,0 +1,15 @@
...
+using Xamarin.Essentials;
+
+namespace Xamarin.Essentials.Types
+{
+    public static class StorageExtensions
+    {
+        public static byte FreePercentage(this StorageInfo info) =>
+            Convert.ToByte(Math.Round(FreePercentage(info.FreeBytes, info.Capacity) * 100));
+
+        public static byte UsedPercentage(this StorageInfo info) => Convert.ToByte(Math.Round(1 - FreePercentage(info.FreeBytes, info.Capacity)));
```
mattleibow: I am not sure if these _really_ add value... It is truly a single division operation.

Xamarin.Essentials/Types/StorageExtensions.shared.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Xamarin.Essentials;
+
+namespace Xamarin.Essentials.Types
+{
+    public static class StorageExtensions
```
mattleibow: I am not sure if these _really_ add value... It is truly a single division operation.

Xamarin.Essentials/Types/StorageExtensions.shared.cs
Mrnikbobjeff: Well I thought about which values to present. If we do not do this it will always round down, now it will return 1 if 0.9% of space is free.

Xamarin.Essentials/DeviceInfo/DeviceInfo.ios.tvos.watchos.cs
Mrnikbobjeff: As far as I am aware it does not offer native external storage solutions.

Mrnikbobjeff:
I finished up the docs and moved the namespaces around after I realised that the Types namespace is gone now. I also added a Tizen implementation which would need to be verified as I can not get the emulator to work on my system. It is based on the docs and should work as it is the simplest implementation from all platforms. It should be all finished up and ready for you to take a final look. The removal of the percentage functions can happen very quickly.

Xamarin.Essentials/Types/StorageExtensions.shared.cs
mattleibow: I actually meant the entire method. I feel that this operation is _OK_ but not particularly useful as I can do `si.Free / si.Total` in my own code.

Xamarin.Essentials/DeviceInfo/DeviceInfo.uwp.cs
```diff
@@ -29,7 +36,7 @@ static string GetVersionString()
                 var v2 = (v & 0x0000FFFF00000000L) >> 32;
                 var v3 = (v & 0x00000000FFFF0000L) >> 16;
                 var v4 = v & 0x000000000000FFFFL;
-                return $"{v1}.{v2}.{v3}.{v4}";
+                return $"{v1.ToString()}.{v2.ToString()}.{v3.ToString()}.{v4.ToString()}";
```
mattleibow: What is the benefit of this change? It is already doing a ToString implicitly.

mattleibow:
I am not a fan of the extensions methods, so I would say just remove them and see if anyone complains. If we ship and we find that it was a big feature that was missing, we can add it back.

Xamarin.Essentials/DeviceInfo/DeviceInfo.shared.cs
```diff
@@ -21,6 +25,53 @@ public static partial class DeviceInfo
...
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
+                return false;
+            return obj is StorageInfo info && Equals(info);
+        }
+
+        public override int GetHashCode() => (Capacity, UsedBytes, FreeBytes, Type).GetHashCode();
+
+        public override string ToString() => $"Capacity: {Capacity.ToString()}{Environment.NewLine}FreeBytes: {FreeBytes.ToString()}{Environment.NewLine}UsedBytes: {UsedBytes.ToString()}{Environment.NewLine}Type: {Type}";
```
mattleibow: I would follow the pattern here with no newlines, comma separated sections:

https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/Types/DisplayInfo.shared.cs#L46-L49

Xamarin.Essentials/DeviceInfo/DeviceInfo.uwp.cs
Mrnikbobjeff: This way you avoid the box. String.format accepts the object type as parameter. If you pass a primitive it will first create a box and then call rostering on the box. This way we avoid the box.

Mrnikbobjeff: This way you avoid the box. String.format accepts the object type as parameter. If you pass a primitive it will first create a box and then call rostering on the box. This way we avoid the box.

mattleibow: Interesting. I had a look to see what the IL actually was and it is pretty interesting.

```csharp
int v1 = 123;
Console.WriteLine($"message: {v1} :end");
Console.WriteLine($"message: {v1.ToString()} :end");
```

```csharp
Console.WriteLine(string.Format("message: {0} :end", num));
Console.WriteLine("message: " + num.ToString() + " :end");
```

https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQaAMOaAzAAQDeOJVJaq5l1jAbgIYBOJLJAvCQCQAiFmXQBfFgIDcDRlVYcARj34CFIgHQAVAPYBlYGzAA7AOYAKAJSiFUmSVF3kpZOgBsNACwkAgpfrZZanlOZUFhMWFUcTIicVsA6gcEqicaN08SACE/CmTZYKVeQTV0LT0DY3MrNVQy/UNTS2sYuorG6vjGJKSgA

It goes from boxing, to string allocations. Not sure which is better at this point. Let me do some research... (aka: ask on Twitter 😉)

Mrnikbobjeff: Well the string allocations are obvious, they will incur either way. The string format method will call tostring themselves. If you want some guidance on that, take a look at the heap allocation Roslyn analyzer. They have a rule for this case

Mrnikbobjeff: Here, for further reference the place the actual object.tostring call happens: https://github.com/dotnet/coreclr/blob/a9f3fc16483eecfc47fb79c362811d870be02249/src/System.Private.CoreLib/shared/System/Text/StringBuilder.cs#L1795
This is called from string format here: https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/Common/src/CoreLib/System/String.Manipulation.cs#L528

Mrnikbobjeff: Here, for further reference the place the actual object.tostring call happens: https://github.com/dotnet/coreclr/blob/a9f3fc16483eecfc47fb79c362811d870be02249/src/System.Private.CoreLib/shared/System/Text/StringBuilder.cs#L1795
This is called from string format here: https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/Common/src/CoreLib/System/String.Manipulation.cs#L528

Mrnikbobjeff: Okay, after some benchmarking I found out that 4 parameters are a breakpoint (There are overloads for up to three parameters, thus there is no array allocation for the params array). If we were able to guarantee that all numbers being stringified are interned (possible through string.intern) we would be getting a benefit from four parameters onward. I did not know that string.Format has overloads for small parameter packs avoiding the array allocation, the advice about not passing value types to reference type parameters still holds though. In this case I would not use ToString as interning the string definitely brings no noticable benefit, so I reverted the change.

Mrnikbobjeff:
@mattleibow  are you happy with the last changes?

jamesmontemagno:
Thanks for this PR and working with us to try to figure out if it is a good fit. After reviewing with the team we have decided that here there are too many unknowns and inconsistencies between the platforms and that this API is not a great fit to roll into essentials to maintain. For that reason we are going to close it, but really appreciate the work that you did on the PR and is a good reference for people to look at to implement themselves.

PR #642, opened 2018-12-09
[UWP] [XBOX] Gamepad vibration support

JTOne123:
As you know Xamarin apps able to run on Xbox (or on PC with the gamepad) as well.

As for me, Vibration is so important at Xbox and I was frustrated when finding out that Essentials does not support vibration on UWP platform as it can.


Xamarin.Essentials/Vibration/Vibration.uwp.cs
```diff
@@ -1,20 +1,55 @@
...
+using System.Threading.Tasks;
 using Windows.Foundation.Metadata;
+using Windows.Gaming.Input;
 using Windows.Phone.Devices.Notification;
 
 namespace Xamarin.Essentials
 {
     public static partial class Vibration
     {
+        static Gamepad firstGamepad;
```
Mrnikbobjeff: Replacing this with a Lazy<Gamepad> for consistency would be advisable

Xamarin.Essentials/Vibration/Vibration.uwp.cs
```diff
@@ -1,20 +1,55 @@
...
+                if (firstGamepad == null)
+                    firstGamepad = Gamepad.Gamepads.FirstOrDefault();
+
+                return firstGamepad;
+            }
+        }
+
         internal static bool IsSupported
-            => ApiInformation.IsTypePresent("Windows.Phone.Devices.Notification.VibrationDevice") && DefaultDevice != null;
+            => FirstGamepad != null || (ApiInformation.IsTypePresent("Windows.Phone.Devices.Notification.VibrationDevice") && DefaultDevice != null);
```
Mrnikbobjeff: This would have to be ammended as well if the property is a lazy<T>

Xamarin.Essentials/Vibration/Vibration.uwp.cs
```diff
@@ -1,20 +1,55 @@
...
         static VibrationDevice DefaultDevice => VibrationDevice.GetDefault();
 
-        static void PlatformVibrate(TimeSpan duration) =>
-            DefaultDevice.Vibrate(duration);
+        static void PlatformVibrate(TimeSpan duration)
+        {
+            if (FirstGamepad != null)
+            {
+                FirstGamepad.Vibration = new GamepadVibration { RightMotor = 0.85 };
+                Task.Delay(duration).ContinueWith((t) => FirstGamepad.Vibration = new GamepadVibration { });
```
Mrnikbobjeff: Use the default struct constructor here. Also, possibly to indicate that we are not interested in the previous task we could discard it with _ The second is just personal preference though :)

Xamarin.Essentials/Vibration/Vibration.uwp.cs
```diff
@@ -1,20 +1,55 @@
...
+            }
+        }
 
-        static void PlatformCancel() =>
-            DefaultDevice.Cancel();
+        static void PlatformCancel()
+        {
+            if (FirstGamepad != null)
+            {
+                FirstGamepad.Vibration = new GamepadVibration { };
```
Mrnikbobjeff: Default constructor here as well

Xamarin.Essentials/Vibration/Vibration.uwp.cs
```diff
@@ -1,20 +1,55 @@
...
+            {
+                DefaultDevice.Vibrate(duration);
+            }
+        }
 
-        static void PlatformCancel() =>
-            DefaultDevice.Cancel();
+        static void PlatformCancel()
+        {
+            if (FirstGamepad != null)
```
Mrnikbobjeff: Null check to be amended for lazy value

jamesmontemagno:
I have to think about this and talk with the team as this introduces a level of testing and support that I am not sure we are ready to support officially.

jamesmontemagno:
We discussed this further and I think the correct thing to do is to create an actual "Controller" API that would do multiple things including Vibrate and give more control

The issue here is that it is a bit unpredictable with what motor to rotate, which controller, etc. So a proper spec of what a controller API would look like and how it would be implemented on each platform would be ideal. So closing this out for now and please start a new spec so we can discuss it.

Issue #196, opened 2018-04-23
Add browser customisation APIs


prashantvc:
### Description
Essentials Browser does not have API to customise the Browser title and colour.

The plugins has `BrowserOption` class that we can pass when opening a browser object, but there is no alternative for this in the Essentials API

Example: 

```csharp
await CrossShare.Current.OpenBrowser("https://auth.xamarin.com/account/register",
    new BrowserOptions
    {
        ChromeShowTitle = true,
        ChromeToolbarColor = new ShareColor
            {
                A=255,
                R=118,
                G=53,
                B=235
            },
        UseSafariReaderMode = false,
        UseSafariWebViewController = true
});
```

> VS bug [#735673](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735673)

mattleibow:
This may be done by adding a settings type and have properties like `PreferredTitle` and `PreferredTint` or something. UWP does not have a browser tab thing, but does iOS allow for colors or a title.

Evolutionlab:
The old "Share" plugin supported customized colors. Imho you should support that:

```
                await CrossShare.Current.OpenBrowser(uri,
                    new BrowserOptions
                    {
                        SafariBarTintColor     = new ShareColor(0, 73, 138),
                        SafariControlTintColor = new ShareColor(255, 255, 255),
                        ChromeToolbarColor     = new ShareColor(0, 73, 138)
                    });
```

Redth:
I think we need to shy away from adding properties with names specific to one platform.

I think it would be ok to have something like @mattleibow suggested and have `PreferredBackgroundColor` and `PreferredControlColor` with some documentation around what it means on each platform.

jamesmontemagno:
So, System.Drawing.Color is not available in Xamarin.iOS, but instead OpenTK, so for now this work will be delayed until that is fixed. For tracking here: https://github.com/xamarin/xamarin-macios/issues/5273

Mrnikbobjeff:
This can be closed AFAIK

xamarin-release-manager:
[VS sync] The field 'Milestone' contains the value '1.1.0' that is not in the list of supported values

PR #646, opened 2018-12-11
GH-196 Browser Customization

Mrnikbobjeff:
### Description of Change ###
Implemented some Features required for GH- #196. 
### API Changes ###

Added: 

```csharp
public readonly struct EssentialsColor : IEquatable<EssentialsColor>
{
        public byte R { get; }
        public byte G { get; }
        public byte B { get; }
        public byte A { get; }
}

public readonly partial struct BrowserLaunchOptions : IEquatable<BrowserLaunchOptions>
{
        public EssentialsColor? PreferredTitleColor { get; }

        public EssentialsColor? PreferredBackgroundColor { get; }

        public EssentialsColor? PrefferedControlColor { get; }

        public bool? ShouldShowTitle { get; }
}
```
Changed:
 - `Browser.OpenAsync(string uri, BrowserLaunchOptions options = default) // Added default parameter`

### Behavioral Changes ###

None. All values are nullable. If they are not set as they are by default no changes are made.
### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Browser/Browser.android.cs
+++ b/Xamarin.Essentials/Browser/Browser.android.cs
@@ -10,21 +10,21 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options)
+        static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchOptions options)
         {
             var nativeUri = AndroidUri.Parse(uri.AbsoluteUri);
 
-            switch (launchMode)
+            switch (options.LaunchMode)
             {
                 case BrowserLaunchMode.SystemPreferred:
                     var tabsBuilder = new CustomTabsIntent.Builder();
                     tabsBuilder.SetShowTitle(true);
                     if (options.PreferredTitleColor.HasValue)
-                        tabsBuilder.SetToolbarColor((int)options.PreferredTitleColor.Value);
+                        tabsBuilder.SetToolbarColor((int)options.PreferredTitleColor.Value.ToInt());
                     if (options.PrefferedControlColor.HasValue)
-                        tabsBuilder.SetSecondaryToolbarColor((int)options.PrefferedControlColor.Value);
-                    if (options.ShouldShowTitle.HasValue)
-                        tabsBuilder.SetShowTitle(options.ShouldShowTitle.Value);
+                        tabsBuilder.SetSecondaryToolbarColor((int)options.PrefferedControlColor.Value.ToInt());
+                    if (options.TitleMode != BrowserTitleMode.Default)
+                        tabsBuilder.SetShowTitle(options.TitleMode == BrowserTitleMode.Show);
                     var tabsIntent = tabsBuilder.Build();
                     tabsIntent.Intent.SetFlags(ActivityFlags.ClearTop);
                     tabsIntent.Intent.SetFlags(ActivityFlags.NewTask);
@@ -44,10 +44,5 @@ namespace Xamarin.Essentials
 
             return Task.FromResult(true);
         }
-    }
-
-    public readonly partial struct EssentialsColor
-    {
-        public static explicit operator int(EssentialsColor x) => x.A | x.R << 8 | x.G << 16 | x.B << 24;
     }
 }
--- a/Xamarin.Essentials/Browser/Browser.android.cs
+++ b/Xamarin.Essentials/Browser/Browser.android.cs
@@ -10,7 +10,7 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
+        static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options)
         {
             var nativeUri = AndroidUri.Parse(uri.AbsoluteUri);
 
@@ -19,6 +19,12 @@ namespace Xamarin.Essentials
                 case BrowserLaunchMode.SystemPreferred:
                     var tabsBuilder = new CustomTabsIntent.Builder();
                     tabsBuilder.SetShowTitle(true);
+                    if (options.PreferredTitleColor.HasValue)
+                        tabsBuilder.SetToolbarColor((int)options.PreferredTitleColor.Value);
+                    if (options.PrefferedControlColor.HasValue)
+                        tabsBuilder.SetSecondaryToolbarColor((int)options.PrefferedControlColor.Value);
+                    if (options.ShouldShowTitle.HasValue)
+                        tabsBuilder.SetShowTitle(options.ShouldShowTitle.Value);
                     var tabsIntent = tabsBuilder.Build();
                     tabsIntent.Intent.SetFlags(ActivityFlags.ClearTop);
                     tabsIntent.Intent.SetFlags(ActivityFlags.NewTask);
@@ -38,5 +44,10 @@ namespace Xamarin.Essentials
 
             return Task.FromResult(true);
         }
+    }
+
+    public readonly partial struct EssentialsColor
+    {
+        public static explicit operator int(EssentialsColor x) => x.A | x.R << 8 | x.G << 16 | x.B << 24;
     }
 }
--- a/Xamarin.Essentials/Browser/Browser.ios.cs
+++ b/Xamarin.Essentials/Browser/Browser.ios.cs
@@ -8,7 +8,7 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
+        static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options)
         {
             var nativeUrl = new NSUrl(uri.AbsoluteUri);
 
@@ -17,6 +17,12 @@ namespace Xamarin.Essentials
                 case BrowserLaunchMode.SystemPreferred:
                     var sfViewController = new SFSafariViewController(nativeUrl, false);
                     var vc = Platform.GetCurrentViewController();
+
+                    if (options.PreferredTitleColor.HasValue)
+                        sfViewController.PreferredBarTintColor = (UIColor)options.PreferredTitleColor;
+
+                    if (options.PrefferedControlColor.HasValue)
+                        sfViewController.PreferredControlTintColor = (UIColor)options.PrefferedControlColor;
 
                     if (sfViewController.PopoverPresentationController != null)
                     {
@@ -30,5 +36,10 @@ namespace Xamarin.Essentials
 
             return true;
         }
+    }
+
+    public readonly partial struct EssentialsColor
+    {
+        public static explicit operator UIColor(EssentialsColor x) => new UIColor(x.R, x.G, x.B, x.A);
     }
 }
--- a/Xamarin.Essentials/Browser/Browser.ios.cs
+++ b/Xamarin.Essentials/Browser/Browser.ios.cs
@@ -8,21 +8,21 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options)
+        static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchOptions options)
         {
             var nativeUrl = new NSUrl(uri.AbsoluteUri);
 
-            switch (launchMode)
+            switch (options.LaunchMode)
             {
                 case BrowserLaunchMode.SystemPreferred:
                     var sfViewController = new SFSafariViewController(nativeUrl, false);
                     var vc = Platform.GetCurrentViewController();
 
                     if (options.PreferredTitleColor.HasValue)
-                        sfViewController.PreferredBarTintColor = (UIColor)options.PreferredTitleColor;
+                        sfViewController.PreferredBarTintColor = options.PreferredTitleColor.Value.ToPlatformColor();
 
                     if (options.PrefferedControlColor.HasValue)
-                        sfViewController.PreferredControlTintColor = (UIColor)options.PrefferedControlColor;
+                        sfViewController.PreferredControlTintColor = options.PrefferedControlColor.Value.ToPlatformColor();
 
                     if (sfViewController.PopoverPresentationController != null)
                     {
@@ -36,10 +36,5 @@ namespace Xamarin.Essentials
 
             return true;
         }
-    }
-
-    public readonly partial struct EssentialsColor
-    {
-        public static explicit operator UIColor(EssentialsColor x) => new UIColor(x.R, x.G, x.B, x.A);
     }
 }
--- a/Xamarin.Essentials/Browser/Browser.shared.cs
+++ b/Xamarin.Essentials/Browser/Browser.shared.cs
@@ -5,10 +5,10 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        public static Task OpenAsync(string uri) =>
-            OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
+        public static Task OpenAsync(string uri, BrowserLaunchOptions options = default) =>
+            OpenAsync(uri, BrowserLaunchMode.SystemPreferred, options);
 
-        public static Task OpenAsync(string uri, BrowserLaunchMode launchMode)
+        public static Task OpenAsync(string uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options = default)
         {
             if (string.IsNullOrWhiteSpace(uri))
             {
@@ -18,11 +18,11 @@ namespace Xamarin.Essentials
             return OpenAsync(new Uri(uri), launchMode);
         }
 
-        public static Task OpenAsync(Uri uri) =>
-          OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
+        public static Task OpenAsync(Uri uri, BrowserLaunchOptions options = default) =>
+          OpenAsync(uri, BrowserLaunchMode.SystemPreferred, options);
 
-        public static Task<bool> OpenAsync(Uri uri, BrowserLaunchMode launchMode) =>
-            PlatformOpenAsync(EscapeUri(uri), launchMode);
+        public static Task<bool> OpenAsync(Uri uri, BrowserLaunchMode launchMode, BrowserLaunchOptions options = default) =>
+            PlatformOpenAsync(EscapeUri(uri), launchMode, options);
 
         internal static Uri EscapeUri(Uri uri)
         {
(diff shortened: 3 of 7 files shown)


Mrnikbobjeff:
We should really consider dropping .Net Standard 1.x support. System.Drawing.Color had to be manually approximated to have a cross platfrorm argb representation.

jamesmontemagno:
No need for a specific color, we should use System.Drawing.Color

Mrnikbobjeff:
Which is not available in .net standard 1.6, that is why I had to implement this abstraction. System.drawing.color became available in .Net Standard 2.0

jamesmontemagno:
We should limit the API to only .NET Standard 2.0. We do not want to add a new type here.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -5,10 +5,10 @@ namespace Xamarin.Essentials
 {
     public static partial class Browser
     {
-        public static Task OpenAsync(string uri) =>
-            OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
+        public static Task OpenAsync(string uri, BrowserLaunchOptions options = default) =>
```
mattleibow: We must not use default parameters, rather create an overload:

```csharp
Task OpenAsync(string uri) =>
    OpenAsync(uri, null);
Task OpenAsync(string uri, BrowserLaunchMode launchMode) =>
    OpenAsync(uri, new BrowserLaunchOptions(launchMode));
Task OpenAsync(string uri, BrowserLaunchOptions options) =>
    PlatformOpenAsync(uri, options);
```

I don't think there will be an overload that takes both `BrowserLaunchMode` AND `BrowserLaunchOptions` - what is the case where you will set colors and then also say ignore those colors?

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
         }
     }
 
+    public readonly partial struct BrowserLaunchOptions : IEquatable<BrowserLaunchOptions>
```
mattleibow: I think this should be a `class` - options/config are classes, results/returns are structs.

Also, rather than having multiple parameters, add a `BrowserLaunchMode` property here and add another constructor that takes just this:

```csharp
public BrowserLaunchOptions() {
    // preferred to match the default values in the OpenAsync
    BrowserLaunchMode = BrowserLaunchMode.SystemPreferred;
}

public BrowserLaunchOptions(BrowserLaunchMode launchMode) {
    BrowserLaunchMode = launchMode;
}

public BrowserLaunchOptions(EssentialsColor? preferredControlColor, EssentialsColor? preferredBackgroundColor, EssentialsColor? preferredTitleColor, bool? shouldShowTitle) {

    // Default to this as no one should be setting title color and then
    // saying: just open the browser
    // If this really is a case, then they can just set the property manually
    BrowserLaunchMode = BrowserLaunchMode.SystemPreferred;

    PreferredControlColor = preferredControlColor;
    PreferredBackgroundColor = preferredBackgroundColor;
    PreferredTitleColor = preferredTitleColor;
    ShouldShowTitle = shouldShowTitle;
}
```

Take inspiration from this style: https://github.com/xamarin/Essentials/blob/b018723477f848b108cb9a67adecb3608d0ab686/Xamarin.Essentials/Map/MapLaunchOptions.shared.cs

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
...
+    {
+        public BrowserLaunchOptions(EssentialsColor? prefferedControlColor, EssentialsColor? preferredBackgroundColor, EssentialsColor? preferredTitleColor, bool? shouldShowTitle)
+        {
+            PrefferedControlColor = prefferedControlColor;
```
mattleibow:
```suggestion
            PreferredControlColor = preferredControlColor;
```

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
         }
     }
 
+    public readonly partial struct BrowserLaunchOptions : IEquatable<BrowserLaunchOptions>
+    {
+        public BrowserLaunchOptions(EssentialsColor? prefferedControlColor, EssentialsColor? preferredBackgroundColor, EssentialsColor? preferredTitleColor, bool? shouldShowTitle)
```
mattleibow:
```suggestion
        public BrowserLaunchOptions(EssentialsColor? preferredControlColor, EssentialsColor? preferredBackgroundColor, EssentialsColor? preferredTitleColor, bool? shouldShowTitle)
```

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
...
+            PreferredBackgroundColor = preferredBackgroundColor;
+            PreferredTitleColor = preferredTitleColor;
+            ShouldShowTitle = shouldShowTitle;
+        }
+
+        public EssentialsColor? PreferredTitleColor { get; }
+
+        public EssentialsColor? PreferredBackgroundColor { get; }
+
+        public EssentialsColor? PrefferedControlColor { get; }
```
mattleibow:
```suggestion
        public EssentialsColor? PreferredControlColor { get; }
```

Xamarin.Essentials/Browser/Browser.ios.cs
```diff
@@ -18,6 +18,12 @@ static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
                     var sfViewController = new SFSafariViewController(nativeUrl, false);
                     var vc = Platform.GetCurrentViewController();
 
+                    if (options.PreferredTitleColor.HasValue)
+                        sfViewController.PreferredBarTintColor = (UIColor)options.PreferredTitleColor;
+
+                    if (options.PrefferedControlColor.HasValue)
```
mattleibow:
```suggestion
                    if (options.PreferredControlColor.HasValue)
```

Xamarin.Essentials/Browser/Browser.ios.cs
```diff
@@ -18,6 +18,12 @@ static async Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
...
+                        sfViewController.PreferredControlTintColor = (UIColor)options.PrefferedControlColor;
```
mattleibow:
```suggestion
                        sfViewController.PreferredControlTintColor = (UIColor)options.PreferredControlColor;
```

Xamarin.Essentials/Browser/Browser.android.cs
```diff
@@ -19,6 +19,12 @@ static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
                 case BrowserLaunchMode.SystemPreferred:
                     var tabsBuilder = new CustomTabsIntent.Builder();
                     tabsBuilder.SetShowTitle(true);
+                    if (options.PreferredTitleColor.HasValue)
+                        tabsBuilder.SetToolbarColor((int)options.PreferredTitleColor.Value);
+                    if (options.PrefferedControlColor.HasValue)
```
mattleibow:
```suggestion
                    if (options.PreferredControlColor.HasValue)
```

Xamarin.Essentials/Browser/Browser.android.cs
```diff
@@ -19,6 +19,12 @@ static Task<bool> PlatformOpenAsync(Uri uri, BrowserLaunchMode launchMode)
...
+                        tabsBuilder.SetSecondaryToolbarColor((int)options.PrefferedControlColor.Value);
```
mattleibow:
```suggestion
                        tabsBuilder.SetSecondaryToolbarColor((int)options.PreferredControlColor.Value);
```

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
...
+    {
+        public BrowserLaunchOptions(EssentialsColor? prefferedControlColor, EssentialsColor? preferredBackgroundColor, EssentialsColor? preferredTitleColor, bool? shouldShowTitle)
+        {
+            PrefferedControlColor = prefferedControlColor;
+            PreferredBackgroundColor = preferredBackgroundColor;
+            PreferredTitleColor = preferredTitleColor;
+            ShouldShowTitle = shouldShowTitle;
+        }
+
+        public EssentialsColor? PreferredTitleColor { get; }
```
mattleibow: I would suggest getters and setters.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,6 +35,60 @@ internal static Uri EscapeUri(Uri uri)
...
+
+        public bool Equals(BrowserLaunchOptions other) => PreferredTitleColor.Equals(other.PreferredTitleColor)
+                                                          && PreferredBackgroundColor.Equals(other.PreferredBackgroundColor)
+                                                          && PrefferedControlColor.Equals(other.PrefferedControlColor)
+                                                          && ShouldShowTitle.Equals(other.ShouldShowTitle);
+
+        public override int GetHashCode() => (PreferredBackgroundColor, PreferredBackgroundColor, PrefferedControlColor, ShouldShowTitle).GetHashCode();
+    }
+
+    public readonly partial struct EssentialsColor : IEquatable<EssentialsColor>
```
mattleibow: See this comment, but I think we need to discuss this further @redth @jamesmontemagno https://github.com/xamarin/Essentials/pull/646#issuecomment-446466659

Mrnikbobjeff: The TypeForwarding info you gave me was very interesting, I did not know about this feature before. Thank you for that! This indeed makes EssentialsColor obsolete. I will remove this in the next commit.

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: The main difference here is that MapLaunchOptions contains a reference type, thus there is no harm in having it as a reference type as well. But here, we only store structs and bools which in turn are all value types. As the struct is bigger than the size of a reference on all platforms I would rather have it as an In parameter, which would elide copying. Also, I would strongly suggest to keep the immutability of this struct. There should be no reason to modify a BrowserLaunchOptions instance after it was created (at least from a users point I see no use in making it modifyable).

Mrnikbobjeff: I agree with removing the BrowserLaunchMode as a function argument and having it as a meber in the new options struct. Smart solution, I like it!

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: There is none, I will change this in the next commit.

jamesmontemagno:
So, System.Drawing.Color is not available in Xamarin.iOS, but instead OpenTK, so for now this work will be delayed until that is fixed. For tracking here: https://github.com/xamarin/xamarin-macios/issues/5273

mattleibow:
Based on the response:
> Xamarin.iOS does not "conform" to NetStandard in the sense that the Xamarin.iOS.dll is expected to include all the dlls needed for all the NetStandard APIs. Instead Xamarin.iOS provides the Facades with the type forwarders needed to find the actual dlls required for specific types.

I would say this is what we are going to have to do. An we can add the OpenTK-1 reference in the NuGet. It should be "OK" because it is a "system" dll for iOS.

Mrnikbobjeff:
Would it also be possible to type forward to a type forward? So just skip referencing OpenTK and forward Color to the Xamarin.iOS type forward

mattleibow:
@jamesmontemagno I think we can just go ahead and depend on OpenTK for iOS - Xamarin.Forms already references it so we can't really add any more. Sure, Bob the iOS-only developer may not be 100% happy, but it shouldn't affect him that much. And, I don't think anything is going to change anytime soon with iOS and the linker.

jamesmontemagno:
In this regard I am doing some color work and will set it up, so please wait until I have that in a branch for you to pull up.

mattleibow:
NICE! Looking forward to seeing cool things happen in this world!

Mrnikbobjeff:
Hey, so I was doing some work for type forwarding and I would love to have clarification on how I am supposed to implement this. @jamesmontemagno said that this should only be available on netstandard 2.0. I suppose you don't mean ifdefing the entire static class right? If so I am good to go, otherwise some clarification for the shared class would be nice :)

mattleibow:
You are going to have to #if the class and create a dummy type so we don't have to worry about method signatures. Nobody will actually ever get netstandard1.0 in their app - unless they are using PCL. But, I don't think that is actually possible as the new IDEs can't open those projects anymore.

You probably can do something like this:

```csharp
#if NETSTANDARD1_0
// a dummy type for .NET Standard 1.0
public readonly partial struct BrowserLaunchOptions { }
#else
// the real type for all the other platforms
public readonly partial struct BrowserLaunchOptions : … {
    ...
}
#endif
````

@jamesmontemagno does this look fine to you? Not super pretty, but this way we don't have #if all over the show.

Another way might be to create a dummy Color type that is wrapped in a #if. All the code stays in it's beautiful place, but for the netstandard1.0 target, we just include this:

```csharp
#if NETSTANDARD1_0
namespace System.Drawing {
    public struct Color {
    }
}
#endif
```

That may cause issues if someone builds against netstandard1.0, and then uses it in an app. The runtime will look for color in our assembly, but we could also do this:

```csharp
#if NETSTANDARD1_0
namespace System.Drawing {
    public struct Color {
    }
}
#else
[assembly: TypeForwardedTo(typeof(Color))]
#endif
```

That should fix all cases. We don't have to provide any implementation for `Color`, or we can just add the `A` R` G` `B` properties to be friendly... It may need testing, but I can't see any issues as this is what netstandard.dll is built on.

My main concern here is that we are going to write bad code just to satisfy the compilers and things, that is why I am pushing a separate, smaller file that we can just delete when we finally go golden - and nothing breaks anyone.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -1,28 +1,47 @@
...
+
+    public byte R => (byte)(Value >> 16);
+
+    public byte G => (byte)(Value >> 8);
+
+    public byte B => (byte)Value;
+}
+#else
+using System.Drawing;
+[assembly: TypeForwardedTo(typeof(Color))]
```
mattleibow: Rather use the full name instead of the using: `[assembly: TypeForwardedTo(typeof(System.Drawing;.Color))]`

I think there are rules about where usings can go...

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -1,28 +1,47 @@
...
     {
         public static Task OpenAsync(string uri) =>
-            OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
+            OpenAsync(uri, default(BrowserLaunchOptions));
+
+        public static Task OpenAsync(string uri, BrowserLaunchMode launchMode) =>
+            OpenAsync(uri, new BrowserLaunchOptions(launchMode));
 
-        public static Task OpenAsync(string uri, BrowserLaunchMode launchMode)
+        public static Task OpenAsync(string uri, BrowserLaunchOptions options = default)
```
mattleibow: We prefer no default parameters.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -1,28 +1,47 @@
...
-            return OpenAsync(new Uri(uri), launchMode);
+            return OpenAsync(new Uri(uri), options);
         }
 
-        public static Task OpenAsync(Uri uri) =>
-          OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
-
-        public static Task<bool> OpenAsync(Uri uri, BrowserLaunchMode launchMode) =>
-            PlatformOpenAsync(EscapeUri(uri), launchMode);
+        public static Task<bool> OpenAsync(Uri uri, BrowserLaunchOptions options = default) =>
```
mattleibow: We prefere no default parameters.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,9 +54,65 @@ internal static Uri EscapeUri(Uri uri)
         }
     }
 
+    public class BrowserLaunchOptions : IEquatable<BrowserLaunchOptions>
+    {
+        public BrowserLaunchOptions(Color? prefferedControlColor, Color? preferredBackgroundColor, Color? preferredTitleColor, BrowserLaunchMode launchMode, BrowserTitleMode shouldShowTitle)
```
mattleibow: @jamesmontemagno may have an opinion, but I don't think we need this. Users can just use the object initializer syntax: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,9 +54,65 @@ internal static Uri EscapeUri(Uri uri)
...
+        Default = 0,
+        Show = 1,
+        Hide = 2
+    }
+
     public enum BrowserLaunchMode
     {
-        SystemPreferred = 0,
-        External = 1,
+        Default = 0,
```
mattleibow: @jamesmontemagno Do we want a "default" option here? If follows the rules, but is this maybe not useful? I think our original idea that "Default" === "SystemPreferred", but we used a different name to help discoverability? Not sure about this one...

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: Yeah, this is still a remnant from the readonly struct where every parameter can be assigned. I was unsure on whether to leave them in. I'll remove them

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: Stylecop did not complain. If I remove the using the android build fails, not sure why.

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: Done

Xamarin.Essentials/Browser/Browser.shared.cs
jamesmontemagno: I don't think we wan't default here.

SystemPreferred != Default as Default would be launching URI. This is a special case in general and should be the default. I think that we should pass in the LaunchMode as a parameter and LaunchOptions as another. I would hate to create another class just to do systempreferred or external if I don't need any of the options.

Xamarin.Essentials/Browser/Browser.shared.cs
```diff
@@ -35,9 +60,56 @@ internal static Uri EscapeUri(Uri uri)
         }
     }
 
+    public class BrowserLaunchOptions : IEquatable<BrowserLaunchOptions>
```
jamesmontemagno: Any reason this is IEquatable? Doesn't seem like it needs to be at all since it is a class.... should it be a struct?

mattleibow: Probably shod be class with no interfaces. Not sure if this is official, but I go with POCO classes in, read-only structs out...

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: If I might refer to the discussion we had about this being a value or reference type, if this were a value type the object creation and destruction would be negligable. Especially when we consider the use case, this will not be called frequently. Otherwise I can revert to the three parameter style I adopted originally for the internal implementation and just pass null.

Xamarin.Essentials/Browser/Browser.shared.cs
Mrnikbobjeff: See my other comment. While all current features use the style @mattleibow describes I think it would be wortwhile investigating when we should differ from this. Even though this is an async method it could be that a passed value type does not have to be boxed. Then, having this as a value type would IMHO be preferrable due to the lessened heap allocation. All contained properties are value types as well, which is not the case for any other in parameter for the public api.

PR #647, opened 2018-12-12
GH-480 WiFi Signal Strength

jamesmontemagno:
### Description of Change ###

Finalize WiFi signal strength API

### Bugs Fixed ###

- Related to issue #480 

### API Changes ###

Added: 
 
- SignalStrength - SignalStrength Connectivity.WiFiSignalStrength { get; } 

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/DeviceTests/DeviceTests.Shared/Connectivity_Tests.cs
+++ b/DeviceTests/DeviceTests.Shared/Connectivity_Tests.cs
@@ -19,9 +19,16 @@ namespace DeviceTests
         {
             var strength = Connectivity.WiFi.SignalStrength;
             if (Connectivity.ConnectionProfiles.Contains(ConnectionProfile.WiFi))
-                Assert.True(strength != SignalStrength.None && strength != SignalStrength.Unknown);
+            {
+                if (DeviceInfo.Platform == DevicePlatform.iOS)
+                    Assert.True(strength == SignalStrength.Unknown);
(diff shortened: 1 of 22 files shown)


jamesmontemagno:
build

DeviceTests/DeviceTests.Shared/Connectivity_Tests.cs
```diff
@@ -13,5 +13,15 @@ public class Connectivity_Tests
         [Fact]
         public void ConnectionProfiles() =>
             Assert.True(Connectivity.ConnectionProfiles.Count() > 0);
+
+        [Fact]
+        public void WiFiSignale()
```
mattleibow:
```suggestion
        public void WiFiSignal()
```

jamesmontemagno:
So, it looks like iOS is going to cause a big issue here. There is no API in there that returns the active wifi signal. The HotspotHelpers needs special requirements from Apple and you only get back information about your WiFi connection itself if you are managing it. The only option here is to probably put "Unknown" for iOS unless we try to call into a Unix property of some sort.

jamesmontemagno:
After further investigation and discussion on this topic we are not going to add this to Xamarin.Essentials at this time due to no iOS API being available. Since this is a core platform and developers expect each platform supported to work with all APIs out of the box.

Issue #126, opened 2018-03-28
Detect Shake API


mattleibow:
Would we want to have an API that detects if the user shakes the device?

> VS bug [#735682](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735682)

pictos:
We can use the accelerometer to detect this (UWP and Android). For iOS we can override the `MotionEnded ` method (UIView).

mattleibow:
I am thinking we could easily port this API: https://github.com/square/seismic/blob/master/library/src/main/java/com/squareup/seismic/ShakeDetector.java

That is a nice and simple one, and we can do it in a 100% cross-platform way with just the accelerometer. this iOS method, is it really going to produce a better shake detection than just implementing the acceleromenter?

pictos:
Yeah i see this code, and this other: 

https://stackoverflow.com/questions/35075193/accelerometer-shaken-event-on-uwp-not-working-any-solutions

I can try to deploy this and see how you're going to behave

xamarin-release-manager:
[VS sync] The field 'Milestone' contains the value '1.1.0' that is not in the list of supported values

PR #666, opened 2018-12-23
GH-126 Detect Shake API

pictos:
### Description of Change ###

Implement the shake API.

### Bugs Fixed ###
- Related to issue #126 
### API Changes ###
Adding this in DeviceInfo.shared class:

Added: 
 
- double -double accelerationThreshold = 2;

- int - int shakenInterval = 500;

- bool - bool shakenLister;

- DateTime - DateTime shakenTimeSpan = DateTime.Now;

- bool - bool ShakenLister {get; set;}
- event - event EventHandler OnShaked;
- void Accelerometer_ReadingChanged(object sender, AccelerometerChangedEventArgs e);
- double Square(this float q);

### Behavioral Changes ###

Now we can detect when the device is shaken.

### PR Checklist ###

> VS bug [#758225](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/758225)


pictos:
@mattleibow you're welcome. I will change it to Accelerometer class. 

To me the `DeviceShaken` event don't need args, what do you think?

mattleibow:
Probably not. But, is there a way to detect the number of shakes? Maybe there can be a device shaken x times type of vibe... Is there a call for "do something after at least 3 shakes"?

jamesmontemagno:
I am curious if using the android accelerometer messes with compass/accelerometer itself... assumption is yes as far as the speed goes....

jamesmontemagno:
Ahhh reading now. I guess if we combine them together. that makes more sense.... need to well document it.

pictos:
@mattleibow, it's better the dev handle with this conditions. Because the `SensorSpeed`  is a dev choice, so it's easier for him to handle it.
If you run the [example](https://github.com/pictos/Essentials/tree/issue-126), on the accelerometer pages, you will notice that the various behavior according to `SensorSpeed`.

pictos:
@jamesmontemagno  if you want, i can try to update the docs.

pictos:
@mattleibow , i did the changes in project, implement the shake API in Accelerometer class.

I'm not sure if the Github notify's you when a I do some commit, he do?

And happy new year, for all you!

jamesmontemagno:
So, a few things. On iOS there is a platform API for this, however it is in the App Delegate or on a page, and this is pretty tricky to implement via something like Essentials. For shake detection, we can go this route of shared code with the accelerometer, but we need to do the following:

* Implement a Queue that is timestamp based. You can draw inspiration and a reference from [Seismic](https://github.com/square/seismic/blob/master/library/src/main/java/com/squareup/seismic/ShakeDetector.java).

pictos:
@jamesmontemagno , I'm sorry, but I don't know if I understend what you mean... Do you want to be implemented a record of times the cell vibrated?

Mrnikbobjeff:
@pictos As far as I understood all movement events are to be passed in a queue. All events after a certain timeframe are discarded (see purge method in linked file). Every event we check whether the saved events in a queue amount to something considered a shake. If so, invoke the event and continue. The linked example also tells  you on how to differentiate between a shake event and falling.

pictos:
@Mrnikbobjeff and @jamesmontemagno, my implementation does this.

### For timeframe 

The `const int shakenInterval = 500;` (0.5 Milliseconds) it's my lapse time. Any balance within this range of what is discarded.

### For Falling

The resulting axes of the accelerometer are given by the gravitational unit (g), 1g = 9,8 m/s². 

`var g = Math.Round(e.Reading.Acceleration.X.Square() + e.Reading.Acceleration.Y.Square() + e.Reading.Acceleration.Z.Square());`

if I'm not mistaken... In freefall this value it's close to 1g, and the event raise only if this resultant is greater than 2g (19,6 m/s²).

`const double accelerationThreshold = 2;`

I use this [code](https://stackoverflow.com/a/35095081) as a reference.

jamesmontemagno:
Thanks for the info! We will put this into a feature branch and work on a queue system for it as it is pretty close! Thanks!

pictos:
Always orders. And I'm sorry for the misunderstanding.

jamesmontemagno:
No worries, you are great. We end up doing this for almost all PRs to do a full review. It is easier then going back and forth when it is so close to being done :)

Issue #643, opened 2018-12-10
Integrate Android Support 28 NuGets


jamesmontemagno:
Integrate once they hit stable.

swapnil591:
Did nuget size matter

twsI:
```
<PackageReference Include="Xamarin.Android.Support.Design" Version="28.0.0" />
<PackageReference Include="Xamarin.Android.Support.v7.AppCompat" Version="28.0.0" />
<PackageReference Include="Xamarin.Android.Support.v4" Version="28.0.0" />
<PackageReference Include="Xamarin.Android.Support.v7.CardView" Version="28.0.0" />
<PackageReference Include="Xamarin.Android.Support.v7.MediaRouter" Version="28.0.0" />
```
They are now all at version 28.0.0 stable, therefore i guess, we can update. Should I create a PR?

PR #669, opened 2018-12-28
GH-643: Add Android P targeting and support 28 support

jamesmontemagno:
### Description of Change ###

Bump to use Android support v28 and target MonoAndroid90

### Bugs Fixed ###

- Related to issue #643

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


jamesmontemagno:
Will wait on this one for a bit I think.

Redth:
We might want to just plainly add Support 28.0.0 as a package and forget about the older versions for different monoandroid versions since the effect is the same with fat packages in 28.0.0.

jamesmontemagno:
@Redth let me know what you think here.

hashirmalik:
Xamarin.Essentials dose not support latest version!
image

jamesmontemagno:
@hashirmalik Simply install Xamarin.Android.Support.CustomTabs to your app and then update them all together

Xamarin.Essentials/Email/Email.android.cs
```diff
@@ -34,7 +34,12 @@ static Intent CreateIntent(EmailMessage message)
                 if (message?.BodyFormat == EmailBodyFormat.Html)
                 {
                     ISpanned html;
-                    if (Platform.HasApiLevel(BuildVersionCodes.N))
+#if __ANDROID_24__
+                    var hasN = Platform.HasApiLevel(BuildVersionCodes.N);
+#else
+                    var hasN = false;
+#endif
+                    if (hasN)
```
mattleibow: For the amount of times we do this, I would maybe create specific properties in Android's Platform - `HasApiLevelN` so that we can keep this `#if` stuff in one place. I can just see us either forgetting to use the `#if` or not properly doing something else.

Might be good to do this for the other major levels - such as M and P

jamesmontemagno: I like that idea :)

jamesmontemagno:
Let me know what you think @mattleibow  lots of edgecases when compiling so many versions up, but I think i got it :)

hashirmalik:
> @hashirmalik Simply install Xamarin.Android.Support.CustomTabs to your app and then update them all together

When I try to install this package which you above said, It cames with this error and counld not install !
I tried everything but nothing works. Great trouble please help me through this. Thanks!
**ScreenShot :**
error vs2017

Xamarin.Essentials/Connectivity/Connectivity.android.cs
```diff
@@ -18,7 +18,9 @@ static void StartListeners()
 
             conectivityReceiver = new ConnectivityBroadcastReceiver(OnConnectivityChanged);
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
SamuelDebruyn: This [constant](https://developer.android.com/reference/android/net/ConnectivityManager.html#CONNECTIVITY_ACTION) has been deprecated in favor of https://developer.android.com/reference/android/net/ConnectivityManager.html#registerNetworkCallback(android.net.NetworkRequest,%20android.app.PendingIntent) 

IMHO this should be replaced as this means that future Android versions will not support this anymore.

Issue #672, opened 2018-12-31
Add back SmsMessage with single recipient.


jamesmontemagno:
Not sure why we removed this... https://github.com/MicrosoftDocs/xamarin-docs/issues/1284

PR #674, opened 2019-01-02
GH-672: Update SMS recipient

jamesmontemagno:
### Description of Change ###
Adds a constructor for a single recipient.

### Bugs Fixed ###

- Related to issue #672

### API Changes ###

Added: 
 
- `SmsMessage.ctr(string, string);`

### Behavioral Changes ###

None

### PR Checklist ###


Xamarin.Essentials/Sms/Sms.shared.cs
```diff
@@ -31,6 +31,11 @@ public SmsMessage()
         {
         }
 
+        public SmsMessage(string body, string recipient)
```
Mrnikbobjeff: May I suggest outsourcing the body setting to a private constructor? this way we can elide setting the recipient list by allocating an array and copy constructing a list as well as allocating linq decorators. 

```
    public SmsMessage(string body, string recipient)
            : this(body)
        {
            Recipients = string.IsNullOrWhiteSpace(body) ? new List<string>(0) : new List<string>() {recipient};
        }

        private SmsMessage(string body)
        {
            Body = body;
        }

        public SmsMessage(string body, IEnumerable<string> recipients)
         : this(body)
        {
            Recipients = recipients?.Where(x => !string.IsNullOrWhiteSpace(x)).ToList();
        }`

Comment (mattleibow) [reply Xamarin.Essentials/Sms/Sms.shared.cs]: Good point. I would even say go further and forget about the chaining of constructors:

```csharp
public SmsMessage() {
}
public SmsMessage(string body, string recipient) {
    Body = body;
    if (!string.IsNullOrWhiteSpace(recipient))
        Recipients.Add(recipient);
}
public SmsMessage(string body, IEnumerable<string> recipients) {
    Body = body;
    if (recipients != null) {
        Recipients.AddRange(recipients.Where(x => !string.IsNullOrWhiteSpace(x)));
    }
}
// . . .
public List<string> Recipients { get; set; } = new List<string>();
```

We don't have to care about the initial size of the list as the default size is 4 (if corefx is anything to go by).

Mrnikbobjeff:
Well I believe if I were to add multiple recipients you would use the constructor accepting an IEnumerable. If you use the constructor with a single recipient I would rather explicitly set the Capacity of the List as this will probably be enough for the common use case of this constructor.

jamesmontemagno:
Updated! @mattleibow

docs/en/FrameworksIndex/xamarin-essentials-android.xml
```diff
@@ -10,22 +10,18 @@
       <Member Id="P:Xamarin.Essentials.Accelerometer.IsMonitoring" />
     </Type>
     <Type Name="Xamarin.Essentials.AccelerometerChangedEventArgs" Id="T:Xamarin.Essentials.AccelerometerChangedEventArgs">
-      <Member Id="M:Xamarin.Essentials.AccelerometerChangedEventArgs.#ctor(Xamarin.Essentials.AccelerometerData)" />
```
mattleibow: I think something went wrong with this file... There are too many changes.

jamesmontemagno:
@mattleibow should be good now!

Issue #696, opened 2019-02-01
Extra feature: Clipboard changed event


SaskiaL:
I would like to have a clipboard changed event. So that when another app puts something on the clipboard, that my app can pick it up and do something with it.

For instance, if a device has a hardware scanner that scans barcodes and puts the result on the clipboard, our app should get the read barcode.

I have tried to use my old code for the change event ((AddPrimaryClipChangedListener)) together with Clipboard, but then the changed event is no longer triggered.

Kind regards,
Saskia

// Essentials Team Info
Researched a bit:
* Android: Use the [Manager Listener](https://developer.android.com/reference/android/content/ClipboardManager)
* UWP: Easy event to register for [docs](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.datatransfer.clipboard.contentchanged)
*iOS: Need [NSNotificationCenter Observer](https://developer.apple.com/documentation/uikit/uipasteboard#//apple_ref/c/data/UIPasteboardChangedNotification)

API:
```csharp
public static event EventHandler<ClipboardChangedEventArgs> ClipboardChanged;

 public class ClipboardChangedEventArgs: EventArgs
    {

    }
```

We don't need to return any information in the eventargs, because our API should return this. It is a generic clipboard changed, we don't know if it is text or other items....

Mrnikbobjeff:
On IOS this feature might require setting up some code in the app initialization function. Should not be a problem on Android and iOS

jamesmontemagno:
Added documentation, should be pretty easy to do :) Up for grabs :)

Mrnikbobjeff:
Am working on it, a pr should be up before the end of the week 👍

jamesmontemagno:
@Mrnikbobjeff can you base it off of the 1.1.0 branch :)

manishkungwani:
@jamesmontemagno @Mrnikbobjeff looks like the code and docs were done here - https://github.com/xamarin/Essentials/pull/720

What's pending, if anything?
If so, can I pick it up and finish it off.

SaskiaL:
I tried to install version 1.4.0-pre2 to try the new event, but got this error message:
Failed to add reference. The package 'Xamarin.Essentials' tried to add a framework reference to 'System.Drawing.Common.dll' which was not found in the GAC.

Is this a problem with the package? Or do I need to install extra packages?

Thanks, Saskia

manishkungwani:
Hi,
I'm trying to use the package as well, with no luck.
[...]

PR #720, opened 2019-02-26
GH-969 Clipboard Changed

Mrnikbobjeff:
### Description of Change ###
Added Clipboard Change Event
- Related to issue #696

### API Changes ###
Added: 
 On Clipboard
- ` event ClipboardChangedEvent<EventArgs>`

### PR Checklist ###


Mrnikbobjeff:
Sorry for the delay, had uni stuff coming up. This is basically finished and ready to be merged. I still need to update the docs tomorrow. Tested on Android&UWP, ios is so simple that I dont think i made a mistake there.

Samples/Samples/ViewModel/ClipboardViewModel.cs
```diff
@@ -24,6 +25,21 @@ public string FieldValue
...
+        }
+
+        public override void OnDisappearing()
+        {
+            Clipboard.ClipboardContentChanged -= OnClipboardContentChanged;
+        }
+
+        async void OnClipboardContentChanged(object sender, EventArgs args)
+        {
+            await DisplayAlertAsync("Clipoard content changed!");
```
mattleibow: Maybe not have an alert... A status label might be better - you could have a "Clipboard changed: {the text}" or something.

I wonder if there are native features to clear the clipboard for some reason? That might also be a nice feature. But, that can be a separate PR.

Mrnikbobjeff: Not all implementations include the changed content, and as we only support text clipboard content I would advise against including the changed content in the event. You can always grab clipboard content in the event handler, but this way we have the bare minimum api across all platforms without fetching the clipboard :)

mattleibow: On, no. I was talking about instead of using `DisplayAlertAsync`, rather update some label in the sample UI. I want to move away from popups as they break UI tests and are not as nice as a status label.

Mrnikbobjeff: Done

Mrnikbobjeff:
Docs are live, everything should be done. Android device tests should work if you restart them, they did not fault because of me.

Issue #681, opened 2019-01-10
[SecureStorage] Java.Lang.UnrecoverableKeyException


SamuelDebruyn:
### Description

I get an exception when calling GetAsync and SetAsync in SecureStorage

### Steps to Reproduce

Tried to call GetAsync on a key that should exist. Later tried to call SetAsync for that same key.
I don't have a repro yet, but I provided a lot of stack trace & debug info...

Exception happens both in GetAsync and SetAsync

### Expected Behavior

Just return the value

### Actual Behavior

Exception thrown.

Exception message: Failed to obtain information about key
Exception type: Java.Security.UnrecoverableKeyException
Exception stacktrace:
```
  at Java.Interop.JniEnvironment+InstanceMethods.CallNonvirtualObjectMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x00089] in <42dc777b518744fdae9988e94489a4a0>:0 
  at Java.Interop.JniPeerMembers+JniInstanceMethods.InvokeNonvirtualObjectMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters) [0x0001f] in <42dc777b518744fdae9988e94489a4a0>:0 
  at Java.Security.KeyStore.GetKey (System.String alias, System.Char[] password) [0x00043] in <1219ce5aae934ab095dc0e05b2110050>:0 
  at Xamarin.Essentials.AndroidKeyStore.GetSymmetricKey () [0x00011] in C:\agent\_work\69\s\Xamarin.Essentials\SecureStorage\SecureStorage.android.cs:164 
  at Xamarin.Essentials.AndroidKeyStore.GetKey () [0x0002d] in C:\agent\_work\69\s\Xamarin.Essentials\SecureStorage\SecureStorage.android.cs:108 
  at Xamarin.Essentials.AndroidKeyStore.Encrypt (System.String data) [0x00000] in C:\agent\_work\69\s\Xamarin.Essentials\SecureStorage\SecureStorage.android.cs:249 
  at Xamarin.Essentials.SecureStorage.PlatformSetAsync (System.String key, System.String data) [0x00014] in C:\agent\_work\69\s\Xamarin.Essentials\SecureStorage\SecureStorage.android.cs:48 
[...]

PR #736, opened 2019-03-11
Handle Invalid Key Exception

jamesmontemagno:
### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #681

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
+++ b/Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
@@ -244,7 +244,7 @@ namespace Xamarin.Essentials
             return unwrapped;
         }
 
-        internal byte[] Encrypt(string data)
+        internal byte[] Encrypt(string data, bool retry = true)
         {
             var key = GetKey();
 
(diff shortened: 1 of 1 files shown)


Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
```diff
@@ -259,20 +259,34 @@ internal byte[] Encrypt(string data)
...
-                // If we encounter this error, it's likely an old bouncycastle provider version
-                // is being used which does not recognize GCMParameterSpec, but should work
-                // with IvParameterSpec, however we only do this as a last effort since other
-                // implementations will error if you use IvParameterSpec when GCMParameterSpec
-                // is recognized and expected.
-                cipher = Cipher.GetInstance(cipherTransformationSymmetric);
-                cipher.Init(CipherMode.EncryptMode, key, new IvParameterSpec(iv));
+                if (retry)
+                {
+                    keyStore.DeleteEntry(alias);
```
xleon, line 282: Does this mean that the current data will be lost whenever `InvalidKeyException` happens?

Issue #238, opened 2018-05-10
Add Local Notifications


craylward:
Use case: Showing local notifications in Xamarin Forms projects.

https://github.com/edsnider/LocalNotificationsPlugin

> VS bug [#735671](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735671)

jamesmontemagno:
For Xamarin.Essentials we are going to keep away from any UI specific features at this time. Just a lot of dependencies and issues around how to handle navigation cross-platform. Thanks for the feature request though and we will see if it evolves over time.

mattleibow:
@jamesmontemagno @Redth I was looking at this issue and the plugin and this _may_ be something we want to do. These local notifications to the system, outside the app may be useful and they are very different across platforms.

With regards to the "UI specific" features. Sure, this is UI. But then we have the browser, SMS, dialer, email... The data transfer also shows a popup. I think we need to re evaluate the idea of UI with Essentials as many things have UI. It may be better to see that we don't want a "UI component" (such as a widget or button) and if it has a UI, it must be self-contained.

Does this make sense?

jamesmontemagno:
Will leave it as proposal with more discussion to be had.

KSemenenko:
I think that this would be a great addition.
Now I'm looking for a suitable library for this, for example
https://github.com/aritchie/notifications or https://github.com/edsnider/LocalNotificationsPlugin
and I think it would be good if such a basic functional would be part of Xamarin.Essentials

aritchie:
To do this, you would need the addition of "Xamarin.Android.Support.v4".  Taking in additional support libs at this point isn't a big deal since any one of them basically brings the entire suite one way or another.  

However, in order to make the API really decent, you should be able to pull up a list of scheduled notifications.  In Android, there is no API to do this.  I deal with this by creating a reference to sqlite and storing the metadata internally.  This also allows the user to clear the list when a user logs out of an application.

A fallback plan is to simply return an ID of the notification that was registered and let the user deal with how they want to store notification data.

KSemenenko:
I think that this question is only about showing notification bubble. It can be more about badges on the application icon. @aritchie suggested a good idea with store notification data.

jamesmontemagno:
I have attempted to handle notifications and scheduling and it is really a mess especially with icons as that is a different API. This probably needs to be broken down into multiple issues and then mocked up as to what each platform can actually do and what it would like like.

KSemenenko:
[...]

PR #739, opened 2019-03-16
GH238: [WIP] Local Notifications

Redth:
### Description of Change ###
Started adding the implementation of Local Notifications.

Lots of work left to do, but wanted to get the discussions rolling around this.

- Related to issue #238 

### API Changes ###

```csharp
class LocalNotifications {
  Task CancelAll();
  Task Cancel(int id);
  Task Send(LocalNotification notification);
  Task<IEnumerable<LocalNotification>> GetNotificationsAsync();
}

class LocalNotification {
  int Id { get; set; }
  string Title { get; set; }
  string Message { get; set; }
  IDictionary<string, string> Metadata { get; set; }
}
```

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/LocalNotifications/LocalNotifications.android.cs
+++ b/Xamarin.Essentials/LocalNotifications/LocalNotifications.android.cs
@@ -1,0 +1,92 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Android.App;
+using Android.Support.V4.App;
+
+namespace Xamarin.Essentials
+{
+    public static partial class LocalNotifications
+    {
+        const string extraMetadataPrefix = "xemetadata_";
+
+        static Task PlatformCancelAll()
+        {
+            var mgr = NotificationManager.FromContext(Platform.AppContext);
+            mgr.CancelAll();
+
+            return Task.CompletedTask;
+        }
+
+        static Task PlatformCancel(int id)
+        {
+            var mgr = NotificationManager.FromContext(Platform.AppContext);
+
+            mgr.Cancel(id);
+
+            return Task.CompletedTask;
+        }
+
+        static Task PlatformSend(LocalNotification notification)
+        {
+            var context = Platform.AppContext;
+
+            var mgr = NotificationManager.FromContext(context);
+
+            var metadataBundle = new global::Android.OS.Bundle();
+
+            // Prefix extras with a known value since they can contain more than our bundle
+            foreach (var kvp in notification.Metadata)
+                metadataBundle.PutString($"{extraMetadataPrefix}{kvp.Key}", kvp.Value);
+
+            NotificationCompat.Builder builder = default;
+
+#if __ANDROID_26__
+            // We should be good to use this even if _running_ on < 25
+            // since this is a support lib api and it will gracefull degrade
+            builder = new NotificationCompat.Builder(context, "Default");
+#else
+            builder = new NotificationCompat.Builder(context);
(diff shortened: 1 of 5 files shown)


Redth:
Couple of bigger outstanding items/questions for discussion:

1. How do we handle notifications being pressed (callback)?  
2. Can we enable some of the platform specific customization options in a friendly way that makes sense? Maybe we can develop a pattern to use in other places.

Redth:
For callbacks each platform seems to have the concept of basically calling back into the main application to handle things.
- iOS: `UNUserNotificationCenterDelegate` is used to proxy the event to - this needs to be registered in the appdelegate when the app is launched, and then we could raise an event from here
- Android: Intents! Yay! we could easily create an intent that triggers a particular broadcast receiver we also control, OR we could create an intent which calls back the launcher activity for the given package context, and then process the intent and raise an event (the user would have to wire up the event handler early enough on).
- UWP - basically will call back into the application at which point we can handle the call and raise an event.

These all pretty much hinge on the user wiring up an event handler very early on in their application lifecycle, in addition to wiring up some other methods in the app to pass into essentials.

This is going to get a _bit_ ugly but I think it's doable.

bbl-Laobu:
I am not sure if this is the place for me to add my little wish list but I would love to see the following added to LocalNotification Class:

- When: DateTime of when the notification should be activated
- Vibrate: bool indicating if vibration should be activated or not
- Sound: uri to a sound file to play when notification is shown
- Duration: Timespan of how long the sound should play for

Cheers,
Bernard

jrahma:
You should also consider the following:

- Custom Image
- Custom Sound
- Link to AppCenter API so when the device is connected then AppCenter portal should be updated

jrahma:
Also message should receive when the app is on background or foreground

jrahma:
> I am not sure if this is the place for me to add my little wish list but I would love to see the following added to LocalNotification Class:
> 
> * When: DateTime of when the notification should be activated
> * Vibrate: bool indicating if vibration should be activated or not
> * Sound: uri to a sound file to play when notification is shown
> * Duration: Timespan of how long the sound should play for
> 
> Cheers,
> Bernard

This is a local nitrification so **Sound** should be from Resources which is more important when the user is offline and you still want him to receive the notification

thudugala:
if you guys are interested https://github.com/thudugala/Plugin.LocalNotification

jrahma:
You should also consider RemoveNotification to allow developer to remove the notification by notification ID..

Here is updated API:

```
class LocalNotification {
  int Id { get; set; }
  string Title { get; set; }
  string Message { get; set; }
  string Sound { get; set; }
  DateTime Schedule { get; set; }
  DateTime Disappear { get; set; }
  bool Vibrate { get; set; }
  IDictionary<string, string> Metadata { get; set; }
}

```

I guess this will be great kick off to start

jrahma:
One more thing..

If the user missed the notification because his device was switched off?

There should be an option to keep the message until the device is turned on

I repeat, it should be an option because sometimes we don't want to send the message later, for example if it's an offer between 1pm and 2pm and user switched on his device at 8pm then he should not get it

but if it's a Christmas greeting for example then any time user turned on his devices he should get it

jfversluis:
This is amazing! I think this will make a lot of people happy, but let's move this over to [.NET MAUI Essentials](https://github.com/dotnet/maui/discussions/4216) for now.

janseris:
Hi, what is the state of this? MAUI framework is incomplete without notifications. Mobile apps need notifications. It's a basic functionality.

Issue #178, opened 2018-04-11
Record Audio API


jamesmontemagno:
## Description of the Feature

Ability to record audio into a stream

## API Declaration

- Task AudioRecorder.StartAsync();
- AudioFile AudioRecorder.StopAsync();

## Additional Classes or Structs needed
- Audio File is an option here with the stream, information?

## Documentation links for supported platforms
* [Android Document](https://developer.android.com/guide/topics/media/camera.html)
* [iOS Document](https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture)
* [UWP Document](https://docs.microsoft.com/en-us/windows/uwp/audio-video-camera/)

> VS bug [#735676](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735676)

AmrAlSayed0:
Shouldn't Android documentation point to [MediaRecorder](https://developer.android.com/reference/android/media/MediaRecorder) instead of [Camera](https://developer.android.com/guide/topics/media/camera) ?

Redth:
I'm not sure it's reasonable to do anything but record to some sort of temporary/intermediate file here.  

Is there a particular audio format and bitrate that is neutral / optimal across all the platforms?  
AudioFile is probably going to look something like:

```csharp
public Stream ReadAsStreamAsync();
int Bitrate { get; }
AudioFormat Format { get; }
```

Redth:
And it's not a problem to record with a consistent audio bitrate cross the platforms?

Does PInvoke on UWP cause any issues or require special considerations for the app store?

Mrnikbobjeff:
No, consistent Bitrate should not be a problem. While we may not be able to guarantee a specific bitrate for every device, we can definitely allow the user to query supported bitrates and select their preset. For 99%+ of devices this will be a value supported on all platforms. As for your second question, to be more specific we don't need PInvoke but rather COM Interop, I threw together the terms though. I am not 100% sure that it absolutely requires it, but the Implementation we used to work with used it to manipulate native unmanaged buffers. With the MediaCapture class I can not learn enough from the docs to be able to evaluate whether their behaviour matches that of other platforms.

jamesmontemagno:
Probably something like Adrian's https://github.com/adrianstevens/Xamarin-Plugins/tree/master/SimpleAudioRecorder

jfversluis:
[...]

Issue #125, opened 2018-03-28
Screenshot API


mattleibow:
Would we want a Screenshot of the current window/screen?

> VS bug [#735683](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735683)

pictos:
I would like to implement this API. Can you assign this to me?

Redth:
This one needs some more discussion and specifications before it can be implemented.

1. What is the native API like in Android, iOS, and UWP ?
2. Do we need to make the method async?
3. What should this return?  `byte[]` or the filename?
4. What format of image can this be which all platforms support?

pictos:
In the past I did a PR for this API. Using [this](https://github.com/xamarin/Essentials/pull/508) as a reference....

> **1. What is the native API like in Android, iOS, and UWP ?**

When I implemented I did not make use of the native APIs, to ensure that the implementation was similar on all platforms.


> **2. Do we need to make the method async?**

Yes, because in all 3 platforms we use async methods. To get the `byte[]` or to save the sreenshot image.

> **3. What should this return? byte[] or the filename?**
The both. Like this:

```csharp
 public static Task<string> CaptureAsync(ScreenOutputType type = ScreenOutputType.PNG, string fileName = null) =>
            PlataformCaptureAsync(type, fileName);
         public static Task<byte[]> GetImageBytesAsync(ScreenOutputType type = ScreenOutputType.PNG) =>
            PlataformGetImageBytesAsync(type);
```
The developer can use the `byte[] `when he wants to do some other image processing, such as sending to a service. And we go back the path, when developer just wants to save the image on the device.

> 4. **What format of image can this be which all platforms support?**
On all 3 platforms, we can convert the image to .png or .jpg .

Redth:
I think if we can reasonably get a PNG back on each platform, we should just use that and not have an option for JPG (unless there's a compelling reason for needing both).

I also don't think we should have two methods returning different results.  If anything we maybe need to return a `ScreenshotFile` type or something similar and provide the filename and helper `ReadStreamAsync` and `ReadBytesAsync` type methods.  It might be worth considering if we can use the same return type we will eventually be using for the photo picker.

pictos:
First, I have 2 questions:

1. How Photo Picker works?
2. Will the screen capture API be in which folder?

How about we return a struct containing the information of the media types, can be used in Photo Picker and other APIs that handle media files. Like that:

```csharp
[...]

PR #744, opened 2019-03-19
GH-125 Take Screenshot

Mrnikbobjeff:
### Description of Change ###
- Related to issue #125. New PR because its quicker to just cherry pick the commits

### API Changes ###

Added: 
 
```public static partial class Screenshot
    {
        public static bool CanCapture => PlatformCanCapture;

        public static Task<MediaFile> CaptureAsync()
            => PlatformCaptureAsync();
    }
```
And the same media file as in #178 

Docs update can come once we decide on the exact mediafile structure..

### PR Checklist ###


Mrnikbobjeff:
@syed-pelican @mattleibow ping, new pr same code

syed-pelican:
and the merge failed.....

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.IO;
+
+namespace Xamarin.Essentials
+{
+    public class MediaFile : IDisposable
```
mattleibow: We just merged in the file sharing bits, and as a result we got a new `FileBase` type. Could you see if you can re-use that type? We want to use that as a base type for all files. So, for this PR, you would maybe have a `ScreenshotFile : FileBase` and you can add any new methods you need to the new type. We hope that this will work, and you are the first to do something more than what we have. Let us know what doesn't work so we can make that base type be useful.

Have a look at how we did `EmailAttachment` and you can see what we are trying to do.

Xamarin.Essentials/Screenshot/Screenshot.android.cs
```diff
@@ -0,0 +1,42 @@
...
+using Path = System.IO.Path;
+using Uri = Android.Net.Uri;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        static bool PlatformCanCapture => !Platform.WindowManager.DefaultDisplay.Flags.HasFlag(DisplayFlags.Secure);
+
+        static string GetTempFileName() => Path.Combine(Environment.ExternalStorageDirectory.Path, Path.GetTempFileName());
```
mattleibow: I am thinking that this method might be better served on the `FileSystem` type.

Xamarin.Essentials/Screenshot/Screenshot.ios.cs
```diff
@@ -0,0 +1,31 @@
...
+using Foundation;
+using UIKit;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformCanCapture => UIScreen.MainScreen != null;
+
+        static string GetTempFileName()
```
mattleibow: Can we make use of the members on the FileSystem type? If not, it might be better to add something there for other APIs to use.

Xamarin.Essentials/Screenshot/Screenshot.android.cs
Mrnikbobjeff: Definitely, I used similar code in 2 other active branches. Does adding the method to the platform specific Filesystem class with internal visibility sound okay?

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
Mrnikbobjeff: Just took a look,we can definitely reuse the FileBase. I would love to see virtual methods for reading the file though, as on some platforms we can ellide the temporary file creation for the screenshot, which would be preferrable imho. then, if we were to use the same FileBase in another api where it would be passed in (e.g.) EmailAttachment, we could materialize the file in the file system as needed and not earlier. With this we could also add in-memory files to other api's easily.

Xamarin.Essentials/Screenshot/Screenshot.ios.cs
Mrnikbobjeff: No, the current FileSystem type does not provide the necessary functionality. Adding it does sound like a good idea. See the other post

Xamarin.Essentials/Screenshot/Screenshot.android.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        static async Task<MediaFile> PlatformCaptureAsync()
+        {
+            var path = GetTempFileName();
+            var view = Platform.GetCurrentActivity(false).Window.DecorView.RootView;
+            view.DrawingCacheEnabled = true;
```
newky2k: The build is failing because  Screenshot\Screenshot.android.cs(30,13): error CS0618: 'View.DrawingCacheEnabled' is obsolete: 'deprecated' (locally and in CI)

Xamarin.Essentials/Screenshot/Screenshot.android.cs
```diff
@@ -0,0 +1,41 @@
...
+            var bitmap = Bitmap.CreateBitmap(view.GetDrawingCache(true));
```
newky2k: The build is failing because  Screenshot\Screenshot.android.cs(31,46): error CS0618: 'View.GetDrawingCache(bool)' is obsolete: 'deprecated' (locally and in CI)

manishkungwani:
@newky2k @Mrnikbobjeff may I try to rebase and re-raise this PR?

Mrnikbobjeff:
@newky2k Would you mind taking a look at it again? I added docs and removed the deprecated method. I really would like to not have to merge master changes on this many branches :)

Samples/Samples/ViewModel/HomeViewModel.cs
```diff
@@ -156,6 +156,12 @@ public HomeViewModel()
                     typeof(PreferencesPage),
                     "Quickly and easily add persistent preferences.",
                     new[] { "settings", "preferences", "prefs", "storage" }),
+                new SampleItem(
+                    "⚙️",
```
mattleibow: Maybe use the camera icon as this looks more settings-y.

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace Xamarin.Essentials
+{
+    public class ScreenshotFile : ReadOnlyFile, IDisposable
```
mattleibow: On this, is there a reason we wouldn't just store the image in memory as a stream? It should never be that big, then we don't need to worry about anything. The user could then save it to a file or process it right out the gate.

@jamesmontemagno Not sure if we need some sort of guidelines for when to use an in-memory stream or cache to disk first. This will come in with a camera picker or something. The image might be very big. But, I think there should always be enough memory to hold a bitmap of the screen dimensions since the OS just rendered it.

Xamarin.Essentials/Screenshot/Screenshot.android.cs
mattleibow: I have a comment on `ScreenshotFile` but looking at this, if we do go the file route, we should probably use the OS cache location (`FileSystem.CacheDirectory` or something) rather than the external storage - the user may not have permissions or something. We don't have to worry about permissions for sharing as the `Share` API does all that.

Xamarin.Essentials/Screenshot/Screenshot.ios.tvos.cs
```diff
@@ -0,0 +1,30 @@
...
+using Foundation;
+using UIKit;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformCanCapture => UIScreen.MainScreen != null;
+
+        static string GetTempFileName()
```
mattleibow: See my comment on Android - we should probably use the Cache directory so the OS can clean up after us.

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
mattleibow: Hmm... I think I like the file idea actually. This will make is consistent with all the other types of pickers. We probably need to rename the source code file though (from MediaFIle to ScreenshotFile)

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public bool Disposed { get; private set; }
+
+        internal ScreenshotFile(string filePath)
+            : base(filePath, "image/png")
+        {
+        }
+
+#if !NETSTANDARD1_0
+        public Stream AsStream() => System.IO.File.Open(FullPath, FileMode.Open, FileAccess.Read);
```
mattleibow: I would suggest making this `Task<Stream> PlatformOpenReadStreamAsync()` so that it matches the FilePicker: https://github.com/xamarin/Essentials/pull/975/files#diff-afb1518dfb7ed9f416a31b1bd082b4d6R113

I am hoping to have all the pickers inherit from `PickerResultBase` (or a better named class) so that the `ReadOnlyFile` is more for the sending out and we have a "picker" type file that has the extras for opening it and maybe a method (like you have here on the dispose) to clean up.

mattleibow: Also, with the #if, make sure you add a dummy method for netstandard1.0 so that the API exists for the compiler.

```csharp
#if NETSTD1
  Task<Stream> PlatformOpenReadStreamAsync() => Task.FromResult<Stream>(null);
#else
  Task<Stream> PlatformOpenReadStreamAsync() => ...;
#endif
```

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+        protected virtual void Dispose(bool isDisposing)
+        {
+            if (Disposed)
+                return;
+#if !NETSTANDARD1_0
+            if (isDisposing)
+            {
+                try
+                {
+                    System.IO.File.Delete(FullPath);
```
mattleibow: I am not sure how I feel about this just yet... I like the idea, and it is something that I didn't think of when I was working on the file picker. We do want to clean up after ourselves. But... it might be safer for the user to do it as they may be storing the value somewhere to read later. Thoughts?

Xamarin.Essentials/Screenshot/Screenshot.watchos.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using Foundation;
+using UIKit;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformCanCapture => false; // On WatchOs 3 this appears to be possible if activated through settings.
```
mattleibow: Interesting... Can we detect this value or read it? Then we can have this property return true sometimes.

Samples/Samples/ViewModel/HomeViewModel.cs
Mrnikbobjeff: Where do you get the real icons from? I always just copied an existing icon as I don't know how to insert new icons. Is it as simple as opening charmap and copying any font icon in there?

mattleibow: I just use sites like these: https://getemoji.com/

📷

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
Mrnikbobjeff: I really like this pattern, I usually use it to force ownership semantics on the consumer. The file might be deleted eventually (I have no idea how finalizers work on ios/android/tizen, but I assume they have the same "guarantees" for running as normal finalizers, which are basically none.) As such the user is forced to do something with the file and can not just leave an artifact lying aroung, even if he does not dispose of it. I would just annotate this behaviour in the comments.

mattleibow: Good enough.

Xamarin.Essentials/Screenshot/Screenshot.netstandard.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
```
Mrnikbobjeff: I need to add the throw helper here too

Xamarin.Essentials/Screenshot/Screenshot.ios.tvos.cs
```diff
@@ -0,0 +1,30 @@
...
+            var docFolder = Environment.GetFolderPath(Environment.SpecialFolder.Personal);
+            var libFolder = Path.Combine(docFolder, "..", "Library");
+            var tempFileName = Path.ChangeExtension(Path.Combine(libFolder, Path.GetTempFileName()), ".png");
+            return tempFileName;
+        }
+
+        static Task<ScreenshotFile> PlatformCaptureAsync()
+        {
+            var img = UIScreen.MainScreen.Capture();
+            var bytes = img.AsPNG().ToArray();
```
Mrnikbobjeff: See if we can avoid array allocation and somehow use a stream

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
Mrnikbobjeff: Yeah, the branch existed before Media file existed, there are still some remnants. Good catch!

Xamarin.Essentials/Screenshot/MediaFile.shared.cs
Mrnikbobjeff: Sure can do, how would you feel about making the actual filestream returned async as well? We currently create the stream via File.Open which does not allow us to specify the async'ness of the resulting stream - if we call the filestream constructor directly we will be able to specify it to be an async stream.

Xamarin.Essentials/Screenshot/Screenshot.watchos.cs
Mrnikbobjeff: Perhaps, but AFAIK it is not possible currently to take screenshots on watchos programatically. Would make no sense returning true then IMHO

Mrnikbobjeff:
@mattleibow I moved all files to the cache location, changed the icon and renamed the AsStream method including the signature change. Should be ready to go.

Xamarin.Essentials/Screenshot/Screenshot.android.cs
```diff
@@ -0,0 +1,46 @@
...
+using Android.Views;
+using Path = System.IO.Path;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        static bool PlatformCanCapture => !Platform.WindowManager.DefaultDisplay.Flags.HasFlag(DisplayFlags.Secure);
+
+        static string GetTempFileName() => Path.Combine(FileSystem.CacheDirectory, Path.GetTempFileName());
```
mattleibow: @jamesmontemagno @Redth I am still not really happy with this... This folder is publicly available to the user, so _should_ we be putting things in here? @Mrnikbobjeff previously had it in the external storage folder, and I sort of felt it is better in the cache folder. I think it is better now, but maybe we need to decide on what we want to do with files we create. Should we have a special folder in the cache where we work? Should we truly use some OS temp system?

mattleibow: @jamesmontemagno @Redth I am still not really happy with this... This folder is publicly available to the user, so _should_ we be putting things in here? @Mrnikbobjeff previously had it in the external storage folder, and I sort of felt it is better in the cache folder. I think it is better now, but maybe we need to decide on what we want to do with files we create. Should we have a special folder in the cache where we work? Should we truly use some OS temp system?

mattleibow: @Mrnikbobjeff I am thinking this is probably better to just be in the `PlatformCaptureAsync` method as it is a single line.

Xamarin.Essentials/Screenshot/Screenshot.ios.tvos.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformCanCapture => UIScreen.MainScreen != null;
+
+        static string GetTempFileName()
+        {
+            var tempFileName = Path.ChangeExtension(Path.Combine(FileSystem.CacheDirectory, Path.GetTempFileName()), ".png");
```
mattleibow: To be consistent, we should probably use `Path.Combine(FileSystem.CacheDirectory, Path.GetTempFileName());` like Android. Also, as it is a single line, just put it in the `PlatformCaptureAsync` method.

Xamarin.Essentials/Screenshot/Screenshot.uwp.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformCanCapture => true;
+
+        static async Task<ScreenshotFile> PlatformCaptureAsync()
+        {
+            var element = Window.Current.Content as FrameworkElement;
+            var fileOnDisk = await ApplicationData.Current.LocalCacheFolder.CreateFileAsync(Path.ChangeExtension(Path.GetRandomFileName(), ".png"));
```
mattleibow: To be consistent, we should probably use `Path.Combine(FileSystem.CacheDirectory, Path.GetTempFileName());` like Android.

Xamarin.Essentials/Screenshot/Screenshot.watchos.cs
mattleibow: Good enough.

mattleibow:
I think this is really close, that location where we save temp files is still bugging me.

Xamarin.Essentials/Screenshot/Screenshot.android.cs
Mrnikbobjeff: An interesting question would be to see how to save files in the application cache folder, as the user can clear than themself.

Xamarin.Essentials/Screenshot/Screenshot.uwp.cs
Mrnikbobjeff: I actually prefer the uwp implementation over all others, as this allows the user to remove the cached data by cleaning app data.

Mrnikbobjeff: I like this more as the user can clear application cache directly

Xamarin.Essentials/Screenshot/Screenshot.android.cs
Mrnikbobjeff: I would like to save this in the application cache folder, as the user can clear that manually

Mrnikbobjeff:
Should we use Context.CacheDir on Android? https://docs.microsoft.com/en-us/dotnet/api/android.content.context.cachedir?view=xamarin-android-sdk-9

Mrnikbobjeff:
Are there any plans to finally include this branch into the upcoming release? There has been virtually no developement on this branch except deciding where to store the file in the last 8 months. We are already rolling our own EssentialsEx package including this and other branches which remained virtually finished for quite a while. @jamesmontemagno @mattleibow

Redth:
Merged in #1356

Issue #425, opened 2018-07-28
Share: Share Documents/Files


jrahma:
Hi,

Please add an option to Share Documents with a Title and URL..

For example, share PDF in Whatsapp or Google Drive or One Drive or DropBox..

Thanks,
Jassim

> VS bug [#735659](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735659)

mattleibow:
This is probably related to the #129 and #416 and some other issues. The basic problem is that sharing files may require access to internal/local app files. I think #416 is to address the initial limitation.

Redth:
1.1.0 has file sharing!

xamarin-release-manager:
[VS sync] The field 'Milestone' contains the value '1.1.0' that is not in the list of supported values

mfeingol:
Nice to see this land! I gave it a spin with code along these lines:

```c#
                string filePath = path to a PDF;
                string contentType = "application/pdf";

                ShareFile share = new ShareFile(filePath, contentType);
                ShareFileRequest request = new ShareFileRequest(Path.GetFileName(filePath), share);
                await Share.RequestAsync(request);
```

It worked, in that a system dialog opens with some apps as options. However, the apps were Gmail, Save to Drive, and Messages, aka generic options. If I use the more traditional approach (e.g. https://www.syncfusion.com/kb/8707/how-to-share-the-pdf-document-in-xamarin-forms-platform), I get PDF-specific options, such as "Drive PDF Viewer".

Is there some setup required before the feature will work as expected?

Thanks.

jamesmontemagno:
You will need to enable the experimental features:

// Enable currently experimental features
            ExperimentalFeatures.Enable(
                ExperimentalFeatures.EmailAttachments,
                ExperimentalFeatures.ShareFileRequest);

Under the hood it calls:

```csharp
            var contentUri = Platform.GetShareableFileUri(request.File.FullPath);

            var intent = new Intent(Intent.ActionSend);
            intent.SetType(request.File.ContentType);
            intent.SetFlags(ActivityFlags.GrantReadUriPermission);
            intent.PutExtra(Intent.ExtraStream, contentUri);

            if (!string.IsNullOrEmpty(request.Title))
            {
                intent.PutExtra(Intent.ExtraTitle, request.Title);
            }

            var chooserIntent = Intent.CreateChooser(intent, request.Title ?? string.Empty);
            chooserIntent.SetFlags(ActivityFlags.ClearTop);
[...]

PR #773, opened 2019-04-27
Implement Open File in Launcher.  Fixes #601 & #425

### LINKED ISSUE #601: Opening files in the default application
The idea is to provide a method for opening files in the application by default. 
For example, in my application I need to download files of different types, for example pdf, docx, pictures, and I’ll like to be able to open them with applications already installed, for example Acrobat for pdf, but for the system to choose.

jamesmontemagno:
### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #601 & #425

### API Changes ###

Added: 
 
- OpenFileRequest
- ReadOnlyFile: FileBase
- Launcher.OpenAsync(OpenFileRequest)

Changed:

Removed "AttachmentName" from the files as it shouldn't be set.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Launcher/Launcher.shared.cs
+++ b/Xamarin.Essentials/Launcher/Launcher.shared.cs
@@ -53,7 +53,7 @@ namespace Xamarin.Essentials
             ExperimentalFeatures.VerifyEnabled(ExperimentalFeatures.OpenFileRequest);
         }
 
-        public OpenFileRequest(string title, LauncherFile file)
+        public OpenFileRequest(string title, ReadOnlyFile file)
         {
             ExperimentalFeatures.VerifyEnabled(ExperimentalFeatures.OpenFileRequest);
             Title = title;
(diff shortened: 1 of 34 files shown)


jamesmontemagno:
Still need to test UWP and iOS, but should be close

Xamarin.Essentials/Launcher/Launcher.shared.cs
```diff
@@ -36,5 +37,81 @@ public static Task OpenAsync(Uri uri)
...
+
+        public LauncherFile(FileBase file)
+            : base(file)
+        {
+            ExperimentalFeatures.VerifyEnabled(ExperimentalFeatures.OpenFileRequest);
+        }
+
+        string attachmentName;
+
+        public string AttachmentName
```
mattleibow: I see this is the third time we have this implemented (Share and Email). We should probably move the implementation to the base (and make it protected if we don't want it visible). We can call it something like `FileNameOverride` and the implementations can just use this value under the hood.

But, I don't see this being used here at all, so I am not sure if this even needs to be here for this case.

jamesmontemagno: Fixed!

RhomGit:
Hi @jamesmontemagno , firstly thanks for this addition to Xamarin Essentials!

I am testing the latest nuget (v1.2.0-pre) and I don't think the OpenFileRequest is working on iOS. It works perfectly on Android, and the Share seems to work perfectly in both Android and iOS.


`                string fn = this.FileName_FullIncludingPath;`
`                var uri = new Uri(fn);`
`                var supportsUri = await Launcher.CanOpenAsync(uri.Scheme + @"://");`
`                if (supportsUri)`
`                    await Launcher.OpenAsync(new OpenFileRequest(this.fileCaption, new ReadOnlyFile(fn)));`

The above code executes fine but the OpenAsync call just doesn't seem to do anything on iOS. I have tried .png and .pdf files.

Thanks again,
Denis

PR #850, opened 2019-07-18
[Tizen] Fix reference for avoid duplication

sung-su:
### Description of Change ###

I fixed the reference to not include `System.Numerics.Vectors.dll` in `Sample.Tizen` project 

### PR Checklist ###

### DIFF
--- a/Samples/Samples.Tizen/Samples.Tizen.csproj
+++ b/Samples/Samples.Tizen/Samples.Tizen.csproj
@@ -25,6 +25,9 @@
 
   <ItemGroup>
     <PackageReference Include="Tizen.Wearable.CircularUI" Version="1.3.0-pre1-00043" />
+    <PackageReference Include="System.Numerics.Vectors" Version="4.5.0">
+      <ExcludeAssets>Runtime</ExcludeAssets>
+    </PackageReference>
   </ItemGroup>
 
   <Import Project="$(MSBuildThisFileDirectory)..\..\CodeStyles.targets" />
--- a/Xamarin.Essentials/Xamarin.Essentials.csproj
+++ b/Xamarin.Essentials/Xamarin.Essentials.csproj
@@ -97,7 +97,6 @@
   </ItemGroup>
   <ItemGroup Condition=" $(TargetFramework.StartsWith('tizen')) ">
     <PackageReference Include="Tizen.NET" Version="4.0.0" />
-    <PackageReference Include="System.Numerics.Vectors" Version="4.4.0" />
     <Compile Include="**\*.tizen.cs" />
     <Compile Include="**\*.tizen.*.cs" />
   </ItemGroup>


rookiejava:
@sung-su It would be better if you provide as much information as you can. Just one line telling someone something isn't working is usually useless. 

@Redth @jamesmontemagno 
Anyway, to be more specific, we found that there are one problem for referencing `System.Numeric.Vectors` package on tizen TFM. 
https://github.com/xamarin/Essentials/blob/a9ba99559f011f7d0b608f405aef42f4c718b4bc/Xamarin.Essentials/Xamarin.Essentials.csproj#L52

Tizen TFM is based on `netcoreapp` (tizen4.0 == netcoreapp2.0), so every tizen devices provides all `System.*.` libraries included in netcoreapp as itself. In other words, tizen .net application doesns't need to add `System.*` packages that are covered by netcoreapp. In case  of `System.Numeric.Vectors`, as you can see [here](https://apisof.net/catalog/System.Numerics.Vector), it is included in netcoreapp (tizen) by default. 

If you include and refer to the `System.Numeric.Vectors.dll` directly in the app package, there may be problems in operation as shown below.  (**multiple versions of assembly System.Numerics.Vectors**)
```
I/DOTNET_LAUNCHER( 8456):
I/DOTNET_LAUNCHER( 8456): Unhandled Exception:
I/DOTNET_LAUNCHER( 8456):
I/DOTNET_LAUNCHER( 8456): System.IO.FileLoadException: ERROR: Native images generated against multiple versions of assembly System.Numerics.Vectors.
I/DOTNET_LAUNCHER( 8456):    at Xamarin.Essentials.AccelerometerData..ctor(Single x, Single y, Single z)
I/DOTNET_LAUNCHER( 8456):    at Xamarin.Essentials.Accelerometer.DataUpdated(Object sender, AccelerometerDataUpdatedEventArgs e) in P:\workspace\github\Essentials\Xamarin.Essentials\Accelerometer\Accelerometer.tizen.cs:line 29
I/DOTNET_LAUNCHER( 8456):    at Tizen.Sensor.Accelerometer.<EventListenStart>b__24_0(IntPtr sensorHandle, IntPtr eventPtr, IntPtr data)
I/DOTNET_LAUNCHER( 8456):    at Tizen.Applications.CoreBackend.UICoreBackend.Run(String[] args)
I/DOTNET_LAUNCHER( 8456):    at Tizen.Applications.CoreApplication.Run(String[] args)
I/DOTNET_LAUNCHER( 8456):    at Tizen.Applications.CoreUIApplication.Run(String[] args)
I/DOTNET_LAUNCHER( 8456):    at Samples.Tizen.Program.Main(String[] args) in P:\workspace\github\Essentials\Samples\Samples.Tizen\Program.cs:line 22
I/DOTNET_LAUNCHER( 8456): onSigabrt called
I/DOTNET_LAUNCHER( 8456): ******************************************************
I/DOTNET_LAUNCHER( 8456): Unhandled exception is occured. check application code
I/DOTNET_LAUNCHER( 8456): ******************************************************
```
This can be a problem especially when using AOT (Ahead Of Time complication). In tizen, actual target device only support AOT not emulator. 

We have found that the Sample has problems when starting accelerometer sensors (that refer to the system package) on the tizen target device, and this PR is a patch to solve this issue. (Explicitly declaring the  `System.Numeric.Vectors` package not to be included in the app)

However, I think this fix is not a fundamental solution because it applies only to the Sample app. 
Instead, I think it is better to modify `Xamarin.Essentials.csproj` so that it does not refer to the `System.Numerics.Vectors` package for Tizen TFM.

jamesmontemagno:
Yes, agree, not sure how that got there to be honest as this should not be here: https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/Xamarin.Essentials.csproj#L52 but instead should really only be added by ios/android

rookiejava:
Well, the reason why we have a dependency with `System.Numerics.Vector3` is that we represent some values coming from sensors. Exactly 4 sensors - that is, Gyroscope, Orientation, Magnetometer and Accelerometer. Those shared implemetations (`Gyroscope.shared.cs`, `OrientationSensor.shared.cs`, `Magnetometer.shared.cs` and `Acceleromete.shared.cs`) are currently using `System.Numerics.Vector3`.

It's Ok. But the problem is that `System.Numerics.Vector3` is only unavailable for .NET Standard 2.0.  As shown below, xamarin.Android, Xamarin.iOS and even .NET Core (Tizen) are support `System.Numerics.Vector3` by itself. (without package dependencies)
- [Xamarin Android 7.1](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector3?view=xamarinandroid-7.1)
- [Xamarin iOS 10.8](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector3?view=xamarinios-10.8)
- [.NET Core 2.0 / Tizen40](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector3?view=netcore-2.0)
- `System.Numerics.Vector3` is available from  .NET Standard 2.1 preview.

As a result, even if I removed package dependency of `System.Numerics.Vector3` from tizen TFM, application has a dependency with `System.Numerics.Vector3` because of application's shared project uses .NET Starard 2.0. See below dependency tree.
package_deps

In my opinion, there 2 choices to resolve this issue as of now.

A)  Keep current status and force all tizen applications to remove `System.Numerics.Vector3` dependency explictly. (just like this PR)

B) Remove `System.Numerics.Vector3` dependency from all implementations. We can declare our own structure (just like [iOS's way](https://docs.microsoft.com/en-us/dotnet/api/coremotion.cmrotationrate?view=xamarin-ios-sdk-12)) for representing sensor values or just use  arrays (just like [android's way](https://docs.microsoft.com/en-us/dotnet/api/android.hardware.sensorevent.values?view=xamarin-android-sdk-9#Android_Hardware_SensorEvent_Values)) instead. 

Any opinions?

jamesmontemagno:
Ohhhhh I see that this is just in the samples. You are correct and that the NuGet will bring in this dependency for people so it is all good.

We want to keep the Vector3 dependency as it has a lot of nice helpers and we dont' want to create another Vector since it already exists.

What would the change to the Essentials.csproj look like?

rookiejava:
@jamesmontemagno OK. If so, we don't need to modify `Essentials.csproj` except removing below line. 
https://github.com/xamarin/Essentials/blob/a9ba99559f011f7d0b608f405aef42f4c718b4bc/Xamarin.Essentials/Xamarin.Essentials.csproj#L100
(@sung-su, would you update this?)

And, please merge this PR.  :)

jamesmontemagno:
Looks like this PR is good to go :)

Issue #863, opened 2019-07-26
Device Architecture


jrahma:
Hi,

Please add an option to get device architecture, e.g. x86, armeabi-v7a, etc

Thanks,
[Jassim](https://www.jassimrahma.net)

dotMorten:
@jrahma I was looking at the PR, and I was a little bit confused about what the PR is trying to attempt, and it seems to do different things on different platforms.
I think if there was some solid use-cases for why this is needed, it would be easier to design the API specifically around that.

Could you clarify why you need to get the device architecture? What sort of experience or app are you trying to build where this is required, and why?

jrahma:
@dotMorten I am trying to develop a device information app using Xamarin Forms

dotMorten:
Just to play devils advocate then: I you are planning on creating an app that shows very device specific information, would it be reasonable to expect those APIs to be available cross-platform, as many many things would be very specific to the individual platform?

Also what sort of information are you looking for in this specific case: Whether the app is 32 or 64 bit, whether the device is 32 or 64 bit, and what sort of processor architecture it is etc etc.

jamesmontemagno:
After reviewing with the team, we have decided that there isn't a good cross-platform API to understand "supported" vs cpu architecture that is consistent across all of the platforms. We also analyzed the feature and believe that the use case for it is pretty low and the code would be easy enough for a developer to implement if they need this specific functionality. For those reasons we are going to close it out.

PR #894, opened 2019-08-26
GH-863 CPU Architecture

pictos:
- Updated the Sample project

### Description of Change ###

Added an new Property to expose the device's CPU Architecture.

### Bugs Fixed ###

- Related to issue #863 

### API Changes ###

Added: 
 
- `public static string CpuArchitecture => GetCpuArchitecture();`

### PR Checklist ###


jamesmontemagno:
I approve of this, we can pull it into a branch and finish off documentation on this. Thanks for the changes! Looking great! :)

dotMorten:
Shouldn't this use the .NET Standard enum instead? https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.architecture?view=netstandard-2.0
By not parsing the various platform-specific strings, you're just putting that work on the end-developer instead.
I'm also curious what the user-story behind this method is meant for. When do you need to check for this? Understanding that might make it more clear whether you'd prefer an enum or the raw platform-specific string version (I don't see the 'why' anywhere in the linked issues)

Xamarin.Essentials/DeviceInfo/DeviceInfo.android.cs
```diff
@@ -22,6 +23,14 @@ static string GetDeviceName()
             return name;
         }
 
+        static string[] GetArchitectures()
+        {
+            if (Essentials.Platform.HasApiLevel(BuildVersionCodes.Lollipop))
+                return Build.SupportedAbis.ToArray();
```
dotMorten: This property returns "Supported ABIs", not the architecture the app runs as. That's different from all the other implementations. If it returned the actual executing ABI, you wouldn't even have to make this an array (It's really odd to me it's an array of architectures as a process can only run as one thing).

pictos: @dotMorten thanks for your review. You're right, ABI is different from CPU architecture (CA). In my research, I found this method that returns the CA
`Java.Lang.JavaSystem.GetProperty("os.arch"));`
but for some devices, this method doesn't return the correct CA, for example, the [Samsung Galaxy S5 Neo (reference](https://stackoverflow.com/a/49740834)). 

So, maybe, the best way to know the CA is using the `Build.SupportedAbis`... I didn't do a verification, as in the example above, because some times returns more than one value (e.g. Motorola One).

dotMorten: I'm a little confused by this. So instead of some devices returning the wrong thing (that's Android for ya), you're going to make almost all devices return a set of things, where all but one are wrong? The intent (I assume) is to get the architecture the app is running under - So a few Android devices are broken in this respects, but that is leading to a design choice that affects all platforms by returning an array instead, and making it useless on Android?

dotMorten: One more thing to clarify: Are you looking at returning the architecture of your app, or the architecture of the device?
(ie you can run x86 apps on x64 PCs, or Armv7 on Armv8, so which one do you want to return?)

pictos: My English is not so good, so sometimes I don't express my self well. So let me rephrase. 
To me returning the supported ABI give, to the developer, a safer result. I mean, Android platform is very fragmented, every day a new device comes out and we don't know how many devices might have the same issue as the Nexus mentioned above.

> One more thing to clarify: Are you looking at returning the architecture of your app, or the architecture of the device?
I assume we are looking for the architecture that the device supports.

If I'm not very clear, please let me know.

pictos: And, gently asking. When I returning arm64-v8a, this is not mean that the **device architecture** is arm 64bits?

jamesmontemagno:
The only thing @dotMorten is that Windows will return other items too: https://docs.microsoft.com/en-us/uwp/api/windows.system.processorarchitecture

Xamarin.Essentials/DeviceInfo/DeviceInfo.android.cs
jamesmontemagno: I think that SupportedABIs is incorrect then and it should be the device architecture.

dotMorten:
@jamesmontemagno Depends on what you're looking for. Are you looking for what the process runs as, or what the device architecture is. You just mentioned that SupportedABIs should be the device architecture, so those extra values that Windows has doesn't apply.

pictos:
@dotMorten and @jamesmontemagno I'll wait for the definition and do the corrections. @dotMorten tks again for the review :)

jamesmontemagno:
I say whatever @dotMorten thinks for the API I am pretty open in general. I think we need to create our own enum though. I am fine maybe doing 2, one is an enum and one is raw like version numbers?

pictos:
@jamesmontemagno, so can I use this [enum](https://docs.microsoft.com/en-us/uwp/api/windows.system.processorarchitecture) as a model?

nosweat:
For android-specific project. i have a version number "132" and when generated a apk per ABI it becomes "400132" / "200132". Now using the getting Current device Build Number, will it return the per abi generated and not the original "132"? Hence i should do this Architecture checking to remove the unnecessary "400" / "200" prefix from the build number? I'm not sure if I asked correctly.

jamesmontemagno:
Thank you for the PR, conversation on the feature, and revving it. After reviewing with the team, we have decided that there isn't a good cross-platform API to understand "supported" vs cpu architecture that is consistent across all of the platforms. We also analyzed the feature and believe that the use case for it is pretty low and the code would be easy enough for a developer to implement if they need this specific functionality. For those reasons we are going to close it out.

Issue #923, opened 2019-10-11
[Enhancement] DeviceInfo -> Detect Device Theme (Dark Mode)


jamesmontemagno:
## Summary
Provide ability to detect the mode/theme of the device. Dark Mode / Light Mode.

## API Changes

```csharp
public enum UserInterfaceStyle
{
   Unknown,
   Default,
   Dark,
   Light
}
```

Add to DeviceInfo:

```csharp
public UserInterfaceStyle UserInterfaceStyle{ get; }
```

Implementation:
* Android: https://github.com/jamesmontemagno/Hanselman.Forms/blob/themes/src/Android/Helpers/Environment.cs
* iOS: https://github.com/jamesmontemagno/Hanselman.Forms/blob/themes/src/iOS/Helpers/Environment.cs
* UWP: https://stackoverflow.com/questions/38734615/how-can-i-detect-windows-10-light-dark-mode

I do not think it is smart that we detect changes as that is at an app level IMHO.

## Intended Use Case
Enable easier theming of light/dark themes of apps.

manishkungwani:
@jamesmontemagno can I take this up?

jamesmontemagno:
It is already in progress @manishkungwani

huangjinshe:
Why did you close it? We really need this when app startup, then we could use the right theme.

DamianMehers:
> Why did you close it? We really need this when app startup, then we could use the right theme.

It looks like it was merged and available as [AppInfo. PlatformRequestedTheme](https://github.com/xamarin/Essentials/blob/develop/Xamarin.Essentials/AppInfo/AppTheme.shared.cs)

huangjinshe:
@DamianMehers  Thanks.

jamesmontemagno:
Yes, there is a bug on iOS currently that will be fixed in 1.5.2 that you have to call it OnStart/OnResume and not in the constructor

huangjinshe:
@jamesmontemagno Thank you for let me know, but do you mean I need to use it in OnStart/OnResume before the bug fixed(after that I can use it anywhere)? or you mean I need to use it always in OnStart/OnResume?

jamesmontemagno:
You can always use it there. Just right now on iOS it won't work in your app constructor.

huangjinshe:
@jamesmontemagno Sure, Thanks again.

PR #927, opened 2019-10-13
[WIP] GH-923 Implement AppTheme for iOS/tvOS/Android/UWP (Need Tizen Still)

jamesmontemagno:
### Description of Change ###

Implement AppTheme for Light/Dark/Unspecified !

### Bugs Fixed ###

- Related to issue #923

### API Changes ###

Added: 
 
enum AppTheme => Light, Dark, Unspecified

Changed:
- AppTheme AppInfo.RequestedAppTheme { get; }
 

### Behavioral Changes ###

none

### PR Checklist ###


Xamarin.Essentials/AppInfo/AppInfo.android.cs
```diff
@@ -48,5 +49,15 @@ static void PlatformShowSettingsUI()
             settingsIntent.AddFlags(ActivityFlags.ExcludeFromRecents);
             context.StartActivity(settingsIntent);
         }
+
+        static AppTheme PlatformRequestedTheme()
+        {
+            return (Platform.AppContext.Resources.Configuration.UiMode & UiMode.NightMask) switch
```
VladislavAntonyuk: Ensure the device is running Android Froyo or higher because UIMode was added in Android Froyo, API 8.0
if (Build.VERSION.SdkInt >= BuildVersionCodes.Froyo)

jamesmontemagno: Xamarin.Essentials only supports KitKat and above, so no need to check it was my thinking

Xamarin.Essentials/AppInfo/AppInfo.ios.tvos.watchos.cs
```diff
@@ -23,5 +23,23 @@ static string GetBundleValue(string key)
         static void PlatformShowSettingsUI() =>
             throw new FeatureNotSupportedException();
 #endif
+
+#if __IOS__ || __TVOS__
+        static AppTheme PlatformRequestedTheme()
+        {
+            if (!Platform.HasOSVersion(13, 0))
+                return AppTheme.Unspecified;
```
sparkist97: shouldn't it be AppTheme.Light? since it is what the OS used pre-13

jrahma:
Hope it will also have Auto so the app will determine if Light or Dark modes depends on the device status just like other apps are doing

VladislavAntonyuk:
@jrahma I think it should be event OnThemeChanged

Xamarin.Essentials/AppInfo/AppInfo.ios.tvos.watchos.cs
jamesmontemagno: That is a good question. I mean technically there was no "theme" at all.... @redth?

Redth: I think unspecified is a sane default here since pre 13 there really wasn't a theme like James said, it was just... the theme.

Redth:
I *think* the only way we can do a notification is if we force users to override their main `UIViewController`'s `TraitCollectionDidChange` method so that they could call something like `Xamarin.Essentials.TraitCollectionDidChange(..)`, since you really need to subscribe to changes on the view controller.

I'm not against this idea necessarily, I'm just not sure if we should do it.  I haven't looked at Android yet but I believe we'd need to do something similar in the main activity's on configuration change method.

The event idea needs more discussion.

jamesmontemagno:
So, on android I think you can handle changes yourself in your app if you add it to the activity, however testing yesterday it just kills your activity and re-inflates it. UWP has an event.

@praeclarum I think said that you can listen to any UIKit element all the way up to UIWindow! So that may be the easiest to do.

Issue #504, opened 2018-09-07
Permissions API


jamesmontemagno:
Currently Xamarin.Essentials prompts users for permissions when needed for specific APIs and will throw a permissions exception if the user denies it. There is no current implementation for developers to check permission status and attempt to prompt themselves.

I have a permissions plugin: https://github.com/jamesmontemagno/PermissionsPlugin however I would say that this implementation is not perfect and using an enum for permission is not super ideal because some permissions require objects to be saved around and require info.plist because it isn't super linker safe :(

The recommendation is to enable developers to check and request permissions for the APIs that we have and not attempt to do a generic one. Some thoughts here are to have them be async or should we actually move towards an event based architecture, which is what each platform uses.

Here is what I would do for Android today:

Status:
```csharp
    /// <summary>
    /// Status of a permission
    /// </summary>
    public enum PermissionStatus
    {
        /// <summary>
        /// Denied by user
        /// </summary>
        Denied,
        /// <summary>
        /// Feature is disabled on device
        /// </summary>
        Disabled,
        /// <summary>
        /// Granted by user
        /// </summary>
        Granted,
        /// <summary>
        /// Restricted (only iOS)
        /// </summary>
        Restricted,
        /// <summary>
        /// Permission is in an unknown state
        /// </summary>
        Unknown
    }
```

```csharp

public static class FlashlightPermission
{
     Task<PermissionStatus> CheckStatusAsync();
     Task<PermissionStatus> RequestAsync();
}

public static class LocationWhenInUsePermission
{
     Task<PermissionStatus> CheckStatusAsync();
     Task<PermissionStatus> RequestAsync();
}

```

These two would handle the permissions that are currently used in Xamarin.Essentials

> VS bug [#735658](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735658)

mattleibow:
We could also just include the permission checks in the API:

```csharp
public static partial class Flashlight {
    // existing code
    public static Task TurnOnAsync();
    public static Task TurnOffAsync();

    // new code
    Task<PermissionStatus> CheckStatusAsync();
    Task<PermissionStatus> RequestAsync();
}
```

jamesmontemagno:
Could work out if we standardize this way. We just need to document when we need to call these.

I believe if we keep the functionality in each class small and container then it works out nice.

Perhaps not add a generic API at all and just how you have it.

Mrnikbobjeff:
[...]

PR #949, opened 2019-10-24
Permissions API

Redth:
### Description of Change ###

This adds a number of classes and a Permission API implementation.

This takes the route of using classes to implement the actual permission logic so that we can rely on the linker on iOS to remove the implementation if it's not used, and not trigger the app store to require usage descriptions in the Info.plist.

### Bugs Fixed ###

Issue #504 

### API Changes ###

Added: 
 
```csharp
public class Permissions
{
	public static Task<PermissionStatus> CheckStatusAsync<TPermission>();

	public static Task<PermissionStatus> RequestAsync<TPermission>();

	public static void ShowSettingsUI();

	public abstract partial class BasePermission
	{
		public BasePermission()
		{
		}

		public abstract Task<PermissionStatus> CheckStatusAsync();

		public abstract Task<PermissionStatus> RequestAsync();

		public abstract void EnsureDeclared();
	}

	public partial class Battery : BasePermission
	{
	}

	public partial class CalendarRead : BasePermission
	{
	}

	public partial class CalendarWrite : BasePermission
	{
	}

	public partial class Camera : BasePermission
	{
	}

	public partial class ContactsRead : BasePermission
	{
	}

	public partial class ContactsWrite : BasePermission
	{
	}

	public partial class Flashlight : BasePermission
	{
	}

	public partial class LaunchApp : BasePermission
	{
	}

	public partial class LocationWhenInUse : BasePermission
	{
	}

	public partial class LocationAlways : BasePermission
	{
	}

	public partial class Maps : BasePermission
	{
	}

	public partial class Media : BasePermission
	{
	}

	public partial class Microphone : BasePermission
	{
	}

	public partial class NetworkState : BasePermission
	{
	}

	public partial class Phone : BasePermission
	{
	}

	public partial class Photos : BasePermission
	{
	}

	public partial class Reminders : BasePermission
	{
	}

	public partial class Sensors : BasePermission
	{
	}

	public partial class Sms : BasePermission
	{
	}

	public partial class Speech : BasePermission
	{
	}

	public partial class StorageRead : BasePermission
	{
	}

	public partial class StorageWrite : BasePermission
	{
	}

	public partial class Vibrate : BasePermission
	{
	}
}
```

### Example Usage ###

```csharp
var status = await Permissions.CheckStatusAsync<Permissions.Camera>();
```

```csharp
var status = await Permissions.RequestAsync<Permissions.Camera>();
```

### PR Checklist ###


Xamarin.Essentials/Permissions/Permissions.tizen.cs
```diff
@@ -5,114 +5,200 @@
...
+    public static partial class Permissions
     {
-        static bool PlatformEnsureDeclared(PermissionType permission, bool throwIfMissing)
+        public static bool IsPrivilegeDeclared(string tizenPrivilege)
         {
-            var tizenPrivileges = permission.ToTizenPrivileges(onlyRuntimePermissions: false);
+            var tizenPrivileges = tizenPrivilege;
 
             if (tizenPrivileges == null || !tizenPrivileges.Any())
                 return false;
```
sung-su: how about this ?
+ if (string.IsNullOrEmpty(tizenPrivilege)) return false;

jrahma:
Please add a Callback event to check if the user clicked Allow or Deny.. For example if the app requires Location permission and he clicked Deny then developer can show a (Location is Required) message

Redth:
@jrahma this is already how it will work by design with RequestAsync which only returns a result after the user chooses to allow or deny (or if they previously allowed it). No need for callbacks, we have Task<T>!

jrahma:
What about Health & Motion activity and HomeKit?

Xamarin.Essentials/Permissions/Permissions.shared.cs
```diff
@@ -1,27 +1,131 @@
...
 
-        internal static Task<PermissionStatus> CheckStatusAsync(PermissionType permission) =>
-            PlatformCheckStatusAsync(permission);
+        internal static void EnsureDeclared<TPermission>()
+            where TPermission : BasePermission, new() =>
+                new TPermission().EnsureDeclared();
 
-        internal static Task<PermissionStatus> RequestAsync(PermissionType permission) =>
-            PlatformRequestAsync(permission);
+        public static void ShowSettingsUI() =>
```
jamesmontemagno: Do you think that this is confusing that it may not open the permissions screen? I think it is possible to do on some platforms.

jamesmontemagno:
I think just needs to be merged up, but looks good to me. Left a comment here and there, but else I am good. We will need to heavily document this API though since all platforms don't support everything.

Samples/Samples/Model/PermissionItem.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Text;
+using System.Windows.Input;
+using Xamarin.Essentials;
+using Xamarin.Forms;
+
+namespace Samples.Model
+{
+    public class PermissionItem : INotifyPropertyChanged
```
mattleibow: We can use BaseViewModel so we don't need to dupe code.

JKennedy24:
I tried to refactor out my [Permissions API](https://github.com/jamesmontemagno/PermissionsPlugin) for this today following the steps highlighted in the [releasenotes](https://docs.microsoft.com/en-gb/xamarin/essentials/release-notes/1.4)

```
PermissionStatus status = await Permissions.CheckStatusAsync<Permissions.LocationWhenInUse>();
if(status != PermissionStatus.Granted)
{
    status = await Permissions.RequestAsync<Permissions.LocationWhenInUse>();
}

```
I ran it on UWP and got a crash on the `CheckStatusAsync` line

> Object reference not set to an instance of an object.
>    at Xamarin.Essentials.Permissions.BasePlatformPermission.EnsureDeclared()
>    at Xamarin.Essentials.Permissions.BasePlatformPermission.CheckStatusAsync()
>    at Xamarin.Essentials.Permissions.CheckStatusAsync[TPermission]()
>    at FieldStrikeMove.Business.Helpers.PhotoSavingHelper.<CanUseCamera>d__0.MoveNext() in 

not sure what that error relates to, I couldn't see any initialization I had to do in the release notes

I also can't find the method `ShowSettingsUI()` which is provided in the alternative

JKennedy24:
actually on reading this: https://docs.microsoft.com/en-us/xamarin/essentials/permissions?context=xamarin%2Fandroid&tabs=ios#available-permissions

I was requesting `Camera` which I can see is not supported on UWP

jamesmontemagno:
@JKennedy24 this was filled and fixed in the next release 1.5.1 where you wont get the exception. see: https://github.com/xamarin/Essentials/issues/1129

minaairsupport:
Hey James we need to get access for `ShowSettingsUI()` , and if possible to make CheckStatusAsyn accept multiple permissions

jamesmontemagno:
AppInfo.ShowSettingsUI will do that for you. We will not do multiple permissions, however you can extend it https://docs.microsoft.com/en-us/xamarin/essentials/permissions?context=xamarin%2Fxamarin-forms&tabs=android#extending-permissions

zizusoft:
I’m assuming there is a explanation for this one - but is there any reason permissions for Notifications are missed off this api? 

Thanks

jamesmontemagno:
No specific reason, however they are very much iOS specific and there are also a lot of additional flags that need to be set. So really this is best to just keep in the iOS project or custom per app. https://docs.microsoft.com/en-us/xamarin/ios/platform/user-notifications/deprecated/local-notifications-in-ios

jamesmontemagno:
Somethign like this is what you would be looking for: https://github.com/xamarinhq/app-conference/blob/c379ccaf2c72c7c1005a03a5f7345aa9a2bdecd1/src/Conference.iOS/Helpers/PushNotifications.cs

I am also not sure if you get a call back when they register or not to be honest.

darturdotnetdeveloper:
hi james. I have this problem with Xamarin.Essentials:Permission when my app is freshly installed and first time to be open and i click a button to do the requestAsync well it doesnt continue to its next line of code. But if i close the app and open it again it works. can you please enlighten me how to fix this one. TIA

P.S. I already do all the needed thing to work with permissions.

jamesmontemagno:
Make sure you have fully integrated the permission callback into your app code: https://docs.microsoft.com/en-us/xamarin/essentials/permissions?context=xamarin%2Fxamarin-forms&tabs=android#get-started

PR #954, opened 2019-10-28
Make GetCurrentActivity & GetCurrentViewController public

tuyen-vuduc:
### Description of Change ###
The use of CurrentViewController/CurrentActivity is very common when writing custom platform code.
Normally, we have to the other lib like CurrentActivity and/or Acr.Support library. However, inside this library, we also have the same methods serving that purpose, but they are internal.

Making those methods public will get rid of the other libraries.

### Bugs Fixed ###

None

### API Changes ###

Added: 
 
None

Changed:

 - `internal static Activity GetCurrentActivity(bool throwOnNull)` => `public static Activity GetCurrentActivity(bool throwIfNull = true)`
 - `internal static Activity GetCurrentViewController(bool throwIfNull = true)` => `public static Activity GetCurrentViewController(bool throwIfNull = true)`
 

### Behavioral Changes ###

None

### PR Checklist ###


Redth:
I think we should instead probably add method overloads which don't take the `throwOnError` parameter (or at least default it to false).

tuyen-vuduc:
Hi @Redth,

On iOS, `throwIfNull` is default to `true`.

```
public static UIViewController GetCurrentViewController(bool throwIfNull = true)
```

I think we should do the same to avoid other code change.

How do you think?

Redth:
@tuyen-vuduc yeah good catch, I guess we should keep the other platforms defaulting to `true` as well.  I'm tempted to add properties to expose this publicly like:

```csharp
public static UIViewController CurrentViewController { get; }

public static Activity CurrentActivity { get; }

public static Context CurrentContext { get; }
```

All which return `null` if we can't find one.

jamesmontemagno:
We require @aritchie here.. My thought is bring over the current activity API. How about for IOS?

aritchie:
For iOS, the "GetCurrentViewController" has been a constant moving target over the years.  Even the current method of doing it still tends to have issues when called within modals and such.  You will hit corner cases with it.  

For Android, even top activity is going to be weird with multi-window incoming, but I haven't had a complaint about it thus far.

jamesmontemagno:
Thanks for the PR on this. We discussed in the team and finalized a full PR with expanded functionality: https://github.com/xamarin/Essentials/pull/997

Issue #631, opened 2018-11-28
Make Essentials ready for non-nullable reference types.


Mrnikbobjeff:
Some projects such as YodaTime and similar use [NotNull] and  [CanBeNull] or similar attributes to inform users about contracts. As C# 8.0 brings non-nullable reference types, I thought it might be a good idea to create contracts now so that once 8.0 rolls out we simply can remove the attributes and use the corresponding c# 8.0 nullable syntax. In general I like contracts very much, especially the Resharper info about the bcl function contracts where resharper detects which apis can be called with null and which can not.

> VS bug [#736800](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/736800)

Redth:
This is a good idea.  First step is to just do an audit on the API for any cases where this will need to be applied.

Mrnikbobjeff:
The following API's validate arguments to be not null or should validate: Browser, Clipboard, Connectivity (ConnectionProfiles), Email, Launcher, MainThread, Maps, PhoneDialer, Preferences, SecureStorage, Share, TTS
The following API's have return arguments which would have to be enforced non-nullable (e.g. by returning empty enumerables): AppInfo, DeviceInfo, Geocoding, Geolocation, Sms
All our event based classes are inherently non-nullable as the EventArgs are constructed by value types and thus can not ever be null.

Redth:
So I'm assuming you are proposing defining our own `[NotNull]` and `[CanBeNull]` attributes?  Are these attribute _names_ something like resharper is looking for?

Mrnikbobjeff:
Yes I propose defining them in our own code - this is something which still works with resharper, even if we dont use their predefined attributes. As long as the names match everything is fine. In general I would see the need for four attributes, [CanBeNull],[NotNull], [ItemNotNull] and [ItemCanBeNull]. The first two are used for single elements, the others for collections which can / can not contain null values. A combination of these should be able to let us specify nullability behaviour just as precise as C# 8.0.

Mrnikbobjeff:
Sooo, I installed the preview and we have 130 errors with non-nullable references. Many are hidden from the user in the Platform implementations in the start and stop events. The listeners are assigned null after stopping the sensor and have to be nullable. The main other concern are our DTO's such as EmailMessage or Placemark. We should discuss whether whether we want these objects to implement nullable properties or to remain empty if they  are uninitialized. Other than that we directly pass null in preferences, all other instances are encapsulated in dto's.

[...]

PR #966, opened 2019-10-30
[WIP] Bring nullable reference types to Xamarin.Essentials

Mrnikbobjeff:
### Description of Change ###
This  is the conversion towards nullable reference types. There are still some unresolved issues where I supressed nullable reference types as the solution is not obvious. This also includes minor changes such as calling ToString on primitives before using them in string.Format to avoid boxing them beforehand. This pr also aims to be completely stable considering the public api considering exceptions, even though this will still throw NullReferenceExceptions in ugly places.
Finally, we would have to update the stylecop (or editorconfig, no idea) to be amenable to the construct of nullable arrays, as it always complains about a space missing when writing 

> object[]? arr = null;

- Related to issue #631

### Behavioral Changes ###
Different exception locations when passing invalid values to xamarin essentials.

### PR Checklist ###


Mrnikbobjeff:
Okay, so one issue I had was that the preferences code uses generics which are not constrained to be reference or value types. As we accept reference and value types as parameters, we can not use nullable syntax here (T? is unavailable as it could be Nullable<ValueType> or just syntax sugar). I have no clue how to solve that one. Also, we should check for null in a couple of places as we rely on certain values to not be null. I "fixed" a couple of those here and there by simply using the null conditional operator, but this will likely just pass this exception on towards the first framework method which validates that the parameters are not null

Xamarin.Essentials/Platform/Platform.tizen.cs
```diff
@@ -76,11 +77,11 @@ internal static Sensor GetDefaultSensor(SensorType type)
                         Platform.orientationSensor = new TizenOrientationSensor();
                     return Platform.orientationSensor;
                 default:
-                    return null;
+                    throw new InvalidOperationException("switch case should be exhaustive.");
```
Mrnikbobjeff: switch case is currently exhaustive, should not change behaviour and safeguard us against forgetting to amend that switch case and it failing silently

Xamarin.Essentials/Preferences/Preferences.ios.tvos.watchos.cs
```diff
@@ -2,6 +2,7 @@
 using System.Globalization;
 using Foundation;
 
+#nullable disable
 namespace Xamarin.Essentials
 {
     public static partial class Preferences
```
Mrnikbobjeff: I have no clue how to correctly differentiate between Nullable<ValueType> and syntax sugar, so I just disabled it until I know how to handle it

Xamarin.Essentials/Share/Share.ios.cs
```diff
@@ -36,9 +37,9 @@ static Task PlatformRequestAsync(ShareFileRequest request)
         {
             var items = new List<NSObject>();
 
-            var fileUrl = NSUrl.FromFilename(request.File.FullPath);
-            if (!string.IsNullOrEmpty(request.Title))
-                items.Add(new ShareActivityItemSource(fileUrl, request.Title)); // Share with title (subject)
+            var fileUrl = NSUrl.FromFilename(request?.File?.FullPath);
```
Mrnikbobjeff: Will throw on null with different stack trace than e.g. android or tizen

Xamarin.Essentials/TextToSpeech/TextToSpeech.shared.cs
```diff
@@ -15,29 +15,29 @@ public static partial class TextToSpeech
...
         {
             if (string.IsNullOrEmpty(text))
                 throw new ArgumentNullException(nameof(text), "Text cannot be null or empty string");
 
             if (options?.Volume.HasValue ?? false)
             {
-                if (options.Volume.Value < VolumeMin || options.Volume.Value > VolumeMax)
-                    throw new ArgumentOutOfRangeException($"Volume must be >= {VolumeMin} and <= {VolumeMax}");
+                if (options!.Volume! < VolumeMin || options.Volume!.Value > VolumeMax)
+                    throw new ArgumentOutOfRangeException($"Volume must be >= {VolumeMin.ToString()} and <= {VolumeMax.ToString()}");
```
Mrnikbobjeff: added tostring to avoid boxing of value types

Xamarin.Essentials/Types/Placemark.shared.cs
```diff
@@ -30,34 +31,53 @@ public Placemark(Placemark placemark)
...
-        public override string ToString() =>
-            $"{nameof(Location)}: {Location}, {nameof(CountryCode)}: {CountryCode}, " +
-            $"{nameof(CountryName)}: {CountryName}, {nameof(FeatureName)}: {FeatureName}, " +
-            $"{nameof(PostalCode)}: {PostalCode}, {nameof(SubLocality)}: {SubLocality}, " +
-            $"{nameof(Thoroughfare)}: {Thoroughfare}, {nameof(SubThoroughfare)}: {SubThoroughfare}, " +
-            $"{nameof(Locality)}: {Locality}, {nameof(AdminArea)}: {AdminArea}, " +
-            $"{nameof(SubAdminArea)}: {SubAdminArea}";
+        public override string ToString()
+        {
+            var sb = new StringBuilder();
```
Mrnikbobjeff: replaced huge strinig format+concat with stringbuilder. StringJoin is not available sadly

Xamarin.Essentials/Preferences/Preferences.ios.tvos.watchos.cs
Mrnikbobjeff: Okay, after some further investigation and thought I have come to a conclusion. There are two possible ways of handling this. Either constrain two methods to either struct or class and use them directly, e.g. PlatformGetValueType<T> and PlatformGetString(), so that we can annotate and use nullable directly, or use the null forgiving operator to lie to the compiler that a value passed is not null even though it may be, thus keeping the method signature intact. This would also mean lying when returning a value, since null may be returned. This second approach could be contained to the internals so that our public interface is null compliant, internally this would feel a little bit hacky though

Mrnikbobjeff:
@mattleibow I would love some advice on how to handle all things touching preferences before doing the work. My advice would be to constrain two methods to different parameters and thus correctly differentiate between Nullable<ValueType> and e.g. string?

PR #971, opened 2019-11-04
Fix broken docs link

mrlacey:
Can't have a relative path to the wiki here


mrlacey:
failed CI tests are unrelated to the change made here.

Issue #979, opened 2019-11-11
[Bug] Share Text does not work on iPadOS 13.2 & 13.2.2


bobwhitten:
### Description
Share Text for iPadOS 13.2 and 13.2.2 does not bring up "pop-over" window, so does not function.

It appears to work fine on iPhone and and versions iPadOS 13.1 and previous.

Asked on [Xamarin Forum](https://forums.xamarin.com/discussion/173535/share-from-xamarin-essentials-mysteriously-quit-working-after-updates-but-only-on-ipad-on-ios13-2).

### Steps to Reproduce

1.   Attached is sample program -- empty Blank XF program with single button added to share text.
2.   Click on button, nothing happens on iPadOS 13.2, but works on iPhone, older iPads.

### Expected Behavior
Pop-Over (activity controller) allows shared text to be sent to various apps

### Actual Behavior
Nothing happens

### Basic Information

- Version with issue:  1.3.1,  XF 4.2.910310
- Last known good version:  unknown
- IDE:   VS for PC 16.3.7
- Platform Target Frameworks: 
  - iPadOS:  13.2 and 13.2.2
- Affected Devices:
     iPad Air

### Reproduction Link

[ShareTest.zip](https://github.com/xamarin/Essentials/files/3833573/ShareTest.zip)

iWeaverMan:
Share File also does not work

AbabeiAndrei:
As a workaround/hack until the PR is approved you can call the Share.RequestAsync twice. For some reason, it works this way, to some extent.

```
await Share.RequestAsync(new ShareTextRequest
{
    Text = "Share me!"
});

if (Device.Idiom == TargetIdiom.Tablet)
{
    //todo: change after PR #979 for Xamarin.Essensialas is intergrated (https://github.com/xamarin/Essentials/issues/979)
    await Share.RequestAsync(new ShareTextRequest
    {
        Text = "Share me!"
    });
}
```

bobwhitten:
I've done a little more investigation of this. On my iPad with 12.4.2, the "SourceRect" in the native code is "X=0,Y=0,Width=0,Height=0".  On my iPad with 13.2, it is "X=Infinity,Y=Infinity,Width=0,Height=0".   

And, now that I think about it, the system "feels" like there is a non-visible pop-up up.

I think if nothing else changes, I will change my PR to check for infinities and replace with 0, unless there is a better idea.  (For example, Issue #944 suggests that the pop-up should be top-right).

[...]

Issue #944, opened 2019-10-22
[Bug] Positioning of Share dialog in iPad iOS is top left instead of top right When using Xamarin.Essentials share feature


poliahrameshkumar:
### Description
Share dialog in iPad iOS is positioned on top left instead of top right When using Xamarin.Essentials share feature.
### Steps to Reproduce

1. Follow the official link https://docs.microsoft.com/en-us/xamarin/essentials/share?tabs=ios to create share option in iOS iPad device
2. The share dialog is positioned in top left instead of top right

### Basic Information

- Version with issue: Xamarin.Essentials latest version
- Last known good version: No good version
- IDE: Visual Studio
- Platform Target Frameworks: 
  - iOS:  13.0
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages: xamarin.essentials
- Affected Devices: All iPads

Redth:
It seems like this is because we are setting the SourceView to be effectively the entire view of the view controller: https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/Share/Share.ios.cs#L29

I guess this probably grabs 0,0 as the location, hence the top left.  Since we don't really know which view you want it appearing from, it's not really something we can infer.

We could allow a global static property to override the location I suppose, OR maybe we could add an overload which takes a Forms `View` and then on iOS we'd grab the native view and use that as the source.

@mattleibow @newky2k @jamesmontemagno any thoughts?

poliahrameshkumar:
Could you guide me how to to achieve setting global static property or overload forms view?

poliahrameshkumar:
@jamesmontemagno can you help me on how to bring share dialog on top right corner instead of top left corner on iPad?

bobwhitten:
PR #980 will put the share popover at top-right, which would close this issue, I believe.

mattleibow:
I got some comments here that may be useful to discuss: https://github.com/xamarin/Essentials/pull/980#issuecomment-554398486

jamesmontemagno:
Fixed with https://github.com/xamarin/Essentials/pull/980

PR #980, opened 2019-11-12
Allow ShareText, Share Fix on iPadOS 13.2, and popover at top-right

bobwhitten:
Thanks to LandLu from forums.xamarin.com 

### Description of Change ###

Add ability to do Share functionality for both Text & File on iPadOS 13.2 and 13.2.2, and position the "Share" activity popover in the upper right of screen.

Background: In iOS 12 and before, the "SourceRect" for the PopoverPresentationController is not set, and appers to default to (X=0,Y=0,Width=0,Height=0).  But in iPadOS 13.2 and 13.2.2, the default appears to be (X=Infinite, Y=Infinite, Width=0,Height=0).  This results in the Activity popover not appearing on the screen.

Solution:  Set the SourceRect to (X={width of current view controller}, Y=0, Width=0, Height=0).  This will put the Activity popover at the top right of the iPad screen.

No tests added -- not needed.

No samples added -- current sample tests this functionality.

### Bugs Fixed ###

Addresses issue #979 and #944

### API Changes ###

none

### Behavioral Changes ###

Position "share" popover in top right of screen.

### PR Checklist ###

- [ ] Updated documentation ([see walkthrough]


Redth:
Can you explain this a little bit more?  Does this not just change where the popup appears on iPad?

mattleibow:
I think we have the opportunity to fix this properly.

Let's take the iPad Pro for example - this is a big device.

This issue just brings up a current limitation of the API. Where _should_ the position be? At the top? What about if I have a button at the bottom and the user has focused there? I tap that button and then nothing happens. I have to move my eyes and try and find a popup.

> As a side note, this is a bigger issue on macOS since the share popup is tiny, and the screens are huge.

There are options to take a dependency on some UI platform and pass views and things, but I think we can do better. All we really need is a position/rect. A position may fine though.

There are actually 2 types of position that a user may care about - approximate and exact. For both cases, we could just have an extra property in the share options, so this is not too invasive.

For approximate, we could have an enum with a set of values, such as: `TopLeft`, `TopRight`, `TopCenter`, `Center`, `BottomCenter`, etc. This is simple and we can calculate it on each platform.

For exact, I would avoid taking any dependencies in Essentials, so we will have to rely on the platforms to do it. Most platforms have a way to get screen coordinates of a local position, so we are good there. With the Xamarin.Forms, we may need to ask the team to add a method to calculate/fetch the screen position:

```csharp
// var get forms view position
View formsView = ...
Pont screenPoint = formsView.GetScreenPosition();

// share something
await Share.RequestAsync(new ShareTextRequest
{
    Text = text,
    Title = "Share Text"
    Position = screenPoint
});
```

Redth:
@mattleibow maybe for `TopLeft`, `TopRight`, `Center`, etc. we could have those actually calculate and return the `Point` so the api is one property to set regardless of how precise you want to be or not.

bobwhitten:
I concur that the Share would be more complete with @mattleibow's suggestion.  And the idea of having "pre-cooked" options like top-right, top-center, top-left, etc. would be good too.

For iOS/iPadOS on iPad, at least, my proposed solution is not the best, because of this positioning issue.   The "right" solution would be to pass a point or rect -- The OS brings up the popup looks like it is current modelled after a speech bubble, with the tail coming from the view that initiated it.   My PR would make it point toward upper right corner.

popup

jamesmontemagno:
I like the idea of pre-defined positions on each platform (even though technically this only applies to iOS). I think this also needs to be applied to Launcher APIs.

Love it.

mattleibow:
@bobwhitten would you be interested in having a stab at adding support for actual points? Maybe have a  look at @Redth's comment (https://github.com/xamarin/Essentials/pull/980#issuecomment-554402197) and see if you can use that or improve?

bobwhitten:
in re: @mattleibow question -  Yes, I think I could take a shot it at this. But, in discussion of the bug (#979),  @daniels7 has expressed some urgency for the current fix which doesn't include any update to API.  Would it make sense to push the PR as is, then open a new PR for the enhanced capability?

mattleibow:
@JamesMontemagno this may be your call.

mattleibow:
Just thinking what will happen if the share button is at the bottom of the screen. This pop-up will still appear at the top?

bobwhitten:
I have a interesting data point on this issue -- I have an older version (I built it on 10-oct-19) of my code that I recently had to install.   Works fine. 

So, now this seems like it's not necessarily JUST iPadOS 13.2, but the possibly the 13.2 SDK as well? This doesn't fit all the evidence (new apps work OK on older versions of iPadOS), but maybe this is part of the mystery?   Have others seen this?

Redth:
That is an odd change by Apple. 

I think we should update this PR with a `ScreenLocation` property taking a `Point` to use and let the developer take care of calculating where they want it. Let’s add sensible values for `Center`, `TopLeft`, `TopMiddle`, `TopRight`, `BottomLeft`, `BottomMiddle`, `BottomRight`. 

@bobwhitten do you want to take a shot at this or would you like us to start a new PR?

bobwhitten:
I think that the current PR addresses an urgent problem, and is pretty simple.   Adding ScreenLocation adds new functionality which is good, but only applies to iPad, as far as I know.  And, it would require updates to Share.shared.cs, sample code, and documentation as well.   So moving ahead as it is not a bad idea.

That said, I'll work on the ScreenLocation.

Redth:
Superceded by #1002

Issue #978, opened 2019-11-10
Add Weight conversion to UnitConverters


munkii:
Bug report best practices: [Submitting Issues](https://github.com/xamarin/Essentials/blob/master/SUBMITTING-ISSUES.md)

### Description
Xamarin Essentials currently has conversions for distance, temperature and others. I feel that it should include Weight to such as Kilograms to Pounds. To be honest I have forked the repro here https://github.com/munkii/Essentials and was about to submit a PR when I read the contribution rules. 

If I understand them correctly you want something raised as an issue before it is considered for a PR

### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

Add conversions for

- Pounds to Kg
- Pounds to stones
- Stones to pounds 
- Kg to pounds

jamesmontemagno:
Send the PR! :)

jamesmontemagno:
Merged and awaiting release

PR #983, opened 2019-11-13
Add Weight conversion functionality

munkii:
### Description of Change ###

Added Weight conversion functionality similar to the distance and other Unit Conversion methods
Add Unit Tests for these new conversions

### Bugs Fixed ###

#978 
https://github.com/xamarin/Essentials/issues/978

### API Changes ###

Added: 

```csharp
    public static double PoundsToKilograms(double pounds)
    public static double PoundsToStones(double pounds)
    public static double StonesToPounds(double stones)
    public static double KilogramsToPounds(double kilograms)
```


jamesmontemagno:
Will need API docs generated and we should just need to fix the conflict in that file, easiest to just override what you had with master.

Issue #385, opened 2018-07-17
Add Application/Screen Size API


KSemenenko:
Currently the DeviceDisplay is for the actual device, not the app. It would be nice to have an API to get the application size, or current view size if you will.

Reference:
* UWP: [Application View](https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.applicationview)
* Android: I think Display Metrics
* iOS: [UIScreen](https://developer.apple.com/documentation/uikit/uiscreen)

//Original Comment to consider
I think that "Device Display Information" does not have enough information about SafeArea on the iPhone.
It contains "metrics.Orientation" and I think that it would be nice to get "metrics.SafeArea", which would be of the type Thickness.

Redth:
We are closing this issue since no response was received.  Please open a new issue referring to this one if you have more information.

mattleibow:
@Redth your bot is too aggressive with the closing ;)

donaldnp:
+1 for this feature. I want to adjust what's displayed in my UI according to the actual app window size. I just tried using the DeviceDisplay in Essentials thinking this is what it would tell me, but found it gives the actual screen size, not the size of your app window - unless I'm missing something, I don't see anything in Essentials to get the current app window size.

Redth:
Yeah this would be good to add.

pictos:
@Redth, I did some tests and I thought of this spec: 

(The events were inspired from `DeviceDisplay` implementation)

```csharp
class AppInfo
{
      // ...
      public static WindowSize WindowAppSize() => PlatformWindowAppSize();
      public static event EventHandler<WindowSizeChangedEventArgs> WindowSizeInfoChanged
        {
            add
            {
                var wasRunning = WindowSizeChangedInternal != null;

                WindowSizeChangedInternal += value;

                if (!wasRunning && WindowSizeChangedInternal != null)
                {
                    SetCurrent(GetWindowSizeInfo());
                    StartScreenMetricsListeners();
                }
            }

            remove
            {
                var wasRunning = WindowSizeChangedInternal != null;

                WindowSizeChangedInternal -= value;

                if (wasRunning && WindowSizeChangedInternal == null)
                    StopScreenMetricsListeners();
            }
        }

        static void OnWindowSizeChanged(DisplayInfo metrics)
            => OnWindowSizeChanged(new WindowSizeEventArgs(metrics));

        static void OnWindowSizeChanged(WindowSizeEventArgs e)
        {
[...]

Issue #358, opened 2018-07-07
Add Application State Event API


aritchie:
## Description of the Feature

Events for detecting app state transitions (foregrounding & background) in a global fashion.  Allows for good decoupling of services that need to clean up or do something (ie. send a notification as an app goes to background)

## API Declaration

Integration Point: AppInfo

```csharp
public static AppState State { get; private set; }

public delegate void AppStateChangedEventHandler(AppStateChangedEventArgs e);

public class AppStateChangedEventArgs : EventArgs
{
    internal AppStateChangedEventArgs (AppState appState) => AppState = appState;
    public AppState State { get; }
}

public enum AppState
{
    Foreground,
    Background
}
```

## Documentation for supported platforms

* iOS: https://developer.apple.com/documentation/uikit/uiapplication/1622944-willenterforegroundnotification, 
* Android: Requires a bit of black magic, but the necessary mechanics are already in essentials with ActivityLifecycleContextListener - https://github.com/aritchie/deviceinfo/blob/master/Plugin.DeviceInfo/Platforms/Android/AppStateLifecycle.cs
* UWP: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.window.visibilitychanged

Current implementation to port from: https://github.com/aritchie/deviceinfo/tree/master/Plugin.DeviceInfo

EDIT: Updated by @redth to finalize API spec

jamesmontemagno:
I would also add

```csharp
public static AppState State {get; private set;}
```

Then I would use do a custom EventHandler to follow our pattern:

```csharp
 public delegate void CompassChangedEventHandler(CompassChangedEventArgs e);

    public class AppStateChangedEventArgs : EventArgs
    {
        internal AppStateChangedEventArgs (AppState appState) => AppState = appState;

        public AppState State { get; }
    }
```

jamesmontemagno:
I think looks pretty straight forward to me @Redth @mattleibow can comment.

Mrnikbobjeff:
@aritchie Even though I generally support your proposal this will be in conflict with @jamesmontemagno 's proposal to add the State property. All platforms would need to hook to keep the property up to date. Not sure this is possible then.

aritchie:
@Mrnikbobjeff Only android requires a hook, but it gets built into with the top activity mechanics in any case.  The state property can be checked without any hooks on UWP & iOS.

Mrnikbobjeff:
[...]

PR #993, opened 2019-11-17
GH-385 - Add method to get the application window's size

pictos:
### Description of Change ###

Created a new method inside of `AppInfo` that returns the `Width` and `Height` from the current application window's size.

### Bugs Fixed ###

- Related to issue #385 

### API Changes ###

Added: 
 
- `WindowSize PlatformWindowAppSize()`

### Behavioral Changes ###

### PR Checklist ###


Xamarin.Essentials/Types/WindowSize.shared.cs
```diff
@@ -0,0 +1,35 @@
...
+            Width.Equals(other.Width) && Height.Equals(other.Height);
+
+        public override bool Equals(object obj) =>
+            (obj is WindowSize windowSize) && Equals(windowSize);
+
+        public override int GetHashCode() =>
+            (Width, Height).GetHashCode();
+
+        public override string ToString() =>
+            $"{nameof(Width)} : {Width} \n {nameof(Height)}:{Height}";
```
DamianMehers: `{Environment.NewLine}` instead of `\n`?

jamesmontemagno:
Should call it `AppViewInfo` just like `DisplayInfo` this way it can hold more information about the view in the future.

Xamarin.Essentials/AppInfo/AppInfo.android.cs
```diff
@@ -48,5 +51,16 @@ static void PlatformShowSettingsUI()
             settingsIntent.AddFlags(ActivityFlags.ExcludeFromRecents);
             context.StartActivity(settingsIntent);
         }
+
+        static WindowSize PlatformWindowAppSize()
+        {
+            var context = Platform.GetCurrentActivity(false) ?? Platform.AppContext;
+            var windowManager = context.GetSystemService(Context.WindowService);
+            var windows = windowManager.JavaCast<IWindowManager>();
```
jamesmontemagno: needs using statements here. We also have helpers that get a lot of this information in displayinfo, may want to see if we have some re-use there.

Xamarin.Essentials/AppInfo/AppInfo.netstandard.cs
```diff
@@ -11,5 +11,7 @@ public static partial class AppInfo
         static string PlatformGetBuild() => throw ExceptionUtils.NotSupportedOrImplementedException;
 
         static void PlatformShowSettingsUI() => throw ExceptionUtils.NotSupportedOrImplementedException;
+
+        static WindowSize PlatformWindowAppSize() => throw ExceptionUtils.NotSupportedOrImplementedException;
```
jamesmontemagno: I am curious if we should be returning an IEnumerable of this... Apps can now have multiple windows....

jamesmontemagno: For start we can do "

static AppViewInfo CurrentView {get;}

And then in the future we could return IEnumerable based on demand. This is how we structured the displays

Xamarin.Essentials/Types/WindowSize.shared.cs
mattleibow: Do we need a new size type? Can we not just use `Size`?

mattleibow: If anything, I would copy corefx: https://github.com/dotnet/corefx/blob/master/src/System.Drawing.Primitives/src/System/Drawing/Size.cs#L192

pictos: @mattleibow I tried to use this `Size`, that you mentioned above. But I got compile errors from `netStandard1.0`

Xamarin.Essentials/AppInfo/AppInfo.android.cs
pictos: I can use the methods in `DeviceDisplay.Android.cs`, tks for the tip.

pictos: @jamesmontemagno, I realized if I use the method `DeviceDisplay.GetDefaultDisplay()` it will return the wrong value (returns the screen size). I think the reason it's because the method uses the `ApplicationContext` and to get the app's window a need to use the `CurrentActivity`.

Mrnikbobjeff:
Does this work for picture in picture or however the android mode is called? What is with split screen? I remember some oddities when working on #358

pictos:
@Mrnikbobjeff, for now, I just implement a handle to get the actual screen size. After this first step gets approval I'll implement the event handlers to get that value in real-time change. 

In Android, I tested using the split-screen, until now works fine.

Redth:
@mattleibow @jamesmontemagno are we sure we want a new type for this instead of just `Size` ?  I'm not sure that makes sense.  Is there opportunity to add more info to this class which would make sense?

jamesmontemagno:
I am not sure if we should put this into "AppInfo" to be honest. I am almost thinking that it should be:

```csharp
public class AppWindow
{
    static WindowInfo MainWindowInfo {get;}
}

public readonlystruct WindowInfo : IEquatable<WindowInfo>
{
        public double Width { get; }

        public double Height { get; }
}
```

Now, if we would like, we could change it to a Size, but wouldnt' be in .net standard 1.0 :( other thing is if we make it WindowInfo then we can easily extend it with more properties if we needed to.

This would allow us to add

mattleibow:
@jamesmontemagno you got a point there. We may actually want to add extra properties. One useful one (not for this PR) may be `IsFullscreen` from UWP.

Xamarin.Essentials/Types/AppViewInfo.shared.cs
```diff
@@ -0,0 +1,35 @@
...
+        public override bool Equals(object obj) =>
+            (obj is AppViewInfo windowSize) && Equals(windowSize);
+
+        public override int GetHashCode() =>
+            (Width, Height).GetHashCode();
+
+        public override string ToString() =>
+            $"{nameof(Width)} : {Width} \n {nameof(Height)}:{Height}";
+
+        public static bool operator ==(AppViewInfo left, AppViewInfo right) =>
```
Mrnikbobjeff: This will allocate two boxes, as both sides are structs - just use left.Equals(right)

jamesmontemagno:
Going to pull this into a branch and finish it off over the holiday.

varyamereon:
Umm... what happened to this? I see it got merged but can't find it in the library. 🤔

PR #1002, opened 2019-11-25
Add Share Source Position

Redth:
### Description of Change ###

Adds the ability to specify a Source Rectangle for Share*Requests.  This is really only used for now on iOS where on some iOS 13 versions it seems the `sourceRect` needs to be set in order for the dialog to display.

This should address #980 #979 #944 by allowing users to set their own source position.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Share/Share.shared.cs
+++ b/Xamarin.Essentials/Share/Share.shared.cs
@@ -29,7 +29,7 @@ namespace Xamarin.Essentials
         public string Title { get; set; }
 
 #if !NETSTANDARD1_0
-        public Rectangle SourceScreenPosition { get; set; } = Rectangle.Empty;
+        public Rectangle PresentationSourceBounds { get; set; } = Rectangle.Empty;
 #endif
     }
 
--- a/Xamarin.Essentials/Share/Share.shared.cs
+++ b/Xamarin.Essentials/Share/Share.shared.cs
@@ -24,7 +24,7 @@ namespace Xamarin.Essentials
         }
     }
 
-    public class ShareRequestBase
+    public abstract class ShareRequestBase
     {
         public string Title { get; set; }
 
--- a/Xamarin.Essentials/Share/Share.shared.cs
+++ b/Xamarin.Essentials/Share/Share.shared.cs
@@ -1,6 +1,9 @@
 ﻿using System;
 using System.IO;
 using System.Threading.Tasks;
+#if !NETSTANDARD1_0
+using System.Drawing;
+#endif
 
 namespace Xamarin.Essentials
 {
@@ -21,7 +24,16 @@ namespace Xamarin.Essentials
         }
     }
 
-    public class ShareTextRequest
+    public class ShareBase
+    {
+        public string Title { get; set; }
+
+#if !NETSTANDARD1_0
+        public Rectangle SourceScreenPosition { get; set; } = Rectangle.Empty;
+#endif
+    }
+
+    public class ShareTextRequest : ShareBase
     {
         public ShareTextRequest()
         {
@@ -32,8 +44,6 @@ namespace Xamarin.Essentials
         public ShareTextRequest(string text, string title)
             : this(text) => Title = title;
 
-        public string Title { get; set; }
-
         public string Subject { get; set; }
 
         public string Text { get; set; }
@@ -41,7 +51,7 @@ namespace Xamarin.Essentials
         public string Uri { get; set; }
     }
 
-    public class ShareFileRequest
+    public class ShareFileRequest : ShareBase
     {
         public ShareFileRequest()
         {
@@ -68,8 +78,6 @@ namespace Xamarin.Essentials
         {
             File = new ShareFile(file);
         }
-
-        public string Title { get; set; }
 
         public ShareFile File { get; set; }
     }
--- a/Xamarin.Essentials/Share/Share.shared.cs
+++ b/Xamarin.Essentials/Share/Share.shared.cs
@@ -24,7 +24,7 @@ namespace Xamarin.Essentials
         }
     }
 
-    public class ShareBase
+    public class ShareRequestBase
     {
         public string Title { get; set; }
 
@@ -33,7 +33,7 @@ namespace Xamarin.Essentials
 #endif
     }
 
-    public class ShareTextRequest : ShareBase
+    public class ShareTextRequest : ShareRequestBase
     {
         public ShareTextRequest()
         {
@@ -51,7 +51,7 @@ namespace Xamarin.Essentials
         public string Uri { get; set; }
     }
 
-    public class ShareFileRequest : ShareBase
+    public class ShareFileRequest : ShareRequestBase
     {
         public ShareFileRequest()
         {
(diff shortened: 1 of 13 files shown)


Xamarin.Essentials/Share/Share.shared.cs
```diff
@@ -21,7 +24,16 @@ public static Task RequestAsync(ShareFileRequest request)
         }
     }
 
-    public class ShareTextRequest
+    public class ShareRequestBase
+    {
+        public string Title { get; set; }
+
+#if !NETSTANDARD1_0
+        public Rectangle SourceScreenPosition { get; set; } = Rectangle.Empty;
```
mattleibow: I think "Position" implies "point". Maybe a Bounds or Frame. Or even Rectangle, but not too much a fan of this. I like Bounds the most.

Redth: `PresentationSourceBounds`

Redth:
We were going to add sensible presets like TopLeft, TopRight, etc. however calculating this on each platform is a bit out of scope of this work for now.  It should have some more thought and discussion around what that means for each platform and what those coordinates look like for things like the status bars, etc.  For now I think it's best we enable this functionality to unblock iOS iPad apps.

Xamarin.Essentials/Share/Share.shared.cs
```diff
@@ -21,7 +24,16 @@ public static Task RequestAsync(ShareFileRequest request)
         }
     }
 
-    public class ShareTextRequest
+    public class ShareRequestBase
+    {
+        public string Title { get; set; }
+
+#if !NETSTANDARD1_0
+        public Rectangle PresentationSourceBounds { get; set; } = Rectangle.Empty;
```
mattleibow: LOL for netstandard1.x

bobwhitten:
Thanks @Redth. I've tested this and it works.  But I am concerned that some will be disappointed to find that if they don't set this SourceScreenPosition they will NOT get this to work on the iPad with 13.2.  For example, in the current sample code. 

Would an additional comment in the documentation clarify?

Redth:
I suppose we could set a default for the case where it’s a Tablet idiom and iOS. Do you think `Rectangle(width /2, height /2, 0, 0)` is a sane default for this?

bobwhitten:
I think center of screen is as good as any, though someone has asked for top right.

Most important thing is that this sourceRect get set in the iPad case.

jamesmontemagno:
LGTM

hemersonvarela:
If I'm allowed to ask, when will be release that includes this PR?

Redth:
We are working on a 1.4.0. I can’t give you a date yet but trying to sort out a bunch of PRs first to issue a release.

Ruddy2007:
@Redth In the future would it be possible to release hot fixes without including them in a bigger release? #980 was a super simple fix and by the looks of it would have helped a lot of people with a quick release (myself included).

As we do not have a release date for 1.4.0 it is forcing some of us to fork this project to make the changes ourselfs (as I will have to do now). Would be great if in the future we could treat these types of scenarios for what they really are... a hot fix with a quick release.

Redth:
Yes, we can try and do smaller fix releases in the future.  I would also remind everyone that you can go grab a .nupkg to use yourself: https://dev.azure.com/xamarin/public/_build?definitionId=7&_a=summary

We will look at creating some sort of nightly package feed for master builds which would help even more for these cases.

mauro3005:
Is this fix currently in the 1.4.0 pre-release? I installed the 1.4.0 pre-release nuget and Share sheet is still not showing on iPad iOS 13.3. I just see a little white dot in the top left corner when I tap share from the nav bar and if I tap again white dot goes away.

jamesmontemagno:
You can set the
```
PresentationSourceBounds = Device.RuntimePlatform == Device.iOS && Device.Idiom == TargetIdiom.Tablet
                            ? new System.Drawing.Rectangle(0, 20, 0, 0)
                            : System.Drawing.Rectangle.Empty
```

mauro3005:
> You can set the
> 
> ```
> PresentationSourceBounds = DeviceInfo.Platform== DevicePlatform.iOS && DeviceInfo.Idiom == DeviceIdiom.Tablet
>                             ? new System.Drawing.Rectangle(0, 20, 0, 0)
>                             : System.Drawing.Rectangle.Empty
> ```

Thank you James! This does indeed work with that PresentationSourceBounds Property set on 1.4.0 pre-release. We will be on the lookout for this on latest stable :)

MalukuSeito:
I tried this, however, now I can't run it on Android, I get the following exception:
`System.TypeLoadException: Could not set up parent class, due to: Could not load type of field 'Xamarin.Essentials.ShareRequestBase:<PresentationSourceBounds>k__BackingField' (1) due to: Could not resolve type with token 0100008d from typeref (expected class 'System.Drawing.Rectangle' in assembly 'System.Drawing.Common, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51') assembly:System.Drawing.Common, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 type:System.Drawing.Rectangle member:(null) `

I updated all projects to the same version and added System.Drawing.Common to the imported packages, but it's no help.
Any idea what might be missing? It worked fine on android before pre-2.

StevoKeano:
No popup....  Ran in 1/3 screen and full screen on the Model: iPad6,11 iPad Generation 5 - A1822.   Cannot believe I'm debugging over wifi from windows; but, this single call is broken..   Got here in 12 hours from Dev Account creation to almost there...

`        private static async Task ShareFile(string nNumber, string fltLogCSVOutputPath)

        {
            await Share.RequestAsync(new ShareFileRequest
            {
                Title = nNumber + " ShareFile",
                File = new ShareFile(fltLogCSVOutputPath),
                // PresentationSourceBounds = new System.Drawing.Rectangle(0, 0, 600, 300)
                PresentationSourceBounds = DeviceInfo.Platform == DevicePlatform.iOS && DeviceInfo.Idiom == DeviceIdiom.Tablet
                            ? new System.Drawing.Rectangle(0, 20, 0, 0)
                            : System.Drawing.Rectangle.Empty
            }); ;
        }`

mattleibow:
@StevoKeano  is it possible to open a new issue so we can track this properly?
Also, what happens when you run the share sample in the sample app in this repo?

Could you maybe attach a sample? I just did a test and it shows up for me. Maybe you are not doing it on the UI thread?

StevoKeano:
Using the UI Thread I wasn't.   Thanks @mattleibow !

mattleibow:
iOS is stricter. Anything that touches the UI at all must be on the UI thread.

StevoKeano:
Used TaskCompletionSource to get the (a)wait  effect... 
`        private Task<bool> ShareFile(string nNumber, string fltLogCSVOutputPath)

        {
            var tcs = new TaskCompletionSource<bool>();
            Device.BeginInvokeOnMainThread(() =>
            {
              Share.RequestAsync(new ShareFileRequest
                    {
                        Title = nNumber + " ShareFile",
                        File = new ShareFile(fltLogCSVOutputPath),
                        // PresentationSourceBounds = new System.Drawing.Rectangle(0, 0, 600, 300)
                        PresentationSourceBounds = DeviceInfo.Platform == DevicePlatform.iOS && DeviceInfo.Idiom == DeviceIdiom.Tablet
                            ? new System.Drawing.Rectangle(0, 20, 0, 0)
                            : System.Drawing.Rectangle.Empty
                    });
            });
            return tcs.Task;
        }`

Issue #996, opened 2019-11-21
[Spec]  Calendar Read-Only API


nickrandolph:
# Calendar Read-Only API

Calendar API for reading calendar/event information seamlessly across uwp, iOS and Android

# API

### Methods

| API | Description |
| ------------- | ------------- |
| ``Task<IEnumerable<DeviceCalendar>> GetCalendarsAsync()`` | Retrieve all calendars for device |
|  ``Task<IEnumerable<DeviceEvent>> GetEventsAsync( string calendarId = null, DateTimeOffset? startDate = null, DateTimeOffset? endDate = null)`` | Retrieve all events for a specified range/calendar, null will use defaults: calendarId will default to all Calendars retrieved, startDate will default to use, DefaultStartDate , endDate will default to use DefaultEndDate |
| ``Task<DeviceEvent> GetEventByIdAsync(string eventId)`` | Gets more detailed information about a specific event by the event id |

### Classes
```
public class DeviceCalendar
{
    string Id { get; set; }

    string Name { get; set; }
}

public class DeviceEvent
{
    string Id { get; set; }

    string CalendarId { get; set; }

    string Title { get; set; }

    string Description { get; set; }

    string Location { get; set; }

    bool AllDay { get; set; }

    DateTimeOffset? Start { get; set; }

    DateTimeOffset? End { get; set; }

    TimeSpan? Duration { get; set; }

    IReadOnlyList<DeviceEventAttendee> Attendees { get; set; }
}

public class DeviceEventAttendee
{
    string Name { get; set; }

    string Email { get; set; }
}
```
# Scenarios

- User wants to see upcoming events.
- User wants to see who is attending an upcoming event.
- User wants to know when they will be free.

# Platform Compatibility
- Target Frameworks: 
  - iOS:  Support on iOS for the API
  - Android: Support on Android for the API
  - UWP: Support on UWP for the API
# Backward Compatibility
- UWP: Windows 10 (introduced v10.0.10240.0) 
(https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.appointments.appointment)
- Android: android:minSdkVersion="14" 
(https://developer.android.com/reference/android/provider/CalendarContract)
- iOS: All versions until reminders are added, then iOS 6
(https://developer.apple.com/documentation/eventkit/ekeventstore)
# Difficulty : low

jamesmontemagno:
[...]

PR #1009, opened 2019-12-04
Calendar Read-Only API

nickrandolph:
# Calendar Read-Only API

Calendar API for reading calendar/event information seamlessly across uwp, iOS and Android
(Issue https://github.com/xamarin/Essentials/issues/996)
# API

### Methods

| API | Description |
| ------------- | ------------- |
| ``Task<IEnumerable<DeviceCalendar>> GetCalendarsAsync()`` | Retrieve all calendars for device |
|  ``Task<IEnumerable<DeviceEvent>> GetEventsAsync( string calendarId = null, DateTimeOffset? startDate = null, DateTimeOffset? endDate = null)`` | Retrieve all events for a specified range/calendar. Null values for parameters will result in default values being used: null calendarId will return events from all calendares, startDate will default to the Now, endDate will default to startDate+14 days |
| ``Task<DeviceEvent> GetEventByIdAsync(string eventId)`` | Gets more detailed information about a specific event by the event id |

### Classes
```
public class DeviceCalendar
{
    string Id { get; set; }

    string Name { get; set; }
}

public class DeviceEvent
{
    string Id { get; set; }

    string CalendarId { get; set; }

    string Title { get; set; }

    string Description { get; set; }

    string Location { get; set; }

    bool AllDay { get; set; }

    DateTimeOffset Start { get; set; }

    DateTimeOffset? End { get; set; }

    TimeSpan? Duration { get; set; }

    IEnumerable<DeviceEventAttendee> Attendees { get; set; }
}

public class DeviceEventAttendee
{
    string Name { get; set; }

    string Email { get; set; }
}
```
# Scenarios

- User wants to see upcoming events.
- User wants to see who is attending an upcoming event.
- User wants to know when they will be free.

# Platform Compatibility
- Target Frameworks: 
  - iOS:  Support on iOS for the API
  - Android: Support on Android for the API
  - UWP: Support on UWP for the API
# Backward Compatibility
- UWP: Windows 10 (introduced v10.0.10240.0) 
(https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.appointments.appointment)
- Android: android:minSdkVersion="14" 
(https://developer.android.com/reference/android/provider/CalendarContract)
- iOS: All versions until reminders are added, then iOS 6
(https://developer.apple.com/documentation/eventkit/ekeventstore)

# Contributors
@ScottBTR
@mkieres

### DIFF
--- a/DeviceTests/DeviceTests.Android/Properties/AndroidManifest.xml
+++ b/DeviceTests/DeviceTests.Android/Properties/AndroidManifest.xml
@@ -5,10 +5,11 @@
 	<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 	<uses-permission android:name="android.permission.BATTERY_STATS" />
+  <uses-permission android:name="android.permission.READ_CALENDAR" />
 	<uses-permission android:name="android.permission.CAMERA" />
-	<uses-permission android:name="android.permission.FLASHLIGHT" />
+  <uses-permission android:name="android.permission.FLASHLIGHT" />
 	<uses-permission android:name="android.permission.INTERNET" />
-	<uses-permission android:name="android.permission.VIBRATE" />
-	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+  <uses-permission android:name="android.permission.VIBRATE" />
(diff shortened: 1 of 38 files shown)


nickrandolph:
@jamesmontemagno  build is failing, I think on the tests. Issue with the calendar is that the tests require the user to have a calendar. Any ideas on a work around?

nickrandolph:
Aborted this PR until we can solve the failing test

PR #1028, opened 2019-12-17
Fixes #1027 [Andriod] Don't use the configuration to detect the UIMode

### LINKED ISSUE #1027: [Bug] DeviceInfo.Idiom return TV instead of Phone while Xamarin.Forms.Device.Idiom return idiom correct
### Description
Returns inconsistent values from Forms and Essentials.

### Steps to Reproduce
1. Compare returns values from Xamarin.Essentials.DeviceInfo.Idiom and Xamarin.Forms.Device.Idiom on Xiaomi Mi9 SE

### Expected Behavior
Xamarin.Forms.Device.Idiom and Xamarin.Essentials.DeviceInfo.Idiom behave in same way.

### Actual Behavior
Xamarin.Forms.Device.Idiom returns TargetIdiom.Phone
Xamarin.Essentials.DeviceInfo.Idiom returns DeviceIdiom.TV

### Basic Information

- Version with issue: 1.3.1
- Last known good version: null
- IDE: Visual Studio 2019 16.4
- Platform Target Frameworks:
  - Android: 9.0, stable MIUI Global 11.0.4.0
- Affected Devices: Xiaomi Mi9 SE

### Screenshots
I have modified Sample Xamarin.Essentials project to compare Essentials to Forms.
Screenshot_2019-12-17-19-18-09-718_com xamarin essentials

### Additional Information
My device return CurrentModeType = 12 which means CurrentModeType.HasFlag(UiMode.TypeTelevision); will returns true.

jamesmontemagno:
We were using the configuration, which has multiple items inside of it. Some tablets/phones may have the TV mode set, which is weird :(  This matches Xamarin.Forms funcationality and adds an extra check just incase we don't have the configuration for some reason.

For reference:
 https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.Android/Forms.cs#L325

### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #1027

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###


jamesmontemagno:
@PawKanarek I think that when you use UIModeManager it only returns 1 of them in which the check for == or HasFlag doesn't matter because it returns the single type. 

The key difference is that we are using the UIModeManager => CurrentModeType which is just 1 type and would return Normal first. The issue arises when using AppContext.Resources?.Configuration because it contains all sorts of variances. 

Can you re-run your tests that you did in the issue with using: 

```
 using var uiModeManager = UiModeManager.FromContext(Essentials.Platform.AppContext);
 var uiMode = uiModeManager?.CurrentModeType ?? UiMode.TypeUndefined;
```

PawKanarek:
@jamesmontemagno During my tests i was using 2 approaches: 
```
 var uiModeManager = UiModeManager.FromContext(Essentials.Platform.AppContext);
 var uiMode = uiModeManager?.CurrentModeType ?? UiMode.TypeUndefined;
``` 
and 
```
  var manager = (UiModeManager)Application.Context.GetSystemService(Context.UiModeService);
  var currentType = manager.CurrentModeType
```
And both of them were returing 1(TypeNormal), and as far as i see from post @kcrg  
https://github.com/xamarin/Essentials/issues/1027 he was using second approach and that returned 12. My worries are not about place from we get **CurrentModeType** but about method that we determine whether CurrentModeType is TV, Watch or anything. 

When im using Application.Context.Resources.Configuration.UiMode i got this: 

image

PawKanarek:
Can you checkout this comment: https://github.com/xamarin/Essentials/issues/1027#issuecomment-566970050 :)

kcrg:
@jamesmontemagno I checked this pull request and it works now. Thanks for quick response and fix :)

Screenshot_2019-12-19-19-37-38-426_com xamarin essentials

MagicAndre1981:
@jamesmontemagno can you fix the code for Xamarin.Forms, too? I [used](https://github.com/xamarin/Xamarin.Forms/pull/7933) your old code to "fix" TV detection it for XF.

jamesmontemagno:
I believe the Xamarin.Forms functionality is correct already.

MagicAndre1981:
What have you fixed here with the PR? @jamesmontemagno I used your old code with HasFlags

jamesmontemagno:
So, I actually fixed 2 issues in my code to match first how we compare for tablets modes, but yes, I will ask @hartez and @kingces95 as in this PR: https://github.com/xamarin/Xamarin.Forms/pull/7933 you did HasFlags, however this is incorrect based on https://github.com/xamarin/Essentials/issues/1027 as on some android devices they will return TV even though they aren't TV and a direct comparison is correct.

MagicAndre1981:
> So, I actually fixed 2 issues in my code to match first how we compare for tablets modes, but yes, I will ask @hartez and @kingces95 as in this PR: [xamarin/Xamarin.Forms#7933](https://github.com/xamarin/Xamarin.Forms/pull/7933) you did HasFlags, however this is incorrect based on #1027 as on some android devices they will return TV even though they aren't TV and a direct comparison is correct.

have you asked them? Forms will not use [dependency on Essentials anymore](https://github.com/xamarin/Xamarin.Forms/pull/9926#issuecomment-612492826) this has to be synced.

Issue #1029, opened 2019-12-18
[Enhancement] Some struct equality comparisons still allocate. 


Mrnikbobjeff:
### Description
Some struct equality operations still allocate after the latest 'fix', caused by using the generic enum Equals method. This happens in the Equals method, which is called by the == operator as well.
### Steps to Reproduce
1.  Compare two DisplayInfo structs either via == or Equals
2. Other enum value is boxed

### Expected Behavior
Just use the equality operator instead of the equals function for enums
### Actual Behavior
object.equals is called for enums
### Basic Information

- Version with issue: All
- Last known good version: None
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Screenshots
[proof](https://imgur.com/a/lNqszxt) that it applies generally, see DisplayInfo as an example

jamesmontemagno:
Iss this a duplicate of https://github.com/xamarin/Essentials/issues/965 at all?

Mrnikbobjeff:
No, this issue is solely for the DisplayInfo struct, as it is the only struct containing an enum being compared with object.equals.

Mrnikbobjeff:
Closed, James apparently fixed this somewhen for the only struct where it occurred

mattleibow:
Very nice of him.

Issue #965, opened 2019-10-30
[Enhancement] Redundant allocation in TextToSpeech.Android


Mrnikbobjeff:
### Description
The c# spec allows anonymous functions delegates to be cached, method group conversions may not be cached though and have to be allocated delegate instances every time. We have several instances in Xamarin.Essentials where we use method group conversions for brevity allowing them to allocate unnecessarily.

mattleibow:
Duplicate of https://github.com/xamarin/Essentials/issues/1029

PR #1030, opened 2019-12-18
Removed Method Group conversions, more additions to equality.

Mrnikbobjeff:
### Description of Change ###
Removed most method group conversions to be able to cache the delegate as this is currently forbidden by the standard for method group conversions. [Here](https://github.com/Mrnikbobjeff/BenchmarkingSolution/blob/master/BenchmarkingSolution/EnumAllocations.cs) is the enum case benchmark as linked in the first issue, [here](https://github.com/Mrnikbobjeff/BenchmarkingSolution/blob/master/BenchmarkingSolution/MethodGroupConversions.cs) we see the effect of method group conversions.

- Related to issue 
#1029, #965
### API Changes ###

None

### Behavioral Changes ###

They should simply see less allocations throughout the app. 

### PR Checklist ###


jamesmontemagno:
When it comes to this will have @terrajobst validate.

terrajobst:
> When it comes to this will have @terrajobst validate.

@jaredpar is more appropriate to validate this change

Mrnikbobjeff:
Just for further reference, the current c# spec says the following about method group conversions in section 6.6 :"A new instance of the delegate type D is allocated. If there is not enough memory available to allocate the new instance, a System.OutOfMemoryException is thrown and no further steps are executed."
Concerning anonymous, the C# spec states the following in section 6.5.1: "Conversions of semantically identical anonymous functions with the same (possibly empty) set of captured outer variable instances to the same delegate types are permitted (but not required) to return the same delegate instance."
There was some work going on to allow method group conversions to be cached, but this does not appear to be on the roadmap to happen in the near future.

mattleibow:
Interesting... I always thought these anonymous delegates would be worse. Seems C# tries to catch me out even more ;)

Mrnikbobjeff:
@jaredpar Did you have a change to look over the changes yet?

jaredpar:
@Mrnikbobjeff the method group conversions to delegates look good

AmrAlSayed0:
https://github.com/dotnet/roslyn/issues/5835 I think it is related to this issue, though this is the roslyn compiler, I don't know if the same issue is in the mono compiler.

Mrnikbobjeff:
That was the issue I referenced in the ticket. No matter what mono does, as long as it abides by the standard it will be exactly the same. Feel free to try yourself with the linked benchmark, you just have to add a ClrJob for mono

Issue #315, opened 2018-06-22
UWP Idiom must be called on UI Thread


jrahma:
Hi,

I have below class but when I call the following from my **Signup.xaml.cs**:

`await eventLog.UpdateEventLog(Convert.ToString(data[0].id), "SNUP", "test", "Successfully Signed up.");
`
it only run until:

`Console.WriteLine("event log starting");
`
and then stops... but the code works fine if I run it from the **Signup.xaml.cs** directly.

here is the eventLog code:

```
using System;
using System.Collections.Generic;
using Xamarin.Forms;
using Microsoft.AppCenter;
using Microsoft.AppCenter.Analytics;
using Microsoft.AppCenter.Crashes;
using System.Collections.ObjectModel;
using System.Net.Http;
using System.Net.Http.Headers;
using Newtonsoft.Json;
using Newtonsoft.Json.Linq;
using System.Text.RegularExpressions;
using System.Net;
using System.Text;
using Com.OneSignal;
using Syncfusion.SfPicker.XForms;
using Xamarin.Essentials;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;

namespace XoomPage
{
    public class EventLog
    {
        public async Task<HttpResponseMessage> UpdateEventLog(string user_id, string category, string reference, string text)
        {
            string country;
            double latitude;
            double longitude;

            // Console.Clear();

            Console.WriteLine("event log starting");

            var location = await Geolocation.GetLocationAsync();

            Console.WriteLine("event log location is ok");

            latitude = location.Latitude;
            longitude = location.Longitude;

            Console.WriteLine("latitude: " + Convert.ToString(latitude));
            Console.WriteLine("longitude: " + Convert.ToString(longitude));

            var placemarks = await Geocoding.GetPlacemarksAsync(latitude, longitude);
            var placemark = placemarks?.FirstOrDefault();

            country = placemark.CountryCode;

            string model = DeviceInfo.Model;
            string manufacturer = DeviceInfo.Manufacturer;
            string device = DeviceInfo.Name;
            string version = DeviceInfo.VersionString;
            string platform = DeviceInfo.Platform;
            string idiom = DeviceInfo.Idiom;

            using (var client = new HttpClient())
            {
                client.BaseAddress = new Uri("http://www.domain.com/ws/event_log.php");

                var content = new FormUrlEncodedContent(new[]
                {
                    new KeyValuePair<string, string>("user", user_id)
                    /*
[...]

PR #1036, opened 2019-12-20
GH-315 Used a cached approach for getting the idiom when not on UI for UWP

jamesmontemagno:
### Description of Change ###

Cache first and then capture exceptions when calling not from UI

### Bugs Fixed ###

- Related to issue #315

### Behavioral Changes ###

Users much call DeviceInfo.Idiom atleast once on the UI thread.

### PR Checklist ###


mattleibow:
Can we not use the main Dispatcher or other RunOnMainThread logic?

jamesmontemagno:
@mattleibow so, the issue here is that it is just a property. Unless we called to invoke on the main thread and waited :(.

What Xamarin.Forms does is just set it on the Init() and never updates it again. This value could possibly change. I think best that in the future or a 2.0 we introduce an Init() call for iOS/UWP and even if we don't use it have it at our disposal.

PR #1038, opened 2019-12-21
Add color converter samples :)

jamesmontemagno:
Validates that the linker is working alright with color converters. Nice samples :)

image

### DIFF
--- a/Samples/Samples/Samples.csproj
+++ b/Samples/Samples/Samples.csproj
@@ -24,4 +24,10 @@
     <ProjectReference Include="..\..\Xamarin.Essentials\Xamarin.Essentials.csproj" />
   </ItemGroup>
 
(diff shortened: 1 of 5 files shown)


Samples/Samples/Samples.csproj
```diff
@@ -24,4 +24,10 @@
     <ProjectReference Include="..\..\Xamarin.Essentials\Xamarin.Essentials.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Update="View\ColorConvertersPage.xaml.cs">
```
mattleibow: I see VS is up to its old tricks again :)

jamesmontemagno: Good catch! Fixed :)

Issue #299, opened 2018-06-20
Device Contacts


Quentinb:
Hi Team, thanks for the great library, will save a huge amount of time.

Is it possible to add access to the contacts list into the API? Perhaps it can return a List<contacts> or something similar?

> VS bug [#735685](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735685)

jamesmontemagno:
This is a pretty loaded API and extremely complicated. What I think most developers actually may want is a "Contact Picker" so you launch and pick a contact.

Can you give more use case on how you (and other developers please chime in) how you would use this API.

Quentinb:
@jamesmontemagno thanks for the feedback. I my use case I need access to the list of contacts in code to populate a custom UI element as well format a message structure. For the most part I only need to read the list and obtain mobile number and/or email when working with the data itself and the UI element would be a picker type control, so I would be building the datasource for the picker.

The current app is Xamarin.IOS and not forms. #ContraintLayout

Redth:
I think we could consider a really slimmed down API for this.

The first step I think is to figure out how each platform deals with this, and what is the minimum amount of information we'd want to provide for a 'contact' that would be returned from this API?  

The first pass would be a read only API to get a list of contacts and some basic fields for them.  I do seem to remember Android being particularly bad between API levels and vendors in terms of different field names were used to store the same information.

Before we go ahead with this I'd like to see some prototyping done on Android to see how feasible this really is.

markbrents:
I would love to see this feature added. Even a read only list of contacts would be helpful.

JakeStaTeresa:
I definitely see myself using this once available. Here are my thoughts on this from experience working on various cross platform frameworks. 

This API needs to support streaming/paging result sets.The device API's can take long to return 1000's of contacts. Also, the fetch operation needs to happen on a background thread.

pictos:
What info do we want to return? Just phone numbers and name or email too?
In iOS and UWP we can get all info about contacts very easily, but in Android is a little a more complicated.

Redth:
I think we want to return whatever the biggest subset of useful data is that's easy to figure out on all the platforms....

- Email(s)
- Phone(s)
- Address(es)
- Name
- Birthday

Are those all easily identifiable and easy enough to capture on all platforms (looking at you android)?  Again I seem to remember Android having different field names/mappings for the same data across vendors and api levels.

Redth:
[...]

PR #1040, opened 2019-12-21
Update for PR [GH-299] - Pick a contact

pictos:
Updated PR #752.


Samples/Samples/ViewModel/ContactViewModel.cs
```diff
@@ -0,0 +1,134 @@
+﻿using System;
+using System.Windows.Input;
+using Xamarin.Essentials;
+using Xamarin.Forms;
+using Xamarin.Forms.Internals;
```
jamesmontemagno: Shouldn't need this

Xamarin.Essentials/Contacts/Contact.android.cs
```diff
@@ -0,0 +1,161 @@
...
+                }
+                cursor.Close();
+
+                projection = new string[]
+                {
+                        ContactsContract.CommonDataKinds.Email.Address,
+                        ContactsContract.CommonDataKinds.Email.InterfaceConsts.Type
+                };
+
+                cursor = context.Query(ContactsContract.CommonDataKinds.Email.ContentUri, null, ContactsContract.CommonDataKinds.Email.InterfaceConsts.ContactId + " = ?", idQ, null);
```
jamesmontemagno: should there be a space in = ?

Xamarin.Essentials/Contacts/Contact.ios.cs
```diff
@@ -0,0 +1,137 @@
...
+using Foundation;
+using UIKit;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Contact
+    {
+        internal static Action<PhoneContact> CallBack { get; set; }
+
+        internal static UIViewController UIView => Platform.GetCurrentViewController();
```
jamesmontemagno: I would say not needed and just call it below in each method, could also check against null.

Samples/Samples/ViewModel/ContactViewModel.cs
pictos: It's just to use the `ForEach` extension method. If you prefer I can use the normal one.

jamesmontemagno: Yeah, would just use the normal one

Xamarin.Essentials/Contacts/Contact.ios.cs
pictos: done.

Samples/Samples/ViewModel/ContactViewModel.cs
pictos: done.

Xamarin.Essentials/Contacts/Contact.android.cs
pictos: I did some local tests, and just the `AND` needs the space before and after.

pictos:
@jamesmontemagno if you need more changes, please let me know.

Xamarin.Essentials/Contacts/Contact.ios.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (contact == null)
+                return default;
+
+            try
+            {
+                var contactType = ToPhoneContact(contact.ContactType);
+                var phones = new Dictionary<string, ContactType>();
+
+                foreach (var item in contact.PhoneNumbers)
+                    phones.Add(item?.Value?.StringValue, contactType);
```
follesoe: The same contact type is used for all numbers and is based on the mapping from `CNContactType`. After testing this code on iOS, the behaviour of the API is to return `CNContactType.Person` for any contact that has a given name and family set. It returns `CNContactType.Organization` only for contacts where these values are empty, but the company name is set.

Instead, I think the correct approach would be to add a new mapping function that maps the `item.Label` to `ContactType`, based on constants defined in `CNLabelPhoneNumberKey`, such as `CNLabelPhoneNumberKey.Mobile` or `CNLabelPhoneNumberKey.Main`, or raw values such as `_$!<Work>!$_` (which is the label for work number, but which is not assigned a const in the API). This would be more aligned with the UWP `GetPhoneContactType` method.

Perhaps a new enum of type `PhoneType` could be considered to distinguish between mobile and other? But I totally see the point of keeping the API small and simple.

Xamarin.Essentials/Contacts/Contact.ios.cs
```diff
@@ -0,0 +1,143 @@
...
+
+                var emails = new Dictionary<string, ContactType>();
+
+                foreach (var item in contact.EmailAddresses)
+                    emails.Add(item?.Value?.ToString(), contactType);
```
follesoe: Same case as for numbers. I think a better approach would be to add a mapping function similar to the UWP implementation and use keys such as `CNLabelKey.Work` and `CNLabelKey.Home` to determine the e-mail address type. The iOS API has some constants defined, but users can also assign labels at will, and the Contacts app even suggests pre-defined labels that has no corresponding constant, so covering at least the key labels suggested by the OS would be smart.

Xamarin.Essentials/Contacts/Contact.ios.cs
```diff
@@ -0,0 +1,143 @@
...
+
+                var name = $"{contact.GivenName} {contact.MiddleName} {contact.FamilyName}";
```
follesoe: Should check for empty string in `MiddleName`, to prevent generating a full name consisting of double spaces between `GivenName` and `FamilyName`.

If the contact type is `CNContactType.Organization`, the `OrganizationName` property should be used as the contact name to prevent returning a contact without a name.

pictos:
@follesoe thank you for this awesome feedback❣️I will look into your comments as soon as possible and fix those issues and probably I will ping you for more info 😉.

follesoe:
@pictos Your welcome 👍. As it will take some time for this to make it into Xamarin.Essentials, I will continue with my local implementation based on this PR, and provide you with feedback if I discover anything else. I will also share the mappings I end up using between label types and Work/Personal e-mail and phone numbers.

Xamarin.Essentials/Contacts/Contact.shared.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                await Permissions.RequestAsync<Permissions.ContactsRead>();
+                await Permissions.RequestAsync<Permissions.ContactsWrite>();
+            }
+        }
+    }
+
+    public enum ContactType
+    {
+        Unknow = 0,
```
follesoe: A small typo. Should be `Unknown`.

Xamarin.Essentials/Contacts/ContactActivity.android.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            if (resultCode == Result.Canceled)
+                Finish();
+
+            try
+            {
+                if (requestCode != 101)
+                    return;
+
+                Contact.CallBack(Contact.PlatformGetContacts(data.Data));
```
follesoe: If you cancel contact selection by pressing back, `data` is null. Should return `null` if `data` is null.

Xamarin.Essentials/Contacts/ContactActivity.android.cs
```diff
@@ -0,0 +1,53 @@
...
+            intent.SetType(ContactsContract.CommonDataKinds.Phone.ContentType);
+            StartActivityForResult(intent, 101);
+        }
+
+        protected override void OnActivityResult(int requestCode, [GeneratedEnum] Result resultCode, Content.Intent data)
+        {
+            base.OnActivityResult(requestCode, resultCode, data);
+
+            if (resultCode == Result.Canceled)
+                Finish();
```
follesoe: The `Finish` method returns `void`, so the code execution will continue, reulsting in a null reference exception at line 41. Also, simply returning early will cause the `TaskCompletionSource` in `Contact.android.cs` to never reach a completed state.

Xamarin.Essentials/Contacts/ContactActivity.android.cs
```diff
@@ -0,0 +1,53 @@
...
+            try
+            {
+                if (requestCode != 101)
+                    return;
+
+                Contact.CallBack(Contact.PlatformGetContacts(data.Data));
+            }
+            catch (Exception ex)
+            {
+                throw ex;
```
follesoe: Consider adding a new `Contact.ErrorCallback` which can set `tcs.TrySetException(exception);`?

follesoe:
Adding some general inputs on the proposed API instead of doing it as line comments in the different classes.

An issue with both the iOS and Android implementation is that it does not deal with cancelation. The Contact `Activity` and `ViewController` can both be canceled without selecting a contact. In both cases, the Task returned from `Task<PhoneContact> PlatformPickContactAsync()` never completes.

I think we need a cancel callback so that the `ContactActivity` and `ContactDelegate` can signal that selection was canceled. 

In the canceled callback we can either use `tcs.TrySetCanceled`, which I think will return in a TaskCancelledException in the calling method awaiting the result. Alternatively, the `PlatformPickContactAsync` method could return `null`. But, since `PhoneContact` is a `struct` returning `null` isn't an option. I am not sure if there are similar APIs in Xamarin.Essentials that we can look at to deal with methods that may return a value.

pictos:
@follesoe, when you have time, could you take a look at the Android implementation? I made some adjustments to avoid `NullReferenceException`.

Instead of a class, I use a `NullableReferenceType`. What do you think about this?

follesoe:
@pictos That would solve the null issue, and be a way to communicate that no contact was selected. So that sounds like a good approach 👍.

One suggestion for the Android implementation on mapping phone and e-mail types to `ContactType` enum is to use the `PhoneDataKind` and `EmailDataKind` enums.

I created two helper methods to do the mapping:

```csharp
static ContactType GetPhoneContactType(string type)
{
	if (int.TryParse(type, out int typeInt))
	{
		try
		{
			var phoneKind = (PhoneDataKind)typeInt;
			return phoneKind switch
			{
				PhoneDataKind.Home => ContactType.Personal,
				PhoneDataKind.Mobile => ContactType.Personal,
				PhoneDataKind.Main => ContactType.Personal,
				PhoneDataKind.Work => ContactType.Work,
				PhoneDataKind.WorkMobile => ContactType.Work,
				PhoneDataKind.CompanyMain => ContactType.Work,
				PhoneDataKind.WorkPager => ContactType.Work,
				_ => ContactType.Unknown
			};
		}
		catch(Exception)
		{
			return ContactType.Unknown;
		}
	}
	return ContactType.Unknown;
}

static ContactType GetEmailContactType(string type)
{
	if (int.TryParse(type, out int typeInt))
	{
		try
		{
			var emailKind = (EmailDataKind)typeInt;
			return emailKind switch
			{
				EmailDataKind.Home => ContactType.Personal,
				EmailDataKind.Work => ContactType.Work,
				_ => ContactType.Unknown
			};
		}
		catch (Exception)
		{
			return ContactType.Unknown;
		}
	}
	return ContactType.Unknown;
}
```

A bit more verbose than the current implementation, but don't depend on magic strings.

For iOS I have created two arrays to keep the work and personal keys, and check if the label exists in either of them before returning Unknown.

```csharp
private static readonly string[] WorkKeys =
{
	CNLabelKey.Work,
};

private static readonly string[] PersonalKeys =
{
	CNLabelKey.Home,
	CNLabelPhoneNumberKey.iPhone,
	CNLabelPhoneNumberKey.Main,
	CNLabelPhoneNumberKey.Mobile
};
```

I have not checked all the label values for the types of labels that don't have an iOS API defined, so I will add more keys to each of the arrays after doing more testing on iOS.

Xamarin.Essentials/Contacts/ContactActivity.android.cs
```diff
@@ -0,0 +1,53 @@
...
+using Android.OS;
+using Android.Provider;
+using Android.Runtime;
+using Content = Android.Content;
+
+namespace Xamarin.Essentials
+{
+    [Activity(ConfigurationChanges = Content.PM.ConfigChanges.Orientation | Content.PM.ConfigChanges.ScreenSize)]
+#pragma warning disable IDE0040 // Add accessibility modifiers
+    internal class ContactActivity : Activity
```
Redth, line 12: Please check out `IntermediateActivity` and see if you can use that instead of creating another activity.

Xamarin.Essentials/Types/PhoneContact.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Xamarin.Essentials
+{
+    public readonly struct PhoneContact : IEquatable<PhoneContact>
```
Redth, line 7: Please rename to `Contact`

Issue #981, opened 2019-11-12
[Bug] OrientationSensor: inconsistent behavior on iOS vs Android


janusw:
### Description

I am trying to use the OrientationSensor as of Xamarin Essentials version 1.3.1 on Android and iOS.

On Android I see essentially the expected behavior, as documented at https://docs.microsoft.com/en-us/xamarin/essentials/orientation-sensor, namely: Pointing the device in a particular direction gives me a particular quaternion, irrespective of the orientation in which the sensor was started. When stopping the sensor and restarting it without changing orientation, the quaternion I get is (roughly) the same as before the restart.

On iOS this is different. There the quaternions I obtain depend on the device orientation in which the sensor was started: When starting the sensor while the phone lies flat on the table, the initial quaternion is always (0,0,0,1), no matter in which direction the device is pointing. If I stop and restart the sensor, the quaternion always goes back to (0,0,0,1).

However, when the sensor is started while the device is out of the table plane, the initial quaternion is not (0,0,0,1). The inclination angle seems to be reflected in the initial quaternion, but the azimutal angle is not.

This means that the orientation quaternion obtained on iOS is somehow relative to the initial orientation, while on Android it is 'absolute' (in the sense that it describes the rotation of the phone coordinate system into the earth coordinate system, as explained in the documentation).

### Steps to Reproduce

1. Put the device flat on the table, with the screen facing upwards and the top of the device pointing west.
2. Start the orientation sensor (OrientationSensor.Start).
3. Observe the reported quaternion (via OrientationSensor.ReadingChanged).

### Expected Behavior

The observed quaternion should be (0, 0, 0.707, 0.707), or at least close to it.

### Actual Behavior

The observed quaternion is (0, 0, 0, 1). Seen on an iPhone 8 with iOS 13.2.

### Basic Information

- Version with issue: Essentials 1.3.1
- Last known good version: ?
- IDE: Visual Studio for Mac 8.3.7
- Platform Target Frameworks: 
  - iOS:  8.0
  - Android: 10
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices: iPhone 8

Lelelo1:
This could somehow have to do with [CMAttitudeReferenceFrame](https://developer.apple.com/documentation/coremotion/cmattitudereferenceframe?language=objc). I don't know what the default is. I have very limited experience with the Android sensors.

janusw:
[...]

PR #1050, opened 2020-01-08
bugfix for OrientationSensor on iOS

janusw:
### Description of Change ###

* this is a fix-up for commit 84579595f36a008e172824aa41269596d83d167c
* quaternions are not commutative, and the original commit had the multiplication in the wrong order
* it worked for some special cases, but not in general

### Bugs Fixed ###

- Related to issue #981

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/OrientationSensor/OrientationSensor.ios.watchos.cs
+++ b/Xamarin.Essentials/OrientationSensor/OrientationSensor.ios.watchos.cs
@@ -26,12 +26,16 @@ namespace Xamarin.Essentials
 
             var field = data.Attitude.Quaternion;
 
-            // the quaternion returned by the MotionManager refers to a frame where the X axis points north
+            // the quaternion returned by the MotionManager refers to a frame where the X axis points north ("iOS frame")
             var q = new Quaternion((float)field.x, (float)field.y, (float)field.z, (float)field.w);
 
-            // we rotate it by 90 degrees around the Z axis, so that the Y axis points north and the X axis points east
+            // in Xamarin, the reference frame is defined such that Y points north and Z is vertical,
+            // so we first rotate by 90 degrees around the Z axis, in order to get from the Xamarin frame to the iOS frame
             var qz90 = Quaternion.CreateFromAxisAngle(Vector3.UnitZ, (float)(Math.PI / 2.0));
-            q = Quaternion.Multiply(q, qz90);
+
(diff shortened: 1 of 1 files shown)


janusw:
@jamesmontemagno @Redth I hope that this fix can still be considered for the 1.4.0 release, otherwise the OrientationSensor will continue to be broken!

jamesmontemagno:
Sure can :) should we write some unit tests for this?

janusw:
> Sure can :)

Great, thanks a lot. I'm sorry that I missed this problem in the original commit. Obviously my testing was not sufficient (I mostly tested rotations around the z axis, which is exactly the special case where the ordering of the quaternions does not matter :see_no_evil:).

I only noticed the problem when I tested an app that I'm building with Xamarin.Essentials 1.4.0-pre.

Btw, is there a simple way to build an app with a custom (non-release/non-tagged) version of Essentials? I usually just have a PackageReference in my project files, so that nuget fetches the package, but that only works with released/tagged versions, right?

> should we write some unit tests for this?

Yes, might be a good idea. One should probably check for a few cases that `OrientationSensor.DataUpdated` produces the correct output, when given specific `CMDeviceMotion` data.

Unfortunately I don't have a lot of experience with unit testing in the Xamarin framework yet, so it would probably take me some time. If you could provide a small skeleton for such a test, that would be a great help. Then I could take care of inserting the proper input and output values ...

janusw:
> Btw, is there a simple way to build an app with a custom (non-release/non-tagged) version of Essentials? I usually just have a PackageReference in my project files, so that nuget fetches the package, but that only works with released/tagged versions, right?

Just found https://help.github.com/en/github/managing-packages-with-github-packages/configuring-dotnet-cli-for-use-with-github-packages

I guess I could use that mechanism for referencing a xamarin-essentials version from my forked repo on github ...?

jamesmontemagno:
So on every PR that is built successfully you can download the nuget package from azure devops! <3 We are also working on a CI feed, but that will be master branch only

janusw:
> So on every PR that is built successfully you can download the nuget package from azure devops!

Ah, great. Just found it (https://dev.azure.com/xamarin/public). Thanks for the hint, and for merging!

Just tested my app thorougly with the nuget package from Azure, and it works great, so I'm pretty confident that the OrientationSensor behaves correctly on iOS now. 👍

PR #1056, opened 2020-01-14
Dev/file picker fix android

EvoPulseGaming:
### Description of Change ###
Due to some code loss from a previous merge, I re-added android permissions, and fixed the implementation. Since "Data" is obsolete, I've just returned the URI as path since Android10 doesn't like paths anymore xD

### Behavioral Changes ###

Android will return the original URI for GetPath. This might be wrong, but it appears that is the direction android wants anyways.

### PR Checklist ###

I don't think I need to make a test for this or update documentation... I've tested this on an android phone (9.0)

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.android.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.android.cs
@@ -16,7 +16,7 @@ namespace Xamarin.Essentials
         {
             // we only need the permission when accessing the file, but it's more natural
             // to ask the user first, then show the picker.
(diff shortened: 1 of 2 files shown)


EvoPulseGaming:
So I've tried a few different things to get this get all the checks to pass.

What I noticed, is if I download the device-test branch, and swap out the Essentials source-code with the masters version, all kinds of errors pop up. I'm thinking the device-test branch hasn't been been looked at in awhile. I'm not sure. What I am sure of, is this builds and runs on my device just fine...

Plus my changes are so small i'm having a hard time figuring out whats wrong...

Issue #1055, opened 2020-01-12
[Bug] SecureStorage.SetAsync fails on Tizen if key already exists


DamianMehers:
### Description
Calling `SecureStorage.SetAsync` with a key that already exists fails on Tizen (`[Tizen.Security.SecureRepository] Failed to save data. alias=test, error=Provided alias already exists in the database`)

### Steps to Reproduce

Create a simple Xamarin app that calls 

``` c#
    await Xamarin.Essentials.SecureStorage.SetAsync("test", "one");
    await Xamarin.Essentials.SecureStorage.SetAsync("test", "two");
```

### Expected Behavior

Calling `SetAsync` twice with the same key should overwrite the existing entry.

### Actual Behavior
On Tizen you get an exception

```
D/kanbann (26499):    at Interop.CheckNThrowException(Int32 err, String msg)
D/kanbann (26499):    at Tizen.Security.SecureRepository.DataManager.Save(String alias, Byte[] data, Policy policy)
D/kanbann (26499):    at Xamarin.Essentials.SecureStorage.PlatformSetAsync(String key, String data)
D/kanbann (26499):    at Xamarin.Essentials.SecureStorage.SetAsync(String key, String value)
D/k
```

### Basic Information

Tizen.NET 6.0.0.14995
Xamarin Essentials 1.3.1

DamianMehers:
The underlying [DataManager.Save](https://developer.tizen.org/dev-guide/csapi/api/Tizen.Security.SecureRepository.DataManager.html) API has no way to overwrite the existing value, so the solution may be to remove and then save.  That is what I will do to work around this behavior.

DamianMehers:
I've created a fix and added a test in my local repository [here](https://github.com/DamianMehers/Essentials/pull/1/commits) but I need to test.  I've not had time to set up my dev machine to be able to build and test Essentials locally.  Can I easily run my changes through the Essentials CI bot?

jamesmontemagno:
@DamianMehers if you do a PR it will fire off a CI build :)

DamianMehers:
> @DamianMehers if you do a PR it will fire off a CI build :)

[Done](https://github.com/xamarin/Essentials/pull/1061).  Builds and tests pass now.  Thanks!

PR #1061, opened 2020-01-15
Allow tizen secure storage overwrite

DamianMehers:
### Description of Change ###

Prior to saving a value first try to remove it on tizen so that overwriting an existing value works.

Also fix GetAsync to return `null` if the key does not exist.

### Bugs Fixed ###

https://github.com/xamarin/Essentials/issues/1055
### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)
- [X] Has samples (if omitted, state reason in description): No API surface change
- [X] Updated documentation ([see walkthrough](https://github.com/xamarin/Essentials/wiki/Documenting-your-code-with-mdoc)): No API surface change


DamianMehers:
I'm creating this merge requests to allow CI build/test to run since I've not set up my dev machine for Essentials yet.  Underyling bug is still in triage state.

Xamarin.Essentials/SecureStorage/SecureStorage.tizen.cs
```diff
@@ -23,6 +23,14 @@ static Task PlatformSetAsync(string key, string data)
         {
             try
             {
+                try
+                {
+                    // Swallow the exception since we are just removing it if it exists
+                    DataManager.RemoveAlias(key);
```
jamesmontemagno: Is it good to check if it exists first? I would also log Debug message on the catch.

DamianMehers: Re checking first, I wan't sure if we wanted to triple the cost of this call, instead of doubling as it stands - getting it throws an exception if it doesn't exist so you wouild be pretty much back to the current situation.

My thinking is that logging a failed delete will polute the log because it will be happening  all the time as values are overwritten. If I were to call PlatformRemove instead then that would do the log.  This is what I did in the earlier version, but decided to make it silent since it is not actually an error.  If you'd prefer I can log a message though - I'm not sure it is actionable.

DamianMehers: I'm open to implementing either of your suggestions after you've considered my reasoning.  Let me know.

jamesmontemagno:
You can grab and test the nuget from the CI build: https://dev.azure.com/xamarin/public/_build/results?buildId=12699&view=artifacts&type=publishedArtifacts

Xamarin.Essentials/SecureStorage/SecureStorage.tizen.cs
Redth: Exceptions aren't necessarily cheap either, so my gut says we should probably check before removing.

DamianMehers: > Exceptions aren't necessarily cheap either, so my gut says we should probably check before removing.

I know - that is why I'm not sure that checking makes sense because checking throws an exception if the key is not there.  I'll update the code as you suggest though :-)

DamianMehers: Let me know if you'd prefer that I use the native calls rather than the platform wrappers.

DamianMehers: This won't work, since the Tizen `PlatformGetAsync` looks to throw an Exception for non-existent values (a separate bug).  I'll update the code again.

DamianMehers:
@jamesmontemagno @Redth it seems to me that the Tizen `PlatformGetAsync` is broken too because it throws an exception if the key does not exist, rather than returning null.

Would you prefer that I create a new bug/PR, or update this PR to add a new test which does a 

``` c#
PlatformGetAsync(Guid.New.Guid().ToString());
```
and fix the Tizen `PlatformGetAsync` to return `null` if the underlying `DataManager.Get` throws `InvalidOperationException`?

Redth:
Hmmm is there any method on DataManager to check if the key exists?

If not yeah I guess we need to catch the Get.  Same PR is fine if you’re willing to do the work, thanks!

DamianMehers:
@Redth @jamesmontemagno ... ready for a new review. 

I've fixed the `GetAsync` to return null when the key does not exist, and  I've added comments to the code to clarify Tizen's behavior - the Tizen `DataManager` API has no way to check for a key's existence.

DamianMehers:
> Thanks for figuring out the behaviour here and submitting the PR!

Thanks for approving.  I'll pull in the nuget to the app I'm building (A Trello client for Tizen - I've already got it running on WatchOS and Wear OS thanks to Xamarin) and double-check.

ppaneksamsung:
@DamianMehers I tried to use SecureStorage API on Tizen. I'm not sure if your intention was to return a null Task, when a key does not exist. In other words:

`var value = await SecureStorage.GetAsync("SomeNotExistingKey");`

throws NullReferenceException, while I would expect that a task completes with no exception and a null value is returned. Is it intentional? If not, then probably:

```
            catch (System.InvalidOperationException)
            {
                // The DataManager.Get call throws an exception if key does not exist. Not logging
                // anything since this is an expected and normal situation (if the key did not exist).
                return null;
            }
```

needs to be changed to something like:

```
            catch (System.InvalidOperationException)
            {
                // The DataManager.Get call throws an exception if key does not exist. Not logging
                // anything since this is an expected and normal situation (if the key did not exist).
                return Task.FromResult<string>(null);
            }
```

mattleibow:
@ppaneksamsung is correct, let me fix that. Thanks for spotting it!

DamianMehers:
> @DamianMehers I tried to use SecureStorage API on Tizen. I'm not sure if your intention was to return a null Task, when a key does not exist. In other words:
> 
> `var value = await SecureStorage.GetAsync("SomeNotExistingKey");`
> 
> throws NullReferenceException, while I would expect that a task completes with no exception and a null value is returned. Is it intentional? If not, then probably:
> 
> ```
>             catch (System.InvalidOperationException)
>             {
>                 // The DataManager.Get call throws an exception if key does not exist. Not logging
>                 // anything since this is an expected and normal situation (if the key did not exist).
>                 return null;
>             }
> ```
> 
> needs to be changed to something like:
> 
> ```
>             catch (System.InvalidOperationException)
>             {
>                 // The DataManager.Get call throws an exception if key does not exist. Not logging
>                 // anything since this is an expected and normal situation (if the key did not exist).
>                 return Task.FromResult<string>(null);
>             }
> ```

Yep - dumb bug on my part.  Apologies.

Issue #662, opened 2018-12-22
Additional Platform Support


jamesmontemagno:
With Xamarin.Essentials our priority on platform support is focused on iOS, Android, and UWP. We understand that there are many requests for additional platforms. Some are being investigated, some in development (by the Essentials team and some by community). We are going to use this as single place for platform support so we don't have a bunch of open issues.

* [x] iOS - Supported By Microsoft
* [x] Android - Supported By Microsoft
* [x] UWP - Supported By Microsoft
* [x] [macOS](https://github.com/xamarin/Essentials/pull/653)
* [ ] [WinForms/WPF](https://github.com/xamarin/Essentials/pull/1079)
* [ ] GTK#
* [x] [Tizen](https://github.com/xamarin/Essentials/pull/555) - Supported by Samsung
* [x] watchOS - Supported By Microsoft
* [x] tvOS - Supported By Microsoft
* [ ] .NET Core 
* [ ] Web Assembly

AndreiMisiukevich:
Hi James @jamesmontemagno 
I'd like to see Connectivity and Clipboard support for Mac and Gtk
0.0) **Mac Clipboard** is already done by @mattleibow in his fork
0.1) **Mac connectivity** here is PR https://github.com/mattleibow/Essentials/pull/1

Where should I create PR for point 1
1.0) **Gtk Clipboard**
1.1) **Gtk Connectivity**

May i do it to this project or to anybody's fork?

Thx

VladislavAntonyuk:
Since Xamarin Forms made Device.OpenUri obsolete, Xamarin Essentials should implement this feature for all platforms (macOS, WPF, GTK, etc)

danielmeza:
Since GTK run in a variety of platforms principality not mobile like Linux and Windows, it result difficult to find the right way to start implement components specially in Linux platforms. @jamesmontemagno  there is perhaps a get started guide or research document to post here and give a path to the community in order to speed up the collaboration to this platform?

JunielKatarn:
@danielmeza @jamesmontemagno 
I plan to invest some free time to make GTK# happen to add another supported platform to my Xamarin.Forms published app.

Can someone point me in the right direction?
I plan to develop using a Linux environment, and willing to work on the Xamarin.Forms potentially missing pieces.

JunielKatarn:
> Where should I create PR for point 1
> 1.0) **Gtk Clipboard**
> 1.1) **Gtk Connectivity**

[...]

PR #1079, opened 2020-02-03
Added Windows Desktop (WPF) platform implementation

peterfoot:
### Description of Change ###

Adds platform support for Windows desktop (WPF) apps.

### Bugs Fixed ###

- Related to issue #662

Adds the following platform implementations

 - [ ] Accelerometer
 - [x] AppInfo
 - [ ] Barometer
 - [x] Battery
 - [x] Browser
 - [x] Clipboard
 - [ ] Compass
 - [x] Connectivity
 - [x] DeviceDisplay
 - [x] DeviceInfo
 - [x] Email
 - [x] FilePicker
 - [x] FileSystem
 - [ ] Flashlight
 - [ ] Geocoding
 - [x] Geolocation
 - [ ] Gyroscope
 - [x] Launcher
 - [ ] Magnetometer
 - [x] MainThread
 - [x] Map
 - [ ] OrientationSensor
 - [ ] Permissions
 - [x] PhoneDialer
 - [ ] Platform
 - [x] Preferences
 - [x] SecureStorage
 - [x] Screenshot
 - [ ] Share
 - [ ] Shake
 - [x] Sms
 - [x] TextToSpeech
 - [x] VersionTracking
 - [ ] Vibration
 - [ ] WebAuthenticator / AppleSignInAuthenticator

### API Changes ###

Added DevicePlatform.WindowsDesktop

### Behavioral Changes ###

N/A

### PR Checklist ###


Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Collections.Generic;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Preferences
+    {
+        static Dictionary<string, object> preferences = new Dictionary<string, object>();
```
mattleibow: Is this not actually saved to disk at all?

Xamarin.Essentials/Email/Email.win32.cs
```diff
@@ -0,0 +1,21 @@
...
+    public static partial class Email
+    {
+        internal static bool IsComposeSupported =>
+            throw ExceptionUtils.NotSupportedOrImplementedException;
+
+        static Task PlatformComposeAsync(EmailMessage message) =>
+            throw ExceptionUtils.NotSupportedOrImplementedException;
+    }
+
+#if NETSTANDARD1_0 || NETSTANDARD2_0
```
mattleibow: Not sure what this #if is for? This file should not be included for those defines.

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -106,6 +106,22 @@
...
+    <Reference Include="System.Management" />
+    <Reference Include="System.Speech" />
+    <Reference Include="System.Web" />
+    <Reference Include="System.Windows.Forms" />
+    <Reference Include="System.Xaml">
+      <RequiredTargetFramework>4.0</RequiredTargetFramework>
+    </Reference>
+    <Reference Include="WindowsBase" />
+    <Reference Include="PresentationCore" />
+    <Reference Include="PresentationFramework" />
```
mattleibow: Just checking to see if we actually use the presentation bits for this project? We shouldn't be using any of the Xaml-related things. Also, this _should_ work for winforms without this?

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid60;MonoAndroid70;MonoAndroid71;MonoAndroid80;MonoAndroid81;MonoAndroid90;MonoAndroid10.0;tizen40;</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid60;MonoAndroid70;MonoAndroid71;MonoAndroid80;MonoAndroid81;MonoAndroid90;MonoAndroid10.0;tizen40;net461</TargetFrameworks>
```
mattleibow: We will need to think about how we do this... net461 works in too many places.

Xamarin.Essentials/Xamarin.Essentials.csproj
peterfoot: At the moment it uses WPF dependencies for MainThread, to support either WinForms or WPF will require some reworking but this is a sensible point. I will investigate alternative methods of identifying the thread.

Xamarin.Essentials/Preferences/Preferences.win32.cs
peterfoot: Need some discussion over the implementation of this. This is temporary and required for the code to build and allow the sample app to run. There isn't a native API we can use (unless taking a dependency on the Windows 10 UWP API). A simple implementation would be to persist the dictionary(s) to a file but that may require a dependency on Newtonsoft.Json or similar. I think that will be the easiest option for now and we can discuss further the pros/cons.

Xamarin.Essentials/Email/Email.win32.cs
peterfoot: Sorry, that file is a placeholder copy-pasted. I'll add in the Email implementation today.

peterfoot: Replaced with actual Win32 implementation

Xamarin.Essentials/Xamarin.Essentials.csproj
DamianMehers: Taking a dependency on Newtonsoft.Json in Essentials will make it unusable in some platforms such as WatchOS because the AOT generates a massive amount of code https://github.com/xamarin/xamarin-macios/issues/7250

Since only a limited number of types are supported in Preferences (bool, double, int, float, long, string, and DateTime) could you manually convert them to/from strings?

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -106,6 +106,23 @@
     <Compile Include="**\*.tizen.cs" />
     <Compile Include="**\*.tizen.*.cs" />
   </ItemGroup>
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net461')) ">
+    <Compile Include="**\*.win32.*.cs" />
+    <Compile Include="**\*.win32.cs" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
DamianMehers: Taking a dependency on Newtonsoft.Json in Essentials will make it unusable in some platforms such as WatchOS because the AOT generates a massive amount of code xamarin/xamarin-macios#7250

Since only a limited number of types are supported in Preferences (bool, double, int, float, long, string, and DateTime) could you manually convert them to/from strings?

Xamarin.Essentials/Xamarin.Essentials.csproj
peterfoot: I assume this will work with some more detailed tweaking to create multiple net461 dlls in the runtimes folder of the NuGet.

Xamarin.Essentials/Preferences/Preferences.win32.cs
Mrnikbobjeff: No need for newtonsoft, BinarySerializer etc. are all available and fit the scenario. Persistance is IMHO necessary for this to be considered workable.

mattleibow: You could have a look here: https://docs.microsoft.com/en-us/dotnet/standard/serialization/

peterfoot: Thanks, I'll rework this and use one of the built-in options.

Mrnikbobjeff: Going the file route comes with another set of problems:
We definitely have to have a lazy mechanism to store and retrieve settings. Loading all properties on first acess would increse the working set unnecessarily. Also, we do need fast key existance checks. Not only for Key enumeration but also for insertion. I am currently hooking up some different benchmarks for a couple of potential solutions, but this all seems very icky to me. Are there any usbale storage technologies available with winapi we could use? or even com perhaps

Mrnikbobjeff: Okay this could be a stupid idea but how about the registry? Fast key existance checks, key value store just as we like it and as persistent as it can be

peterfoot: I think that's frowned upon, we can't control how much it gets used for and unless the app is installed via Centennial it'll get dumped into the main registry.

Mrnikbobjeff:
Geolocation and coding is available via the com interfaces in locationapi.h as far as I know.  Magnetometers are available in the sensorapi,h header via correspoding com objects
https://docs.microsoft.com/en-us/windows/win32/api/sensorsapi/

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -1,43 +1,63 @@
...
 using System.IO;
-using Newtonsoft.Json;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Formatters.Binary;
 
 namespace Xamarin.Essentials
 {
     public static partial class Preferences
     {
         static readonly string settingsPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), AppInfo.Name, "settings.dat");
```
DamianMehers: Any chance that AppInfo.Name will contain invalid chars for win32 file names, such as a ":" ?

peterfoot: Good catch, the documentation for the attribute says "The assembly title is a friendly name, which can include spaces" but that's not particularly informative.

DamianMehers: It could also be blank ... perhaps fallback to using the file name of the launching assembly if the sanitized name is blank?

``` c#
      var name = System.IO.Path.GetFileNameWithoutExtension(Assembly.GetEntryAssembly().CodeBase);
```

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Preferences
+    {
+        static readonly string settingsPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), AppInfo.Name, "settings.dat");
+
+        static readonly Dictionary<string, Dictionary<string, object>> preferences = new Dictionary<string, Dictionary<string, object>>();
```
mattleibow: This is a tad bit nit-picky, but it might be nicer to just add a usings to the top of the file:

```csharp
using PreferencesDictionary = System.Collections.Generic.Dictionary<string, System.Collections.Generic.Dictionary<string, object>>;
```
This way, we don't have angle brackets all over the show.

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                return string.Empty;
+            }
+
+            return sharedName;
+        }
+
+        static bool PlatformContainsKey(string key, string sharedName)
+        {
+            if (preferences.ContainsKey(CleanSharedName(sharedName)))
```
mattleibow: This could be a TryGetValue:

```csharp
return preferences.TryGetValue(sharedName, out var inner) && inner.ContainsKey(key);
```

Xamarin.Essentials/Preferences/Preferences.win32.cs
mattleibow: This is a tad bit nit-picky, but it might be nicer to just add a usings to the top of the file:

```csharp
using PreferencesDictionary = System.Collections.Generic.Dictionary<string, System.Collections.Generic.Dictionary<string, object>>;
```
This way, we don't have angle brackets all over the show.

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        static void Save()
+        {
+            using (var stream = File.OpenWrite(settingsPath))
+            {
+                var formatter = new BinaryFormatter();
+                formatter.Serialize(stream, preferences);
+                stream.Close();
```
mattleibow: The dispose will close here.

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            if (File.Exists(settingsPath))
+            {
+                using (var stream = File.OpenRead(settingsPath))
+                {
+                    try
+                    {
+                        var formatter = new BinaryFormatter();
+                        var readPreferences = (Dictionary<string, Dictionary<string, object>>)formatter.Deserialize(stream);
+                        stream.Close();
```
mattleibow: no need to close, the dispose will do it

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                return preferences[sharedName].ContainsKey(key);
+            }
+
+            return false;
+        }
+
+        static void PlatformRemove(string key, string sharedName)
+        {
+            if (preferences.ContainsKey(CleanSharedName(sharedName)))
```
mattleibow: Maybe use the TryGetValue here so save a lookup.

```csharp
if (prefs.TryGetValue(sN, out var inner)) {
    inner.Remove(key);
    Save();
}
```

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        static void PlatformSet<T>(string key, T value, string sharedName)
+        {
+            if (!preferences.ContainsKey(CleanSharedName(sharedName)))
+            {
+                preferences.Add(CleanSharedName(sharedName), new Dictionary<string, object>());
+            }
+
+            if (preferences[CleanSharedName(sharedName)].ContainsKey(key))
```
mattleibow: Maybe use a TryGetValue for these.

In fact, just use the `[key] = value` here as it will add any missing keys. No need to check and then branch code.

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+            Save();
+        }
+
+        static T PlatformGet<T>(string key, T defaultValue, string sharedName)
+        {
+            if (preferences.ContainsKey(CleanSharedName(sharedName)))
+            {
+                if (preferences[CleanSharedName(sharedName)].ContainsKey(key))
+                {
+                    return (T)preferences[CleanSharedName(sharedName)][key];
```
mattleibow: Maybe a `if (prefs.TryGetValue() && inner.TryGetValue())`

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                using (var stream = File.OpenRead(settingsPath))
+                {
+                    try
+                    {
+                        var formatter = new BinaryFormatter();
+                        var readPreferences = (Dictionary<string, Dictionary<string, object>>)formatter.Deserialize(stream);
+                        stream.Close();
+
+                        if (readPreferences is object)
```
Mrnikbobjeff: Just check for null the normal way. https://github.com/xamarin/Essentials/blob/master/CONTRIBUTING.md

Xamarin.Essentials/Preferences/Preferences.win32.cs
```diff
@@ -1,16 +1,18 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
+using PreferencesDictionary = System.Collections.Generic.Dictionary<string, System.Collections.Generic.Dictionary<string, object>>;
```
Mrnikbobjeff: Something to consider would be using IStrongBox<T> as value container, as we can ellide boxing/unboxing some of the time this way: IStrongBox<T> {public T Value;}
When reading a key we can instead do (value as IStrongBox<T>).Value. This pattern is common in storing value types without boxing as often.

Mrnikbobjeff:
An easy grab is Telephone and sms, simply invoke the launcher with the corresponding url. Same goes for sms. An improvement to launcher can be made by checking the registry for a corrsponding HKEY_CLASSES_ROOT\%protocol%\shell\open\command registry key to check whether launching said url is supported

Xamarin.Essentials/Geolocation/Geolocation.win32.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Device.Location;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Geolocation
```
Mrnikbobjeff: Consider using the com interfaces for this, we avoid having a dependency for this. The actual implementation uses the COM service as well. Very Optional though as the C# wrapper is quite a lot easier to deal with.

Redth:
A few notes as @mattleibow and I have been discussing:

1. Preferences should really serialize to json or some text format, as binary serialization has some dragons especially moving to .net 5
2. SecureStorage - we might be able to do something on a per user basis here, Potentially using [CNG-DPAPI](https://docs.microsoft.com/en-us/windows/win32/seccng/cng-dpapi)
3. We should consider how to handle Local vs Roaming profiles for preferences, and Secure Storage,  etc.

peterfoot:
@Redth Thanks, Json was rejected before because Json.net adds a dependency and System.Text.Json not available on older .NET versions. The only consistently available text format in-box is XML...

DPAPI/ProtectedData for SecureStorage makes sense, again physical storage is an issue.

I assume all of these will be non-roaming as the other platforms don't roam their preferences?

Redth:
What about DataContractJsonSerializer? Or yeah System.Text.Json. 

Other platforms implicitly can ‘roam’ their data depending on where it’s stored

peterfoot:
@Redth - yes DataContractJsonSerializer looks like a good option.

Looking at the current implementations for Essentials they are all using private/local storage. I assume for WPF we want to match the behaviour of the existing UWP implementation where possible.

Xamarin.Essentials/Xamarin.Essentials.csproj
VladislavAntonyuk: we should also add .Net Core WPF

mattleibow:
System.Text.Json does work on netstandard2.0 and net4.6.1

https://www.nuget.org/packages/System.Text.Json

peterfoot:
@mattleibow Given the simple use-case here is it worth using System.Text.Json (introducing a NuGet dependency) or stick with the built-in DataContractJsonFormatter?

Xamarin.Essentials/Geolocation/Geolocation.win32.cs
peterfoot: @Mrnikbobjeff The System.Devices dll is a built in part of the .NET framework so doesn't introduce a dependency, using the COM interfaces will require more code to achieve the same result.

mattleibow:
@peterfoot I would just use the contract serializer for now. Just wanted to say that it should work... But as I type this, I somehow feel that there was some issue with the serializer... Like dates... Not sure if anyone knows... Maybe it was around deserialization. But since we serialize and deserialize ourselves, we should be good.

Mrnikbobjeff:
@mattleibow as we specify which types we accept (and all types we support are sealed) we do not encounter most serialisation issues such as ignoring validation when deserializing objects or expecting different types upon deserialization. Any text format should work without problems.

Xamarin.Essentials/Battery/Battery.win32.cs
```diff
@@ -0,0 +1,128 @@
...
+            get
+            {
+                NativeMethods.GetSystemPowerStatus(out var powerStatus);
+                return powerStatus.SystemStatusFlag == 1 ? EnergySaverStatus.On : EnergySaverStatus.Off;
+            }
+        }
+
+        static class NativeMethods
+        {
+            [DllImport("Kernel32")]
```
Mrnikbobjeff, line 93: Missing exactspelling attribute for efficient stub lookup

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+        }
+
+        static void StopScreenMetricsListeners()
+        {
+        }
+
+        static class NativeMethods
+        {
+            [DllImport("Kernel32", SetLastError = true)]
```
Mrnikbobjeff, line 62: Add ExactSpelling

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+            internal static extern EXECUTION_STATE SetThreadExecutionState(EXECUTION_STATE esFlags);
+
+            internal enum EXECUTION_STATE : uint
+            {
+                CONTINUOUS = 0x80000000,
+                DISPLAY_REQUIRED = 0x00000002,
+            }
+
+            [DllImport("user32")]
```
Mrnikbobjeff, line 71: Add ExactSpelling

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+            internal static extern IntPtr GetDesktopWindow();
+
+            [DllImport("user32")]
```
Mrnikbobjeff, line 74: Add ExactSpelling

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+            internal static extern IntPtr MonitorFromWindow(IntPtr hwnd, MONITOR_DEFAULTTO dwFlags);
+
+            internal enum MONITOR_DEFAULTTO : int
+            {
+                NULL = 0,
+                PRIMARY = 1,
+                NEAREST = 2,
+            }
+
+            [DllImport("user32")]
```
Mrnikbobjeff, line 84: Add ExactSpelling, switch to GetMonitorInfoW for less string marshalling overhead

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+            internal struct RECT
+            {
+                internal int left;
+                internal int top;
+                internal int right;
+                internal int bottom;
+            }
+
+            [DllImport("gdi32")]
+            internal static extern IntPtr CreateDC(string lpszDriver, string lpszDevice, string lpszOutput, IntPtr lpInitData);
```
Mrnikbobjeff, line 107: Switch to CreateDCW

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            [DllImport("gdi32")]
```
Mrnikbobjeff, line 109: ExactSpelling

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.win32.cs
```diff
@@ -0,0 +1,124 @@
...
+            internal static extern bool DeleteDC(IntPtr hdc);
+
+            [DllImport("gdi32")]
+            internal static extern int GetDeviceCaps(IntPtr hdc, DeviceCap nIndex);
```
Mrnikbobjeff, line 113: ExactSpelling

Mrnikbobjeff (commented):
Some interop droplets

peterfoot:
Thanks @Mrnikbobjeff I think I've fixed up all these interop points. I've also reworked it against the latest in main. And added the FilePicker implementation for Win32 😀

Mrnikbobjeff:
Thank you for taking time to implement these changes. I would provide a Screenshot implementation of you have not started working on it yet, that is a very easy task. I will then create a merge request into this branch

peterfoot:
@Mrnikbobjeff that would be great! It's a feature which wasn't in Essentials when I first created the PR and will be really useful.

Mrnikbobjeff:
> @Mrnikbobjeff that would be great! It's a feature which wasn't in Essentials when I first created the PR and will be really useful.

You have your PR, I also added a missing charset.unicode parameter I missed in the review

peterfoot:
@Mrnikbobjeff awesome thanks!

Mrnikbobjeff:
@peterfoot I created a second pr to check for sms and telephone capabilities.

Mrnikbobjeff:
I finished working on the first sensor implementation, I believe we are able to extend support to the following sensors:
Accelerometer, Barometer, Compass, Gyroscope, Magnetometer, OrientationSensor, Vibration (when a xinput device is present). They all offer com interfaces which sensor device developers can target

Mrnikbobjeff:
I polished the com interop a little and startet a pr. You can see the sensor groundwork [here](https://github.com/inthehand/Essentials/blob/5694bfa6a761db187141b2261ff3676483e6be8b/Xamarin.Essentials/Barometer/Barometer.win32.cs). Adding the other sensors should be easy now, but it would be wonderful if someone checks out the path we are taking here as sensor support is approximately 3k loc

peterfoot:
@Mrnikbobjeff I suppose the thing is - how many Windows devices will have those sensors? And of those, how many running version prior to Windows 10? I'm looking at NuGet runtime support so we can generate multiple .NET Framework assemblies - this will ensure there is a separate windows and linux version. Also means there can be a Windows 7 and above version and a Windows 10 specific version which can use the UWP APIs - I think all the sensor APIs don't require a package id so can be used from desktop apps.

Mrnikbobjeff:
> @Mrnikbobjeff I suppose the thing is - how many Windows devices will have those sensors? And of those, how many running version prior to Windows 10? I'm looking at NuGet runtime support so we can generate multiple .NET Framework assemblies - this will ensure there is a separate windows and linux version. Also means there can be a Windows 7 and above version and a Windows 10 specific version which can use the UWP APIs - I think all the sensor APIs don't require a package id so can be used from desktop apps.

The sensor API is available from Windows 7 onward

peterfoot:
@Mrnikbobjeff yes, but my question was that if we can use the already written UWP API for Windows 10, is it worth writing the Win32 ones. It would really be only a subset of tablets/laptops with the hardware and how many would be running Windows 7/8?
The macOS implementation doesn't have sensors as I don't think they are as relevant for desktop apps.

Mrnikbobjeff:
> The macOS implementation doesn't have sensors as I don't think they are as relevant for desktop apps.

MacOs does not offer an abstraction over general p&p devices which device drivers can target for these sensors. That is the limiting factor IMHO. Also, writing new sensors once the base implementation is done is achieved in [less than 20 loc](https://github.com/inthehand/Essentials/blob/5694bfa6a761db187141b2261ff3676483e6be8b/Xamarin.Essentials/Barometer/Barometer.win32.cs#L2838) up to and including BarometerSensor

peterfoot:
@Mrnikbobjeff I'm not against implementing sensor APIs, I just wonder how many devices will be able to take advantage of them

Mrnikbobjeff:
> @Mrnikbobjeff I'm not against implementing sensor APIs, I just wonder how many devices will be able to take advantage of them

Very few at best, the most common sensors appear to be biometrics and luminosity sensors. My work device seems to have them both, and biometrics seem to be supported by other devices (primarily laptops) as well.

follesoe:
Just want to say this would be highly appreciated and valued. Given that WPF is part of Xamarin.Forms (even if only as preview) it makes perfect sense to get Essentials supported on WPF as well. At Blueye we are using WPF over UWP due to limitations in some third-party libraries which prevents us from using Store Distribution, and side-loading of UWP is not as nice/trivial as with WPF.

I would be glad to help out with the implementation of some of the missing pieces. Would be great with some guidance on what would be required in order to have this merged (i.e. does all APIs need to be ported)?

Thank you for your effort so far @peterfoot! It has already been of value as we have adopted parts of your implementation already while awaiting the merge of this PR.

arivera12:
Oh yes! Xamarin essentials is coming to WPF! When this is planned to be released?

acffordyce973:
Are there any updates for Xamarin Essentials supporting WPF?

peterfoot:
@acffordyce973 the PR was a bit stale so I've updated against the latest main branch so it should unblock the conflicts at least.

Issue #1014, opened 2019-12-11
DeviceInfo.DeviceType which uses EasClientDeviceInformation hangs on UWP 


gmarbury:
Bug report best practices: [Submitting Issues](https://github.com/xamarin/Essentials/blob/master/SUBMITTING-ISSUES.md)

### Description

System.Exception  HResult=0x8001010A  Message=The message filter indicated that the application is busy.A COM call (IID: {54DFD981-1968-4CA3-B958-E595D16505EB}, method index: 10) to an ASTA (thread 36768) appears deadlocked and was timed out.  Source=Windows  StackTrace: at Windows.Security.ExchangeActiveSyncProvisioning.EasClientDeviceInformation.get_SystemProductName() at Xamarin.Essentials.DeviceInfo.GetDeviceType() at Xamarin.Essentials.DeviceInfo.get_DeviceType() at 
--

### Steps to Reproduce

1. 
2. 
3. 

### Basic Information

- Version with issue:
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

jamesmontemagno:
I can not reproduce that here as we have manual tests for it. How are you using this?

gmarbury:
Thanks for the quick response.
I don't have a small test case. We will make one to give you. We were trying to determine if the device was virtual during some lazy property init. So its really hard to say what steps generated it.  It would only happen on a real device, mostly on release build. But occasionally running from code. I did notice that after rebooting the PC, it would work running from code. It has been verified on at least three machines too.
It's also possible the call was being made on a timer fired thread.

tipa:
I am experiencing a similar error message, but with a different stack trace, maybe it helps finding the problem:
```
System.Exception: The message filter indicated that the application is busy. A COM call (IID: {638BB2DB-451D-4661-B099-414F34FFB9F1}, method index: 6) to an ASTA (thread 4640) appears deadlocked and was timed out.

System.Runtime.InteropServices   McgMarshal.ThrowOnExternalCallFailed (Int32, RuntimeTypeHandle) + 0x:21
__Interop   ComCallHelpers.Call (__ComObject, RuntimeTypeHandle, Int32, Void*) + 0xbe
+ 0x:37
Xamarin.Essentials  MainThread.get_PlatformIsMainThread () + 0x:90
[...]

PR #1087, opened 2020-02-07
GH-1014 Try to cache device type on UWP

jamesmontemagno:
### Description of Change ###

An exception may occurs (perhaps backgrounding), so cache it as it never is going to change.

### Bugs Fixed ###

- Related to issue #1014

### Behavioral Changes ###

May return unknown if it can't figure it out if an exception is thrown

### PR Checklist ###


mattleibow:
We should probably be caching more with the device. This will never change.

PR #1095, opened 2020-02-10
AndroidX

Redth:
Make Essentials compile against AndroidX for MonoAndroid10.0 projects.

### DIFF
--- a/DeviceTests/DeviceTests.Android/Tests/FileProvider_Tests.cs
+++ b/DeviceTests/DeviceTests.Android/Tests/FileProvider_Tests.cs
@@ -185,7 +185,11 @@ namespace DeviceTests.Shared
         public void Get_Existing_External_Shareable_Uri(FileProviderLocation location)
         {
             // Save an external directory file
+
+            #if !__ANDROID_29__
             var externalRoot = AndroidEnvironment.ExternalStorageDirectory.AbsolutePath;
+            #endif
+
             var root = Platform.AppContext.GetExternalFilesDir(null).AbsolutePath;
             var file = CreateFile(root);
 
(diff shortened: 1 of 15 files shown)


jamesmontemagno:
Would be worth writing up some info in the readme.txt about AndroidX I think.......

DeviceTests/DeviceTests.Android/Tests/FileProvider_Tests.cs
```diff
@@ -185,7 +185,11 @@ public void Get_Existing_External_Cache_Shareable_Uri(FileProviderLocation locat
         public void Get_Existing_External_Shareable_Uri(FileProviderLocation location)
         {
             // Save an external directory file
+
+            #if !__ANDROID_29__
             var externalRoot = AndroidEnvironment.ExternalStorageDirectory.AbsolutePath;
```
mattleibow: Is this deprecated in API 29? Is there an alternative we should be using?

Xamarin.Essentials/Types/FileProvider.android.cs
```diff
@@ -13,9 +18,9 @@ namespace Xamarin.Essentials
         Exported = false,
         GrantUriPermissions = true)]
     [MetaData(
-        "android.support.FILE_PROVIDER_PATHS",
+        "android.support.FILE_PROVIDER_PATHS", // IMPORTANT: This string doesn't change with AndroidX
```
mattleibow: Good note!

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid60;MonoAndroid70;MonoAndroid71;MonoAndroid80;MonoAndroid81;MonoAndroid90;MonoAndroid10.0;tizen40;</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid80;MonoAndroid81;MonoAndroid90;MonoAndroid10.0;tizen40;</TargetFrameworks>
```
mattleibow: Dropping the 6.0 to 7.1 as this references really old Android Support. It will still RUN on older devices, but we really shouldn't be compiling for this. The tools are really old.

PR #1101, opened 2020-02-16
four more minor changes

Mrnikbobjeff:
### Description of Change ###
Both fields should be readonly, class should be static, That type of collection access should use hashset
### PR Checklist ###


Redth:
@mrnikbobjeff anything you can point to that justifies the switch to HashSet?

Mrnikbobjeff:
Why would we want to iterate through a list if we can perform constant time lookup

Issue #1099, opened 2020-02-13
[Enhancement] Provide accuracy of altitude in GeoLocation API


janusw:
## Summary

Currently, the `Xamarin.Essentials.Location` class only includes one `Accuracy` property, which (I assume) describes the horizontal (or overall?) accuracy of the location. However, the vertical accuracy (concerning the altitude) can differ from the horizontal accuracy quite substantially, e.g. depending on whether a barometer is available or not.

## API Changes

In the simplest case, an additional property `Xamarin.Essentials.Location.AltitudeAccuracy` could be added.

Alternatively, one could introduce a new class for representing altitude information, which could include properties like `Value` and `Accuracy` (or similar). This would be a more disruptive change of API, though.

## Intended Use Case

The primary use case would be any situation where the user needs a good estimate of how accurate his own location is known, both horizontally and vertically. I'm mainly having augmented-reality scenarios in mind, but there are certainly other applications where this is important.

janusw:
For Android, there is [Android.Locations.Location.hasVerticalAccuracy](https://developer.android.com/reference/android/location/Location.html#hasVerticalAccuracy()) and [Android.Locations.Location.getVerticalAccuracyMeters](https://developer.android.com/reference/android/location/Location.html#getVerticalAccuracyMeters()).

For iOS: [CoreLocation.CLLocation.verticalAccuracy](https://developer.apple.com/documentation/corelocation/cllocation/1423550-verticalaccuracy)

janusw:
And finally UWP: [Geocoordinate.AltitudeAccuracy](https://docs.microsoft.com/en-us/uwp/api/windows.devices.geolocation.geocoordinate.altitudeaccuracy)

mattleibow:
Ooh, nice. Is there a PR coming soon?

janusw:
> Ooh, nice. Is there a PR coming soon?

Um, yeah. Probably soon. As you can see I've already started the implementation (but haven't tested it yet). Do you think the commit above is sufficient, or is anything missing?

mattleibow:
It looked good so far. Not sure what Tizen would do - or if they even have this feature. But, yeah, open a PR as soon as you think it good enough. The community will give some extra feedback - if they have any.

Thanks for the work so far!

janusw:
> It looked good so far. Not sure what Tizen would do - or if they even have this feature.

For `LocationExtensions`, there is not even a .tizen.cs file, so I wouldn't really know where to start ...

[...]

PR #1103, opened 2020-02-18
Implement vertical accuracy in GeoLocation API

janusw:
### Description of Change ###

This extends the GeoLocation API to provide the vertical accuracy.

### Bugs Fixed ###

- Related to issue #1099

### API Changes ###

Added: 
 
- `double? Location.VerticalAccuracy { get; set; }`

Changed:

 - None
 

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Samples/Samples/ViewModel/GeolocationViewModel.cs
+++ b/Samples/Samples/ViewModel/GeolocationViewModel.cs
@@ -98,8 +98,9 @@ namespace Samples.ViewModel
             return
                 $"Latitude: {location.Latitude}\n" +
                 $"Longitude: {location.Longitude}\n" +
-                $"Accuracy: {location.Accuracy}\n" +
+                $"HorizontalAccuracy: {location.Accuracy}\n" +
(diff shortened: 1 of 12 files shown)


Redth:
Question about the value on each platform.  Are we sure this is accuracy in meters on iOS, Android and UWP?  Just want to double check we don't need any code to normalize values here.

janusw:
> Question about the value on each platform. Are we sure this is accuracy in meters on iOS, Android and UWP?

Yes, the links to the documentation of the API calls on each platform can be found in #1099. All three explicitly mention that the returned value is in meters.

janusw:
@Redth Thanks for the approval! Any chance to still get this into the 1.5.0 release?

Redth:
Yes I think we can put this into 1.5 still

janusw:
Thanks a lot for merging!

Unfortunately I discovered just in this very moment a problem with older Android versions: The compile-time check does not seem to be sufficient, and we'll need a runtime check in addition :/

Sorry about that. I'll open a follow-up PR with the fix shortly!

PR #1138, opened 2020-02-29
added HapticFeedback

dimonovdd:
### Description of Change ###

Added support for" HapticFeedback " for iOS, Android, and UWP.
But I couldn't test on UWP because I don't have a device that supports " HapticFeedback"

I hope I have correctly created a pull-request


Redth:
Hey @dimonovdd this is a good looking addition.  Are you willing to add tests and docs for this?

jamesmontemagno:
One question I have is how different is it from "Vibrate" api.

mattleibow:
I am thinking the same as @JamesMontemagno... Maybe it might be better to merge this?

Also, UWP is using the same named device, but a different entirely. The one we have is from phone and this one is from UWP general. I think we should keep both, or even drop phone since that is dead now.

Redth:
I think it makes sense to bring this into the Vibrate class perhaps as:

```csharp
public static class Vibration
{
    public static void HapticFeedback(HapticFeedbackType type = HapticFeedbackType.Click)
}
```

dimonovdd:
@Redth @jamesmontemagno @mattleibow 
I believe that vibration and hapticFeedback are different things and should not be merged. But I can trust your experience and do that.

I will write the documentation this weekend, but there is a problem with the tests, I have never written them, but I will try.

Tell me what else needs to be done.

Xamarin.Essentials/HapticFeedback/HapticFeedback.shared.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Xamarin.Essentials
+{
+    public static partial class HapticFeedback
+    {
+        public static void Execute(HapticFeedbackType type = HapticFeedbackType.Click)
```
mattleibow, line 7: Looking at the API s it stands, there seems to be an extra layer of "work"... enum to methods back to enum. I would just use the enum version and iOS can use a switch or split it up that side.

mattleibow:
@dimonovdd I think what we are more saying is that "vibrate" and "haptic feedback" are not the same, they are similar enough for the simplicity of Essentials.

I think it is fine to have `Vibration.Vibrate(TimeSpan)` AND `Vibration.Vibrate(HapticFeedbackType)` in the same place. Effectively, the same motors do the thing. And, this will give use the opportunity to "fake" the feedback on platforms. For example, this is only a think in iOS 13, but what happens on iOS 12? Should we fall back to a 500ms "normal" vibrate? I think this is maybe a bigger issue. We MUST merge the ideas because already UWP is broken on non-mobile devices - it can't vibrate because there is no API.

That is why I think not just because they are similar we should merge, but just because we actually need to. When a user does `HapticFeedback.Execute()` on iOS 12, what should happen? And, when a user does `Vibration.Vibrate()` on UWP, what should happen? I think there should be a fallback to the same "vibrate a bit".

Any thoughts on that? Anyone?

dimonovdd:
@mattleibow 

I think that the Vibrate Vibrate(HapticFeedbackType)  should not be made because it deludes. On iOS and Android the vibration and haptic feedback are absolutely different things, and even if it should be merged, it should not confuse.

on iOS the HapticFeadBack is supported only on devices that support iOS13 starting from iPhone. I haven't found any better way of checking than checking the iOS version which is why I left it that way for now. This is the link of the article

We'll return to UWP later when we find a more correct way of merging those Vibrate and HapticFeedback

Actually I thought and realized that it's worth writing a method that will determine whether Haptic Feedback is supported on a particular device. I'm going to work out this issue in the nearest future

dimonovdd:
An example of HapticFeadback is a response when you tap on the keyboard of iOS and Android devices, and if HapticFeadback did not work, in this case the vibration should not be called and even more a vibration for 500ms, the developer himself should decide

Xamarin.Essentials/HapticFeedback/HapticFeedback.shared.cs
dimonovdd: I'll think

DeviceTests/DeviceTests.Shared/HapticFeedback_Tests.cs
```diff
@@ -0,0 +1,45 @@
...
+            // API 23+ we need user interaction for camera permission
+            // can't really test so easily on device.
+            if (Platform.HasApiLevel(Android.OS.BuildVersionCodes.M))
+                return;
+#elif __IOS__
+            // TODO: remove this as soon as the test harness can filter
+            // the iOS simulator does not emulate a flashlight
+            if (DeviceInfo.DeviceType == DeviceType.Virtual && DeviceInfo.Platform == DevicePlatform.iOS)
+                return;
+#endif
```
mattleibow, line 22: Is this still true? The comments and versions say things but was this just copied or is it actually this case that needs it?

Xamarin.Essentials/HapticFeedback/HapticFeedback.netstandard.tizen.tvos.watchos.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Xamarin.Essentials
+{
+    public static partial class HapticFeedback
+    {
+        internal static bool IsSupported
+            => throw ExceptionUtils.NotSupportedOrImplementedException;
+
+        static async Task PlatformExecute(HapticFeedbackType type)
+        {
+            await Task.FromResult(0);
```
mattleibow, line 13: Can you remove the async and this dummy task?

Xamarin.Essentials/HapticFeedback/HapticFeedback.uwp.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Windows.Devices.Haptics;
+
+namespace Xamarin.Essentials
+{
+    public static partial class HapticFeedback
+    {
+        internal static bool IsSupported => true;
```
mattleibow, line 10: Maybe do a similar ApiInformation check like with the Vibration... I believe this API is only available in newer version of Windows. Might explode on older versions.

Xamarin.Essentials/HapticFeedback/HapticFeedback.android.cs
```diff
@@ -0,0 +1,36 @@
...
+using Android.Views;
+
+namespace Xamarin.Essentials
+{
+    public static partial class HapticFeedback
+    {
+        internal static bool IsSupported => true;
+
+        static async Task PlatformExecute(HapticFeedbackType type)
+        {
```
mattleibow, line 13: Should we check for the correct Android permissions here? Or is that not required for haptic feedback?

Xamarin.Essentials/HapticFeedback/HapticFeedback.android.cs
```diff
@@ -0,0 +1,36 @@
...
+            try
+            {
+                Platform.CurrentActivity?.Window?.DecorView?.PerformHapticFeedback(ConvertType(type));
+                await Task.FromResult(0);
```
mattleibow, line 17: Maybe use `Task.CompletedTask`. And no need to async. Just return it directly.

Xamarin.Essentials/HapticFeedback/HapticFeedback.ios.cs
```diff
@@ -0,0 +1,47 @@
...
+            switch (type)
+            {
+                case HapticFeedbackType.LongPress:
+                    PlatformLongPress();
+                    break;
+                default:
+                    PlatformClick();
+                    break;
+            }
+            await Task.FromResult(0);
```
mattleibow, line 22: Maybe use `Task.CompletedTask`. And no need to async. Just return it directly.

DeviceTests/DeviceTests.Shared/HapticFeedback_Tests.cs
dimonovdd: I just copied it from the vibration tests. These conditions are not necessary, I will delete them and we will check it.

Xamarin.Essentials/HapticFeedback/HapticFeedback.android.cs
dimonovdd: I added everywhere "await Task.FromResult(0)" only because "await"is used on UWP. Tell me how best to remove it.

Xamarin.Essentials/HapticFeedback/HapticFeedback.ios.cs
dimonovdd: I added everywhere "await Task.FromResult(0)" only because "await"is used on UWP. Tell me how best to remove it.

Xamarin.Essentials/HapticFeedback/HapticFeedback.uwp.cs
dimonovdd: Yes, I will add a check.

Xamarin.Essentials/HapticFeedback/HapticFeedback.netstandard.tizen.tvos.watchos.cs
dimonovdd: I added everywhere "await Task.FromResult(0)" only because "await"is used on UWP. Tell me how best to remove it.

Xamarin.Essentials/HapticFeedback/HapticFeedback.android.cs
dimonovdd: In the manifest file this line should be specified
`<uses-permission android:name="android.permission.VIBRATE"></uses-permission>`

Xamarin.Essentials/HapticFeedback/HapticFeedback.uwp.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        static async void PlatformExecute(HapticFeedbackType type)
+        {
+            try
+            {
+                if (Windows.Foundation.Metadata.ApiInformation.IsTypePresent(typeof(VibrationDevice).FullName) && await VibrationDevice.RequestAccessAsync() == VibrationAccessStatus.Allowed)
```
mattleibow, line 16: I think you have to use the string. If the type is not present, then you can't get the typeof.

Xamarin.Essentials/HapticFeedback/HapticFeedback.shared.cs
```diff
@@ -0,0 +1,14 @@
...
```
mattleibow, line 7: I think I like the `void` return here... 👍

rookiejava:
> if you are feeling adventurous, would you think about adding Tizen? Do what you can and we can get the Tizen folks to review: https://docs.tizen.org/application/dotnet/api/tizenfx/api/Tizen.System.Feedback.html
> https://github.com/Samsung/Tizen-CSharp-Samples/blob/b5e089032a9dc917ad5e6ca2232fef5d32f91b08/Mobile/FeedbackApp/FeedbackApp/FeedbackApp/MainPage.cs

Don't worry. we've got your back! 😉

Samples/Samples/View/HaptickFeadbackPage.xaml.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Samples.View
+{
+    public partial class HaptickFeadbackPage : BasePage
```
rookiejava, line 3:
```suggestion
    public partial class HaptickFeedbackPage : BasePage
```

typo :wink:  (here and there ...)

Xamarin.Essentials/HapticFeedback/HapticFeedback.uwp.cs
dimonovdd: ok

dimonovdd:
@mattleibow @rookiejava 
I have added the support for Tizen.

I think the name "Execute" is quite appropriate.
We can rename it "implement"

Samples/Samples/Samples.csproj
```diff
@@ -24,4 +24,10 @@
     <ProjectReference Include="..\..\Xamarin.Essentials\Xamarin.Essentials.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Update="View\HaptickFeedbackPage.xaml.cs">
+      <DependentUpon>HaptickFeedbackPage.xaml</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
```
mattleibow, line 32:
```suggestion
```

VS being funny again.

Samples/Samples/View/HaptickFeedbackPage.xaml.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Samples.View
+{
+    public partial class HaptickFeedbackPage : BasePage
```
mattleibow, line 3: Still a typo lols, "Haptic" (no "k")

dimonovdd:
@mattleibow @Redth @jamesmontemagno @rookiejava 


I apologize for the big amount of text

Now implemented are only the following types of HapticFeedback
| Platform      | Click                           | LongPress                    |
| ------------- | :-------------:                 | :-----:                    |
| Android       | ContextClick                    | LongPress                    |
| UwP           | Click                           | Press                        |
| iOS           | UIImpactFeedbackGenerator Light | UISelectionFeedbackGenerator |
| Tizen         | Tap                             | Hold                         |

But the platforms have many other different types of response. How do you think which of them we should add to xamarin essentials

**Android:**

- LongPress 
- KeyboardPress 
- KeyboardTap 
- ClockTick 
- ContextClick 
- KeyboardRelease 
- VirtualKeyRelease 
- TextHandleMove 

**UWP:**

- BuzzContinuous
- Click
- Press
- Release
- RumbleContinuous

**iOS:**

UIImpactFeedbackGenerator :
- Heavy
- Medium
- Light

UINotificationFeedbackGenerator :
- Success
- Warning
- Error

UISelectionFeedbackGenerator

**Tizen:**

- Tap
- SoftInputPanel
- Key0
- Key1
- Key2
- Key3
- Key4
- Key5
- Key6
- Key7
- Key8
- Key9
- KeyStar
- KeySharp
- KeyBack
- Hold
- HardwareKeyPressed
- HardwareKeyHold
- Message
- Email
- WakeUp
- Schedule
- Timer
- General
- PowerOn
- PowerOff
- ChargerConnected
- ChargingError
- FullyCharged
- LowBattery
- Lock
- UnLock
- VibrationModeAbled
- SilentModeDisabled
- BluetoothDeviceConnected
- BluetoothDeviceDisconnected
- ListReorder
- ListSlider
- VolumeKeyPressed

Samples/Samples/View/HapticFeedbackPage.xaml
```diff
@@ -0,0 +1,24 @@
...
+    </views:BasePage.BindingContext>
+
+    <StackLayout>
+        <Label Text="Quickly and easily make the device HapticFeedback." FontAttributes="Bold" Margin="12" />
+
+        <ScrollView>
+            <StackLayout Padding="12,0,12,12" Spacing="6">
+                <Button Text="Click" Command="{Binding ClickCommand}" />
+                <Button Text="LongPress" Command="{Binding LongPressCommand}" />
+                <Label Text="HaptickFeadback is not supported." TextColor="Red" FontAttributes="Italic"
```
mattleibow, line 18:
```suggestion
                <Label Text="HapticFeadback is not supported." TextColor="Red" FontAttributes="Italic"
```

Samples/Samples/View/HapticFeedbackPage.xaml
```diff
@@ -0,0 +1,24 @@
...
+                xmlns:views="clr-namespace:Samples.View"
+                xmlns:viewmodels="clr-namespace:Samples.ViewModel"
+                x:Class="Samples.View.HapticFeedbackPage"
+                Title="Vibration">
+    <views:BasePage.BindingContext>
+        <viewmodels:HapticFeedbackViewModel />
+    </views:BasePage.BindingContext>
+
+    <StackLayout>
+        <Label Text="Quickly and easily make the device HapticFeedback." FontAttributes="Bold" Margin="12" />
```
mattleibow, line 12:
```suggestion
        <Label Text="Quickly and easily make the device provide haptic feedback." FontAttributes="Bold" Margin="12" />
```

Samples/Samples/ViewModel/HomeViewModel.cs
```diff
@@ -204,6 +204,12 @@ public HomeViewModel()
                     typeof(VibrationPage),
                     "Quickly and easily make the device vibrate.",
                     new[] { "vibration", "vibrate", "hardware", "device" }),
+                new SampleItem(
+                    "📳",
+                    "HapticFeedback",
```
mattleibow, line 209:
```suggestion
                    "Haptic Feedback",
```

Samples/Samples/ViewModel/HomeViewModel.cs
```diff
@@ -204,6 +204,12 @@ public HomeViewModel()
...
+                    typeof(HapticFeedbackPage),
+                    "HapticFeedback",
+                    new[] { "HapticFeedback", "Haptic", "Feedback", "hardware", "device" }),
```
mattleibow, line 212:
```suggestion
                    new[] { "haptic", "feedback", "hardware", "device" }),
```

I think this is case sensitive...

Samples/Samples/ViewModel/HomeViewModel.cs
```diff
@@ -204,6 +204,12 @@ public HomeViewModel()
                     typeof(VibrationPage),
                     "Quickly and easily make the device vibrate.",
                     new[] { "vibration", "vibrate", "hardware", "device" }),
+                new SampleItem(
+                    "📳",
+                    "HapticFeedback",
+                    typeof(HapticFeedbackPage),
+                    "HapticFeedback",
```
mattleibow, line 211: The description from the page.

```suggestion
                    "Quickly and easily make the device provide haptic feedback",
```

mattleibow:
@dimonovdd 

> which of them we should add to xamarin essentials

I think for the first run, lets just keep the two. Once it is merged then we can decide. Unless @Redth or @jamesmontemagno has another opinion?

dimonovdd:
@mattleibow 
I agree.
I can create a issue on this topic.

Issue #1148, opened 2020-03-04
Error running OrientationSensor : System.ArgumentOutOfRangeException


Berets:
Bug report best practices: [Submitting Issues](https://github.com/xamarin/Essentials/blob/master/SUBMITTING-ISSUES.md)

### Description
System.ArgumentOutOfRangeException: 'Specified argument was out of the range of valid values.
Parameter name: index'

### Steps to Reproduce
1. Create an empty  XamarinForms project
2. Add every permission on Android project
3. In MainPage constructor add:

                       OrientationSensor.ReadingChanged += (sss, eee) =>
                        {
                            lblLatConv.Text = eee.Reading.Orientation.X.ToString();
                            lblLongConv.Text = eee.Reading.Orientation.Y.ToString();
                            lblAltConv.Text = eee.Reading.Orientation.Z.ToString();
                        };
                        
                        OrientationSensor.Start(SensorSpeed.Fastest);
Run on device, but also in simulator

### Expected Behavior
Return Quaternion
### Actual Behavior
System.ArgumentOutOfRangeException
### Basic Information

- Version with issue: 1.5.0
- Last known good version: Unknown
- IDE: Visual studio 2019
- Platform Target Frameworks:
  - iOS:  Not tested
  - Android: 28 
  - UWP:  Not tested
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

jamesmontemagno:
Hmmm interesting. Do you have the stack trace?

The only thing I could think is that the sensor isn't returning all 4, we should probably check the length of it.

Berets:
**System.ArgumentOutOfRangeException:** 'Specified argument was out of the range of valid values.
Parameter name: index'

03-06 10:04:12.082 D/Mono    (21502): DllImport attempting to load: '/system/lib64/liblog.so'.
03-06 10:04:12.085 D/Mono    (21502): DllImport loaded library '/system/lib64/liblog.so'.
03-06 10:04:12.085 D/Mono    (21502): DllImport searching in: '/system/lib64/liblog.so' ('/system/lib64/liblog.so').
03-06 10:04:12.085 D/Mono    (21502): Searching for '__android_log_print'.
03-06 10:04:12.085 D/Mono    (21502): Probing '__android_log_print'.
03-06 10:04:12.086 D/Mono    (21502): Found as '__android_log_print'.
03-06 10:04:12.102 I/MonoDroid(21502): UNHANDLED EXCEPTION:
[...]

PR #1150, opened 2020-03-05
GH-1148 Check array length on sensors on Android

jamesmontemagno:
### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #1148

### API Changes ###
None

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
+++ b/Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
@@ -45,6 +45,9 @@ namespace Xamarin.Essentials
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
         {
+            if (e?.Values?.Count < 3)
+                return;
+
             var data = new AccelerometerData(e.Values[0] / gravity, e.Values[1] / gravity, e.Values[2] / gravity);
             Accelerometer.OnChanged(data);
         }
--- a/Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
+++ b/Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
@@ -45,7 +45,7 @@ namespace Xamarin.Essentials
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
         {
-            if (e?.Values?.Count < 3)
+            if ((e?.Values?.Count ?? 0) < 3)
                 return;
 
             var data = new AccelerometerData(e.Values[0] / gravity, e.Values[1] / gravity, e.Values[2] / gravity);
--- a/Xamarin.Essentials/Barometer/Barometer.android.cs
+++ b/Xamarin.Essentials/Barometer/Barometer.android.cs
@@ -41,7 +41,7 @@ namespace Xamarin.Essentials
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
         {
-            if (e?.Values?.Count == 0)
+            if ((e?.Values?.Count ?? 0) == 0)
                 return;
 
             Barometer.OnChanged(new BarometerData(e.Values[0]));
--- a/Xamarin.Essentials/Barometer/Barometer.android.cs
+++ b/Xamarin.Essentials/Barometer/Barometer.android.cs
@@ -40,6 +40,11 @@ namespace Xamarin.Essentials
         }
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
-            => Barometer.OnChanged(new BarometerData(e.Values[0]));
+        {
+            if (e?.Values?.Count == 0)
+                return;
+
+            Barometer.OnChanged(new BarometerData(e.Values[0]));
+        }
     }
 }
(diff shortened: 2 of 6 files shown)


Mrnikbobjeff:
I don't think this is the optimal way to handle this (I specifically looked at the orientationsensor implementation, but I assume it holds for other sensors as well). The event includes the sensor firing it, which in turn has a property defining the type. There are some resources pointing out the structure of the data generated by the sensor. I would rather have a map detailing how many values to expect for a given sensor. As far as I can tell from the online resources we can expect that some phones will always have a different sensor value length, as they will always use a different sensor than the one we use which provides four values, see https://developer.android.com/guide/topics/sensors/sensors_position for more detail

jamesmontemagno:
Each sensor type has a specific number of items in the array. This is documented in the Android docs.

We mostly use motion sensors: https://developer.android.com/guide/topics/sensors/sensors_motion

For Orientation we specifically use: https://developer.android.com/reference/android/hardware/Sensor#TYPE_ROTATION_VECTOR

I could put a const int in there to specify it. Basically no sensor should ever return invalid data, i have never seen it, however apparently some devices do so we shouldn't throw an exception and crash the app, instead we should check the lengths.

Mrnikbobjeff:
I can tell from personal experience when developing a game which was primarily motion sensor focused that some devices disregard the requested sensor type in low battery mode. This is not documented anywhere and probably a mistake in the vendor os. I guess that this is also the error occuring in the ticket which prompted this pr. In the case I recall working on we had a sensor returning a type_orientation sensor which has been deprecated. On another fringe device we had a ROTATION_VECTOR sensor on an API level > 18 and it was still missing the fourth value which is supposed to be present. This was also the reason I commented, while checking might work it would also expose us to errors, e.g. I believe you have to at least update the ROTATION_VECTOR code as it is supposed to have 5 values 

> (values[3], originally optional, will always be present from SDK Level 18 onwards. values[4] is a new value that has been added in SDK Level 18.)

This is why I asked for a seperate check with a map or similar :)

Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
```diff
@@ -45,6 +45,9 @@ void ISensorEventListener.OnAccuracyChanged(Sensor sensor, [GeneratedEnum] Senso
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
         {
+            if (e?.Values?.Count < 3)
```
mattleibow: Is null < 3? If we are sure we will get 3 values, then maybe != 3?

Xamarin.Essentials/Barometer/Barometer.android.cs
```diff
@@ -40,6 +40,11 @@ void ISensorEventListener.OnAccuracyChanged(Sensor sensor, [GeneratedEnum] Senso
         }
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
-            => Barometer.OnChanged(new BarometerData(e.Values[0]));
+        {
+            if (e?.Values?.Count == 0)
```
mattleibow: I think we should check for != 1 because we may have null.

Xamarin.Essentials/Accelerometer/Accelerometer.android.cs
jamesmontemagno: Doesn't null return the default? Which would be 0? Could test this out ;)

jamesmontemagno:
I did checks against the null checks and yes you are right :( I can modify this though to return 0 let me do that.

I think these tests are fine on all sensors. On orientation like you mentioned @Mrnikbobjeff it could return 3 values so we could check for those and return -1 for w if it doesn't exist.

Xamarin.Essentials/Barometer/Barometer.android.cs
```diff
@@ -40,6 +40,11 @@ void ISensorEventListener.OnAccuracyChanged(Sensor sensor, [GeneratedEnum] Senso
         }
 
         void ISensorEventListener.OnSensorChanged(SensorEvent e)
-            => Barometer.OnChanged(new BarometerData(e.Values[0]));
+        {
+            if ((e?.Values?.Count ?? 0) == 0)
```
Redth: I'm just paranoid but... 

```suggestion
            if ((e?.Values?.Count ?? 0) <= 0)
```

Xamarin.Essentials/Compass/Compass.android.cs
```diff
@@ -78,6 +78,10 @@ void ISensorEventListener.OnSensorChanged(SensorEvent e)
             {
                 SensorManager.GetRotationMatrix(r, null, lastAccelerometer, lastMagnetometer);
                 SensorManager.GetOrientation(r, orientation);
+
+                if (orientation.Length == 0)
```
Redth: More paranoia!

```suggestion
                if (orientation.Length == 0)
```

PR #1162, opened 2020-03-10
Don't request location permission if disabled.

### LINKED ISSUE #1161: [Bug] Permissions.RequestAsync
### Description

**The issue is only present on iOS.**

On iOS, requesting the permission `Permissions.LocationWhenInUse` while location service is disabled, the system dialog asking the user to open the settings close by itself and do nothing. 

### Steps to Reproduce

The code for requesting the permission :

```cs
var permissionStatus = await Permissions.CheckStatusAsync<Permissions.LocationWhenInUse>();
if (permissionStatus != PermissionStatus.Granted)
{
    var requestResult = await Permissions.RequestAsync<Permissions.LocationWhenInUse>();
    if (requestResult != PermissionStatus.Granted)
        throw new PermissionException($"Permission '{typeof(Permissions.LocationWhenInUse).Name}' request to the user failed with status '{requestResult}'");
}
```

1. Disable the location service on the device in Settings > Privacy > Location Services >Switch off Location Services
2. Call the code I wrote above in your app (or use the repro app)
3. Enjoy the issue 😄 

### Expected Behavior

The dialog should not close by itself I suppose? And clicking on "Settings" button should open the settings.

### Actual Behavior

The dialog close itself randomly, and if I click on "Settings" button, it just close and nothing happen.

### Basic Information

- Version with issue: 1.5.1
- Last known good version:
- IDE: Visual Studio for Mac 8.4.8
- Platform Target Frameworks: 
  - iOS:  latest (13.2 I guess ?)
  - Android: Seems OK.
  - UWP: Not tested
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Screenshots

See attached .zip for video of the issue.

### Reproduction Link

The following zip contains a video of the issue and a repro solution.
(I'm sorry, I forgot to switch the emulator language to English)

[RequestLocationBugRepro.zip](https://github.com/xamarin/Essentials/files/4310872/RequestLocationBugRepro.zip)

jamesmontemagno:
Currently, we prompt for permission is it is disabled, but we shouldn't and just return that state.

Fixes #1161

### DIFF
--- a/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
@@ -100,7 +100,7 @@ namespace Xamarin.Essentials
                 EnsureDeclared();
 
                 var status = GetLocationStatus(true);
-                if (status == PermissionStatus.Granted)
+                if (status == PermissionStatus.Granted || status == PermissionStatus.Disabled)
                     return status;
 
                 EnsureMainThread();


jamesmontemagno:
One thing I can think is that we are passing in an action like: lm => lm.RequestAlwaysAuthorization() so maybe ti loses the context.... let me try something here.

jamesmontemagno:
So, NM we are good here. what happens is that we get back a denied state and we dispose of the location manager because the task is done. This is the best we can do IMHO.

Issue #1180, opened 2020-03-18
[Bug] [Android] Browser.OpenAsync creates new task


rickclephas:
### Description

The implementation of `Browser.OpenAsync` is always starting a new task for the CustomTabs:
https://github.com/xamarin/Essentials/blob/9e74d1ce176233b87ff5e9a86a363e49da43357d/Xamarin.Essentials/Browser/Browser.android.cs#L30

This means the users will see the CustomTab as a separate app in their recent apps list and won't return to the app when they switch between apps.

### Steps to Reproduce

1. Call `Browser.OpenAsync` with `BrowserLaunchMode.SystemPreferred`
2. When the CustomTab is open switch to another app (without closing the CustomTab) 
3. Now switch back to the CustomTab via the recent apps. 
    You will notice that the CustomTab and the test app are both in the list. 
    But we'll select the CustomTab since that is where we left off.
4. Close the CustomTab

### Expected Behavior

After closing the CustomTab in step 4 you should return to the test app.

### Actual Behavior

After closing the CustomTab in step 4 you return to the home screen.
It looks like the app crashed but it is actually just in the background in another task.

### Basic Information

- Version with issue: 1.5.1
- Platform Target Frameworks:
  - Android: 10
- Affected Devices:
  - Google Pixel 3a XL
  - Samsung Galaxy S20+
  - _Presumably every other device as well_

rickclephas:
The `NewTask` flag was added in #228 to fix #225 .
Opening the CustomTab (or browser) from within the app should be within the same taks to preserve the flow. A new task should only be started if we don't have a running activity.

rickclephas:
When this is fixed it would probably also fix #798.

jamesmontemagno:
Hmmmmm yes, I agree with you here as I never did this before... Curious why this was put in...

Here is my old code: https://github.com/jguertl/SharePlugin/blob/master/Share/Share.Plugin.Android/ShareImplementation.cs#L41

jamesmontemagno:
Although I guess what would happen if you are on a dual screen device and  you want it to launch on the second screen... that flag would need to be there....

rickclephas:
> Although I guess what would happen if you are on a dual screen device and you want it to launch on the second screen... that flag would need to be there....

The flag would indeed be required to launch  it on a second screen (or in multi window mode).
Though I think this should be up to the developer of the app. 
In most situations I guess you would want it to be the same task (looking at other apps like Twitter, LinkedIn and Github).

[...]

Issue #225, opened 2018-05-04
Browser API is broken on Android


mattleibow:
### Description

Opening a chrome tab fails with an exception:

> Calling StartActivity() from outside of an Activity context requires the FLAG_ACTIVITY_NEW_TASK flag.

PR #1183, opened 2020-03-19
GH-1180 ; Don't launch new task on custom tabs.

jamesmontemagno:
### Description of Change ###

Based off feedback of launching new task for system. Can't find google docs showing to do it how we have it and my old plugin didn't have it.

### Bugs Fixed ###

- Related to issue #1180

### Behavioral Changes ###

Will now launch in same activity.

### PR Checklist ###

### DIFF
--- a/Samples/Samples/View/BrowserPage.xaml
+++ b/Samples/Samples/View/BrowserPage.xaml
@@ -53,8 +53,8 @@
                     <Label Text="Launch Adjacent (Android)" Grid.Column="0" Grid.Row="0" />
                     <Switch IsToggled="{Binding LaunchAdjacent, Mode=TwoWay}" Grid.Column="1" Grid.Row="0" />
 
-                    <Label Text="Present Modal (iOS)" Grid.Column="0" Grid.Row="1" />
-                    <Switch IsToggled="{Binding PresentModal, Mode=TwoWay}" Grid.Column="1" Grid.Row="1" />
+                    <Label Text="Present as Form Sheet (iOS)" Grid.Column="0" Grid.Row="1" />
+                    <Switch IsToggled="{Binding PresentAsFormSheet, Mode=TwoWay}" Grid.Column="1" Grid.Row="1" />
 
                     <Label Text="Present as Page Sheet (iOS)" Grid.Column="0" Grid.Row="2" />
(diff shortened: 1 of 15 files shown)


Redth:
The original issue #225 was basically:

> Calling StartActivity() from outside of an Activity context requires the FLAG_ACTIVITY_NEW_TASK flag.

Which makes me think we should maybe check if we have the Activity's context first, and if not, still add the NewTask flag because it's all we can do at that point.

jamesmontemagno:
I think we want to add back in:

#if __ANDROID_24__	
                    if (Platform.HasApiLevelN)	
                        flags |= ActivityFlags.LaunchAdjacent;	
#endif

rickclephas:
> I think we want to add back in:
> 
> #if __ANDROID_24__	
>                     if (Platform.HasApiLevelN)	
>                         flags |= ActivityFlags.LaunchAdjacent;	
> #endif 
> 
> 
> 

Good point but the `LaunchAdjacent` flag requires the `NewTask` flag. See [FLAG_ACTIVITY_LAUNCH_ADJACENT](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_LAUNCH_ADJACENT).

IMHO multi-window support should be an option in `BrowserLaunchOptions`.

Redth:
So, I've added to this PR a new `BrowserLaunchFlags` and property on `BrowserLaunchOptions`.  My thinking here is we can use this to set some platform specific things like Launching adjacent on android or presenting as Modal on iOS.  Things that are useful to expose but don't really have a cross platform meaning.

This takes care of adding optionally adding LAUNCH_ADJACENT flag for Android where we otherwise wouldn't want NewTask.  Thoughts?

rickclephas:
> So, I've added to this PR a new `BrowserLaunchFlags` and property on `BrowserLaunchOptions`.  My thinking here is we can use this to set some platform specific things like Launching adjacent on android or presenting as Modal on iOS.  Things that are useful to expose but don't really have a cross platform meaning.
> 
> This takes care of adding optionally adding LAUNCH_ADJACENT flag for Android where we otherwise wouldn't want NewTask.  Thoughts?

Sounds good to me! 🙂

jamesmontemagno:
Just running final test and should be good

PR #1184, opened 2020-03-19
Cache Android SdkInt at runtime

Redth:
This should never ever change at runtime, so should be safe to cache.

### DIFF
--- a/Xamarin.Essentials/Platform/Platform.android.cs
+++ b/Xamarin.Essentials/Platform/Platform.android.cs
@@ -180,8 +180,15 @@ namespace Xamarin.Essentials
             false;
 #endif
 
+        static int? sdkInt;
+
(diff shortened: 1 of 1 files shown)


Xamarin.Essentials/Platform/Platform.android.cs
```diff
@@ -180,8 +180,13 @@ internal static AndroidUri GetShareableFileUri(string filename)
             false;
 #endif
 
+        static int? sdkInt;
+
+        internal static int SdkInt
+            => sdkInt ?? (sdkInt = (int)Build.VERSION.SdkInt).Value;
```
mattleibow:
```suggestion
            => sdkInt ??= (int)Build.VERSION.SdkInt;
```

Gotta use the new things!

PR #1185, opened 2020-03-20
[Feature] Event for Requested Theme Changes

jamesmontemagno:
**NEEDS REVIEW**

### Description of Change ###

Enable developers to subscribe to app theme request changes. Also add in platform helpers on iOS/Android to bubble up platform events.

### API Changes ###
App Info:
* Added: public static EventHandler<AppThemeChangedEventArgs> RequestedThemeChanged
* Added: AppThemeChangedEventArgs

In Platform files:
* Added on iOS & tvOS: public static void TraitCollectionDidChange(UITraitCollection currentUITraitCollection)
* Added on Android:  public static void OnConfigurationChanged(Configuration newConfig)

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/AppInfo/AppInfo.uwp.cs
+++ b/Xamarin.Essentials/AppInfo/AppInfo.uwp.cs
@@ -1,11 +1,17 @@
 ﻿using System.Globalization;
 using Windows.ApplicationModel;
+using Windows.UI;
+using Windows.UI.ViewManagement;
 using Windows.UI.Xaml;
 
 namespace Xamarin.Essentials
 {
     public static partial class AppInfo
     {
+        static UISettings uiSetting;
+
+        static UISettings UISettings => uiSetting ??= new UISettings();
+
         static string PlatformGetPackageName() => Package.Current.Id.Name;
 
         static string PlatformGetName() => Package.Current.DisplayName;
@@ -24,5 +30,19 @@ namespace Xamarin.Essentials
 
         static AppTheme PlatformRequestedTheme() =>
             Application.Current.RequestedTheme == ApplicationTheme.Dark ? AppTheme.Dark : AppTheme.Light;
+
+        static void StartThemeListeners() =>
+            UISettings.ColorValuesChanged += UISettingsColorValuesChanged;
+
+        static void StopThemeListeners() =>
+            UISettings.ColorValuesChanged -= UISettingsColorValuesChanged;
+
+        static void UISettingsColorValuesChanged(UISettings sender, object args)
+        {
+            MainThread.BeginInvokeOnMainThread(() =>
+            {
+                OnAppThemeChanged(PlatformRequestedTheme());
+            });
+        }
     }
 }
--- a/Samples/Samples.Android/MainActivity.cs
+++ b/Samples/Samples.Android/MainActivity.cs
@@ -1,13 +1,14 @@
 ﻿using Android.App;
 using Android.Content;
 using Android.Content.PM;
+using Android.Content.Res;
 using Android.OS;
 using Android.Runtime;
 using Android.Widget;
 
 namespace Samples.Droid
 {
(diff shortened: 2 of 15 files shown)


jamesmontemagno:
I would also like @PureWeen to take a look at this or any other Xamarin.Forms teams members.

My thought is that Xamarin.Forms could automatically register for the underlying events so the developer does need to do it.

PureWeen:
@jfversluis

jfversluis:
So, this is basically how I have implemented it at some point. My consideration for Android was this. For the Android (auto) dark theme colors, to kick in you will need to recreate the activity. If you do not handle the ConfigurationChanged event this will happen automatically, if you handle the ConfigurationChanged like you are here then you will need to trigger the recreation yourself. By doing that though, you _can_ fire your own event, but that will have very limited use because the activity is recreated right after you did whatever you want to do in our event, right? Does that make sense?

There are of course other use cases and I guess this then makes sense and Forms can hook into this. So from that standpoint this looks fine :)

jamesmontemagno:
So, if you have the default:

```
[Activity(
    // other settings
    ConfigurationChanges = ConfigChanges.ScreenSize | ConfigChanges.Orientation)]
```

Then your application activity is killed and re-inflated, which is really bad and potentially will crash your application. 

If you add
```
| ConfigChanges.UiMode)]
```
Then your activity will remain, your state will remain and you can update the UI. This is how all templates should be setup. 

So, the thought here is that a "FormssAppCompatAcitivty" could forward the event to Essentials to automatically trigger the event handler. 

This means that all a user would have to do is add:
```
| ConfigChanges.UiMode)]
```
and not have to override the event themselves. 

Ideally Xamarin.Forms could also Init essentials, do the permissions override too :) but that is for another conversation.

jfversluis:
Totally on board with how this works, but for the Android OS theme to actually flip in your app when you do the `| ConfigChanges.UiMode)]`, you will now need to manually call `Recreate();`, that is what I understand from it anyway. Hopefully I'm missing something obvious here.

So, if that's something you are ignoring as a developer and do all the theming yourself then that is fine. But if you want to lean into the automatic theme done by the DayNight theme you will still have to recreate the activity. See attached project and screen recording to hopefully get my point across. 

Maybe I'm focussing too much on a use case that is not very common or we are willing to document and get over with :)

[AndroidConfigTest.zip](https://github.com/xamarin/Essentials/files/4372045/AndroidConfigTest.zip)

https://1drv.ms/v/s!Anrk6U9nRknzg58bMYOsBj6dOqcalQ?e=6aUL73

jamesmontemagno:
I mean I just "apply theme" when this happens: https://github.com/jamesmontemagno/Hanselman.Forms

All my styles are dynamic. I think we can document that you may need to re-create, but ideally your app just stays in the same place and not re-boot the entire activity. We can hop on a call to discuss though :)

I think Android will just update the UI if you are using the correct theme: https://developer.android.com/guide/topics/ui/look-and-feel/darktheme ?

jfversluis:
Yeah, so, I guess that makes sense. Most likely scenarios are; you take full control over your theming and use this. Or use the automatic theming and do not worry about the event and just let the recreate do its thing.

Ultimately it just comes down to documenting this well and let people make their own choice.

Xamarin.Essentials/AppInfo/AppInfo.uwp.cs
```diff
@@ -24,5 +30,19 @@ static string PlatformGetVersionString()
...
+
+        static void StartThemeListeners() =>
+            UISettings.ColorValuesChanged += UISettingsColorValuesChanged;
+
+        static void StopThemeListeners() =>
+            UISettings.ColorValuesChanged -= UISettingsColorValuesChanged;
+
+        static void UISettingsColorValuesChanged(UISettings sender, object args)
+        {
+            MainThread.BeginInvokeOnMainThread(() =>
```
Mrnikbobjeff, line 42: Is this needed? We already switch to the mainthread in the OnAppThemeChanged

mattleibow:
I am closing this not because this is not good to have, but with the new way with MAUI, this is better served in that project. Also, with the merging of things, this is better to be re-investigated in the new world.

Issue #148, opened 2018-04-03
App Actions (Quick Actions/App Shortcuts)


jamesmontemagno:
### Description

API for iOS's [Quick Actions](https://developer.apple.com/ios/human-interface-guidelines/extensions/home-screen-actions) and Android's [App Shortcuts](https://developer.android.com/guide/topics/ui/shortcuts.html)

API:
```
event OnAppAction(string actionType);
IEnumerable<AppActionItem> ActionItems {get;set;}
```

AppActionItem
```
string ActionType {get; set;}
string LocalizedTitle {get;set;}
string Icon {get;set;} 
```

On iOS we should pull from xcassets or resources and drawables on Android.

Event must be registered early enough is application lifecycle to handle events. (Application or AppDelegate).

JakeStaTeresa:
@jamesmontemagno App Shortcuts link points to the same link as Quick Actions. It should point to https://developer.android.com/guide/topics/ui/shortcuts.html

Redth:
Community contributions are welcome on this one!

edsnider:
I have the iOS and Android code for this and can contribute... just need to think through the `OnAppAction` event. The way the "actions" come into the application would require the user to wire up something that could then invoke the `OnAppAction`. Doesn't feel very clean though.  Looks like @Redth is also dealing with this in the Local Notifications. Thoughts?

newky2k:
@edsnider  did you get anywhere with this?

edsnider:
@newky2k I was waiting for it to come back out of “Needs triage” before I pushed anything.

heytherewill:
A think a good approach for this API is to have only two methods:

`Task SetShortcuts(Shortcut[] shortcuts)`
This will replace all existing shortcuts with the ones provided in the array (or whichever collection type we settle to use). 

`Task ClearShortcuts()`
Shorthand for `SetShortcuts(new Shortcut[0])`

This much API surface should cover most/all needs of developers and is really straight forward to both implement and maintain. I recommend using `Task` over void because iOS requires Quick Actions to be set on the main thread.

[...]

PR #1190, opened 2020-03-26
[WIP] GH-148: App Actions

edsnider:
### Description of Change ###

This PR adds App Actions API.

### Bugs Fixed ###

- Related to issue #148 

### API Changes ###

Added: 
 
- `ShortcutManager Platform.ShortcutManager { get; }` to `Platform.android.cs`

- New classes:
```csharp
public static partial class AppActions
{
    public static IEnumerable<AppAction> Actions
    {
        get => PlatformGetActions();
        set => PlatformSetActions(value);
    }
}
```
```csharp
public class AppAction
{
    public AppAction(string actionType, string title, string subtitle = null, string icon = null, Uri uri = null)
    {
        ActionType = actionType;
        Title = title;
        Subtitle = subtitle;
        Icon = icon;
        Uri = uri;
    }

    public string ActionType { get; set; }
    public string Title { get; set; }
    public string Subtitle { get; set; }
    public Uri Uri { get; set; }
    internal string Icon { get; set; }
}
```

### Behavioral Changes ###

None

### PR Checklist ###


edsnider:
Will commit samples and tests next, but wanted to get initial feedback on the implementation.

Redth:
I have a couple of questions / thoughts:

I think we can just name the title properties without the 'Localized' part of the name (e.g. `Title`, and `Subtitle`.

I'm a bit torn on the Icon... I don't love that the property exists on the result when we can't reasonably get the value back when we ask for the existing actions...  I'm almost wondering if the AppAction type should take the icon as a ctor parameter but not expose the value as a readable property.  This would make it very intentional that you can set the Icon when creating the AppAction but that there is no expectation you can read what the icon is when asking for the existing app actions.

edsnider:
I'll remove `Localized` from the title and subtitle properties.

As far as the `Icon` property not being available when "getting".... I like your constructor argument idea. The other thought I had about it was... do we even need to "get" existing actions? Could the API just be the ability to set them?

Redth:
I had the same thought of if we really needed to be able to "Get" them or not, but I think it's a pretty simple bit of code you already have implemented to do so, we just need to deal with the icon.

Another thought around getting actions, will getting them using the API's you've built also return actions declared in the info.plist on iOS and/or the android resource files?  If they do return those, that makes a better case for implementing the 'Get' functionality.

edsnider:
No, the static shortcut items are not returned. On iOS they aren't returned by the API. On Android, they are available via a separate method (`ManifestShortcuts` vs `DynamicShortcuts`) but I'm only returning the dynamic shortcuts for consistency.

Redth:
Ok, I think it would still be nice to have the API return the items even if we have to make the icon available as part of the ctor only.  I think it's a wise choice to not use`ManifestShortcuts` on android if there's no iOS equivalent.

If you can make that change I think this will be ready to go :)

Redth:
@edsnider this looks good, just needs some tests and docs.  Let us know if you need help with those!

edsnider:
Awesome. I've got tests locally... was waiting until we had the model changes locked down before committing those. Will get those committed. Do you want me to update the Sample app as well?

Redth:
If you’re able and willing yes please!

edsnider:
I'm running into build issues related to `ShouldShowRationale()`.

edsnider:
@Redth thanks, I'm able to build again... I'll get the docs and sample committed as soon as possible.

MartinZikmund:
Can you add UWP support too? The `JumpList` API represents this functionality :-) .

Xamarin.Essentials/AppActions/AppActions.uwp.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (!string.IsNullOrEmpty(action.Subtitle))
+                item.Description = action.Subtitle;
+
+            if (!string.IsNullOrEmpty(action.Icon))
+            {
+                var dir = IconDirectory.Trim('/', '\\').Replace('\\', '/');
+
+                var ext = IconExtension.TrimStart('.');
+
+                item.Logo = new Uri($"ms-appx:///{dir}/{action.Icon}.{ext}");
```
mattleibow, line 95:
```suggestion
                var ext = IconExtension;
                if (!ext.StartsWith("."))
                    ext = "." + ext;

                item.Logo = new Uri($"ms-appx:///{dir}/{action.Icon}{ext}");
```

PR #1196, opened 2020-03-27
GeoHash Support for Xamarin Essentials

sps014:
### Description of Change ###
Added support for Encoding and Decoding GeoHashes

### GeoHashing
Easy to use C# API for GeoHashing ✔️.
This is the foundation for Geofencing and proximity searching.
The algorithm is based on [Gustavo Niemeyer’s geocoding system](https://en.wikipedia.org/wiki/Geohash)

### Usage
```C#

            var s = GeoHash.Encode(26.83710, 80.92060, 12);
            ///eg output="tuxmgj12sdg1"
            var sp = GeoHash.Encode(new GeoPoint(26.83710, 80.92060), 6);
             ///eg output="tuxmgj"

            var d = GeoHash.Decode(s);
            //d.Latitude=26.8371
            //d.Longitude=80.9205
            
            var n1 = GeoHash.GetAdjacent(s,CardinalDirection.North);
            ///eg output="tuxmgl"
            var n4 = GeoHash.GetAdjacent(s, CardinalDirection.East);
            ///eg output="tuxmgg" and so on for rest two
            var n2 = GeoHash.GetAdjacent(s, CardinalDirection.South);
            var n3 = GeoHash.GetAdjacent(s, CardinalDirection.West);
            
            var ss= GeoHash.Neighbours(s);
            //ss.North="tuxmgl"
            //ss.Southeast="tuxmg1"
           // and so on
```

### PR Checklist ###

- [✔️] Has tests (if omitted, state reason in the description)
- [✔️] Rebased on top of master at time of PR
- [✔️] Changes adhere to the coding standard

### DIFF
--- a/DeviceTests/DeviceTests.Shared/DeviceTests.Shared.csproj
+++ b/DeviceTests/DeviceTests.Shared/DeviceTests.Shared.csproj
@@ -31,7 +31,7 @@
     <ProjectReference Include="..\..\Xamarin.Essentials\Xamarin.Essentials.csproj" />
   </ItemGroup>
   <ItemGroup Condition=" $(TargetFramework.StartsWith('uap10.0')) ">
-    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="6.1.9" />
+    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="6.2.9" />
     <SDKReference Include="WindowsMobile, Version=10.0.16299.0">
       <Name>Windows Mobile Extensions for the UWP</Name>
     </SDKReference>
--- a/DeviceTests/DeviceTests.UWP/DeviceTests.UWP.csproj
+++ b/DeviceTests/DeviceTests.UWP/DeviceTests.UWP.csproj
@@ -115,7 +115,7 @@
   <ItemGroup>
     <PackageReference Include="UnitTests.HeadlessRunner" Version="2.0.0" />
     <PackageReference Include="Xamarin.Forms" Version="4.2.0.848062" />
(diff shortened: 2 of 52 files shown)


sps014:
These two are failing because of following non-related this module related errors.
Since I am only using DotNetCore Code with no Requirement of GPS (Native Device Features) or etc.
error
Any Help will be appreciated.
Thank you.

jamesmontemagno:
This looks super good to me. I am not a grammar person, but my only thought is it would be:

* Southwest and not SouthWest
* Northeast and not NorthEast
*etc.

Instead of "Direction" I may say "CardinalDirection"

I would maybe use "GeoHashNeighbor"

For the structs they should be readonly and IEquatable such as: https://github.com/xamarin/Essentials/blob/develop/Xamarin.Essentials/Barometer/Barometer.shared.cs#L76


For "CellSize" does it need to be public? Same with BoxSizeTable?  Doesn't look like either are used anywhere

With Bound I would maybe also say "GeoHashBound" to be more specific.


I think that would be my API review. @redth | @mattleibow

sps014:
@jamesmontemagno Really thankful to you for your awesome feedback,
I am a bit poor when it comes to naming things 😁.
I have updated most of the things as you suggested and also renamed some methods like ```Adjacent``` to ```GetAdjacent```  so to make the method more expressive by their name.
It was a great learning experience for me.
Thank you.

jamesmontemagno:
Had an API review today and discussions with the team. And while I do think that this PR is awesome, it doesn't fit in with the theme of what Xamarin.Essentials is, which is abstracting platform specific capabilities. GeoHashing better fits in a separate cross-platform library that is just .NET standard for every developer to use. There are a few out there that do this today. Thank you for your work on this though and going back and forth.

sps014:
Thanks will love to further contribute.

Issue #1177, opened 2020-03-17
Location.Altitude not consistent between Android and iOS


janusw:
### Description

This is an issue that I have been wondering about for some time already (and by now I think I have some idea how the behavior comes about) ...

I observe strongly differing heights being returned in Location.Altitude on Android and iOS:

* On iOS I obtain what seems to be the correct altitude above mean sea level (MSL). If I stand on the ground, it agrees with the elevation obtained via the Google Elevation API.
* On Android I get height values that are roughly 50m above the iOS values. The difference is not exactly 50m, of course, but it always seems to be in the same order of magnitude and larger than the reported accuracies.

### Steps to Reproduce

1. Open the Xamarin.Essentials app on an Android and an iOS device.
2. Select the 'Geolocation' page.
3. Hit 'Refresh'.
4. Observe the Altitude.
5. Possibly compare to the altitude reported by other apps or devices.

### Expected Behavior

The altitude values obtained on Android and iOS should be compatible with each other (within accuracy), and in particular they should be given with respect to the same reference level.

### Actual Behavior

The altitudes are not compatible. I suspect that different reference points are used. The difference is roughly the size of the undulation at my location (central Europe) AFAICT.

### Basic Information

- Version with issue: 1.4.0, 1.5.0, 1.5.1 
- Last known good version: ---
- IDE: VS for Mac
- Platform Target Frameworks:
  - iOS:  8.0
  - Android: 10.0
  - UWP:  ---
- Android Support Library Version: ?
- Nuget Packages: ?
- Affected Devices: tried e.g. on Zenfone 6, iPhone 8

janusw:
Checking the documentation of the relevant API calls ...

* [Xamarin](https://docs.microsoft.com/de-de/dotnet/api/xamarin.essentials.location.altitude?view=xamarin-essentials#Xamarin_Essentials_Location_Altitude): `Gets the Altitude, if available in meters above sea level.`
* [Android](https://developer.android.com/reference/android/location/Location#getAltitude()): `Get the altitude if available, in meters above the WGS 84 reference ellipsoid. If this location does not have an altitude then 0.0 is returned.`
* [iOS](https://developer.apple.com/documentation/corelocation/cllocation/1423820-altitude): `The altitude, measured in meters. Positive values indicate altitudes above sea level. Negative values indicate altitudes below sea level.`

[...]

PR #1197, opened 2020-03-29
add new property Location.AltitudeReferenceSystem

janusw:
### Description of Change ###

In order to deal with platform differences of Location.Altitude, this PR adds a new property Location.AltitudeReferenceSystem and a corresponding enum, which is a copy of [Windows.Devices.Geolocation.AltitudeReferenceSystem](https://docs.microsoft.com/en-us/uwp/api/windows.devices.geolocation.altitudereferencesystem).

### Bugs Fixed ###

- Related to issue #1177

### API Changes ###

Added: 

- `enum AltitudeReferenceSystem`
- `AltitudeReferenceSystem Location.AltitudeReferenceSystem { get; set; }`

Changed:

 - None.

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Types/LocationExtensions.uwp.cs
+++ b/Xamarin.Essentials/Types/LocationExtensions.uwp.cs
@@ -8,13 +8,21 @@ namespace Xamarin.Essentials
 {
     public static partial class LocationExtensions
     {
+        internal static AltitudeReferenceSystem ConvertRefSys(Windows.Devices.Geolocation.AltitudeReferenceSystem refsys)
+        {
+            if (Enum.IsDefined(typeof(AltitudeReferenceSystem), refsys))
+                return (AltitudeReferenceSystem)refsys;
+            else
+                return AltitudeReferenceSystem.Unspecified;
+        }
+
         internal static Location ToLocation(this MapLocation mapLocation) =>
             new Location
             {
                 Latitude = mapLocation.Point.Position.Latitude,
                 Longitude = mapLocation.Point.Position.Longitude,
                 Altitude = mapLocation.Point.Position.Altitude,
-                AltitudeReferenceSystem = (AltitudeReferenceSystem)mapLocation.Point.AltitudeReferenceSystem,
+                AltitudeReferenceSystem = ConvertRefSys(mapLocation.Point.AltitudeReferenceSystem),
                 Timestamp = DateTimeOffset.UtcNow
             };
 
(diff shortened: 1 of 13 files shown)


Xamarin.Essentials/Types/LocationExtensions.uwp.cs
```diff
@@ -14,6 +14,7 @@ public static partial class LocationExtensions
                 Latitude = mapLocation.Point.Position.Latitude,
                 Longitude = mapLocation.Point.Position.Longitude,
                 Altitude = mapLocation.Point.Position.Altitude,
+                AltitudeReferenceSystem = mapLocation.Point.AltitudeReferenceSystem,
```
jamesmontemagno: You will need some type of converstion system. Such as https://github.com/xamarin/Essentials/blob/develop/Xamarin.Essentials/Types/SensorSpeed.uwp.cs

janusw: Well, the enum `Xamarin.Essentials.AltitudeReferenceSystem` is an exact copy of `Windows.Devices.Geolocation.AltitudeReferenceSystem`, using the same enum values, therefore conversion should be necessary, right? I even thought about directly reusing the enum in XE without a copy, but since it's only available on Windows, I did not see a way to do that for all platforms.

janusw: Ah, I see the build failure now. So yes, seems I'm missing a cast there (I did not get that error when I built in VS).

janusw: Fixed in 1610891.

jamesmontemagno:
Instead of a hard case I would maybe to a little method that does a switch expression over it. Besides that it looks good.

janusw:
> Instead of a hard case I would maybe to a little method that does a switch expression over it. 

Well, in the current state that is certainly not necessary. What you worry about is probably that `Windows.Devices.Geolocation.AltitudeReferenceSystem` might get extended with additional values in the future? For that case it might be simpler to check with `Enum.IsDefined`?

janusw:
> > Instead of a hard case I would maybe to a little method that does a switch expression over it.
> 
> Well, in the current state that is certainly not necessary. What you worry about is probably that `Windows.Devices.Geolocation.AltitudeReferenceSystem` might get extended with additional values in the future? For that case it might be simpler to check with `Enum.IsDefined`?

I implemented that in c4c4e90. What do you think about it, @jamesmontemagno?

jamesmontemagno:
Did a little modification on what I was talking about. Should be good now.

janusw:
> Did a little modification on what I was talking about. Should be good now.

Agree, looks good. Thank you!

janusw:
Ping! @jamesmontemagno, can this be merged?

Issue #1207, opened 2020-04-03
[Bug] [iOS] TextToSpeech crash in Release


ilyalehchylin:
### Description
App crashes in release (TestFlight) when using TextToSpeech (there is everything alright in debug though). I logged the exception and it throws the following message:

> The lazily-initialized type does not have a public, parameterless constructor.

### Steps to Reproduce

Call `TextToSpeech.SpeakAsync` (release mode)

### Expected Behavior

Voice is starting to pronounce text.

### Actual Behavior

App crashes.

### Basic Information

- Version with issue: 1.5.1, 1.5.2
- Last known good version: N/A
- IDE: Visual Studio for Mac Community 8.5.1 (build 42)
- Platform Target Frameworks: 
  - iOS: 
    - Deployment target: 8.0
    - Tested device version: 13.3.1
- Affected Devices: iPhone 8 Plus

### Reproduction Link

[Here](https://github.com/ilyalehchylin/educats-xamarin/blob/7c1ca437ad8d1fb3eab0d65388a56802b57a5778/source/EduCATS/Helpers/Devices/AppDevice.cs#L93) you can find how the code used.

jamesmontemagno:
Looks like the linker is being a bit aggressive. That code was re-architected with a lazy so let me update and get you a nuget to try asap.

jamesmontemagno:
I am curious as a workaround if you do something like this in your AppDelegate:

``
var test = new AVSpeechSynthesizer();
``

that may keep it linked in.

jamesmontemagno:
Can I bother you to try this one: https://dev.azure.com/xamarin/public/_build/results?buildId=16434&view=artifacts&type=publishedArtifacts

I did a different work around that I think is better. Sorry to keep asking :)

ilyalehchylin:
No problem, you may contact me as much as you need 🙂 

Unfortunately, with this NuGet app crashes. I will check the error.

ilyalehchylin:
The same error: `TextToSpeech error: The lazily-initialized type does not have a public, parameterless constructor`.

jamesmontemagno:
Got it! Let me try one more thing :)

jamesmontemagno:
Alright, one more for ya to test out: https://dev.azure.com/xamarin/public/_build/results?buildId=16525&view=artifacts&type=publishedArtifacts

ilyalehchylin:
Yup, it's working

jamesmontemagno:
[...]

PR #1208, opened 2020-04-03
GH-1207 iOS TTS - Don't use lazy as it isn't linker safe :(

jamesmontemagno:
### Description of Change ###

Seems as thought he lazy isn't linker safe, which makes sense, so re-architect as it isn't buying us anything.

### Bugs Fixed ###

- Related to issue #1207

### PR Checklist ###


mattleibow:
This might not be thread safe. Not a big one, since only one text can be spoken at a time, but may leak... I think that is why the lazy was used. Not sure what the best way to do this is. A simple lock with double check might be enough. Or, init that in the static ctor? If we are accessing the class, then we are going to use it. And I believe that is thread safe.

jamesmontemagno:
I will open a new PR and try that out.

PR #1209, opened 2020-04-04
GH-1207 - Preserve AVSpeechSynthesizer type when linking

jamesmontemagno:
### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #1207 

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/TextToSpeech/TextToSpeech.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/TextToSpeech/TextToSpeech.ios.tvos.watchos.cs
@@ -11,12 +11,6 @@ namespace Xamarin.Essentials
     {
         static readonly Lazy<AVSpeechSynthesizer> speechSynthesizer = new Lazy<AVSpeechSynthesizer>();
 
-        static TextToSpeech()
-        {
-            // Since we are using Lazy, this will ensure it doesn't get linked out.
-            var linkerSafe = new AVSpeechSynthesizer();
-        }
-
         internal static Task<IEnumerable<Locale>> PlatformGetLocalesAsync() =>
             Task.FromResult(AVSpeechSynthesisVoice.GetSpeechVoices()
                 .Select(v => new Locale(v.Language, null, v.Language, v.Identifier)));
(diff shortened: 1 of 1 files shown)


jamesmontemagno:
@mattleibow  looks like this fixes it :)

Xamarin.Essentials/TextToSpeech/TextToSpeech.ios.tvos.watchos.cs
```diff
@@ -11,6 +11,12 @@ public static partial class TextToSpeech
     {
         static readonly Lazy<AVSpeechSynthesizer> speechSynthesizer = new Lazy<AVSpeechSynthesizer>();
 
+        static TextToSpeech()
+        {
+            // Since we are using Lazy, this will ensure it doesn't get linked out.
+            var linkerSafe = new AVSpeechSynthesizer();
```
spouliot, line 17: If your code has `Lazy<AVSpeechSynthesizer>` then `AVSpeechSynthesizer` will marked by the linker - unless `speechSynthesizer` is not used.

Also adding a `[Preserve]` on the `speechSynthesizer` field would be better than adding a `.cctor` to your type, i.e. that requires more code being generated and has a performance impact at runtime.

spouliot: ah, the `.ctor()`, not the type itself, is removed. Try

```
[assembly: Preserve (typeof (AVSpeechSynthesizer))]
```
or (if it fails)
```
[assembly: Preserve (typeof (AVSpeechSynthesizer), AllMembers = true)]
```

Note that `new AVSpeechSynthesizer ();` will throw on macOS before 10.14, which would result in a `TypeLoadException` (which can be hard to catch since it's hard to be sure where/when the `.cctor` can be called).

jamesmontemagno: So yeah, the issue seems to be if we don't add anything then an exception on active is called:

The lazily-initialized type does not have a public, parameterless constructor.

I guess the thing is we don't really want to preserve it if people don't use TextToSpeech, it should be stripped out fully....

spouliot: Good point, you don't want to keep the type around unless it's actually needed. You can try

```
[Preserve (typeof (AVSpeechSynthesizer), AllMembers = true, Conditional = true)]
public static partial class TextToSpeech
```

which, in theory, should only be applied (Conditional) if the type is marked. 

If not possible then you better drop the `Lazy` since the `.cctor` will be creating the instance anyway - so just use it if you can't delay it :)

or add something that can't be optimized around the code, e.g.

```
if (DateTime.UtcNow.Ticks < 0)
    new AVSpeechSynthesizer();
```

so the code will be present (for the linker to see) but won't be executed.

PR #1219, opened 2020-04-12
Two small fixes for the Location class

janusw:
### Description of Change ###

This PR includes two small fixes for the `Location` class:

1. Reduce the verbosity of the `ToString` method, by printing only properties which are not `null`.
2. Add a new constructor which takes the arguments `latitude`, `longitude` and `altitude`.

### Bugs Fixed ###

None.

### API Changes ###

Added: 
 
- `public Location(double latitude, double longitude, double altitude)`

Changed:

 - None.

### Behavioral Changes ###

The output of `Location.ToString` is modified.

### PR Checklist ###


Xamarin.Essentials/Types/Location.shared.cs
```diff
@@ -104,9 +112,11 @@ public Location(Location point)
 
         public override string ToString() =>
             $"{nameof(Latitude)}: {Latitude}, {nameof(Longitude)}: {Longitude}, " +
-            $"{nameof(Altitude)}: {Altitude ?? 0}, {nameof(Accuracy)}: {Accuracy ?? 0}, " +
-            $"{nameof(VerticalAccuracy)}: {VerticalAccuracy ?? 0}, " +
-            $"{nameof(Speed)}: {Speed ?? 0}, {nameof(Course)}: {Course ?? 0}, " +
+            Altitude != null ? $"{nameof(Altitude)}: {Altitude}, " : string.Empty +
```
jamesmontemagno, line 115: Here you would just do:

```
$"{nameof(Altitude)}: {Altitude ?? string.Emptry}
```

jamesmontemagno: I still think it should output numbers though as the idea is to print everything out.

Xamarin.Essentials/Types/Location.shared.cs
```diff
@@ -39,6 +39,14 @@ public Location(double latitude, double longitude, DateTimeOffset timestamp)
             Timestamp = timestamp;
         }
 
+        public Location(double latitude, double longitude, double altitude)
```
jamesmontemagno, line 42: Is there a reason for this new constructor?

Xamarin.Essentials/Types/Location.shared.cs
janusw: I really don't think outputting zeros for non-existing information is a good idea. For the accuracies it's not so much of a problem, since those are usually positive. But the values for altitude, speed and course can easily become zero in real-world applications, so how do you distinguish an actual zero value from a `null` value, if `ToString` outputs a zero for both?

If we leave the value blank, then I don't see much reason to print out the property name either. Often a location will not include the speed and course information, in particular if the variable really just specifies a 'location' in the actual sense (i.e. without movement). In that case it can be a bit annoying if those property names are printed all the time, in particular if you print out lots of locations.

Xamarin.Essentials/Types/Location.shared.cs
janusw: Well, IMHO there is as much a reason for this as for those with (latitude, longitude) or (latitude, longitude, timestamp) parameters. None of them is strictly necessary, since the same can be accomplished also via an object initializer. But all of them can be handy, depending on the specific application.

It seems that the current state of Xamerin.Essentials.Location is mostly tailored towards 2D navigation-type applications, but there certainly are many other applications for the Location class, in particular in 3D, where a location is not complete without the altitude information. That is the specific use case of the new constructor.

Btw, I just noticed that I forgot to include documentation for the new constructor. I will do that, if we can agree that it is a useful addition.

Xamarin.Essentials/Types/Location.shared.cs
mattleibow: We must determine the reason for the existence of the `ToString`. I would go and say that if it is null, then just print `Accuracy=,` This is purely for devs, so the zero was wrong, but excluding it is also not correct.

Xamarin.Essentials/Types/Location.shared.cs
mattleibow: I think a simpler constructor is not bad if this type is to be used anywhere outside of the library. And, I am sure it is, so adding it is a good thing.

Xamarin.Essentials/Types/Location.shared.cs
janusw: > Here you would just do:
> 
> ```
> $"{nameof(Altitude)}: {Altitude ?? string.Emptry}
> ```

This actually does not work:

```
Error CS0019: Operator '??' cannot be applied to operands of type 'double?' and 'string' (CS0019)
```

janusw: > I would go and say that if it is null, then just print `Accuracy=,` This is purely for devs, so the zero was wrong, but excluding it is also not correct.

Alright then, I'm now using the following pattern for the nullable data members:

```
$"{nameof(Altitude)}: " + Altitude != null ? $"{Altitude}" : string.Empty + ", " +
```

janusw:
Thanks for the comments, @jamesmontemagno and @mattleibow. I rebased the branch on top of the current 1.6.0 branch, and addressed your comments on the `ToString` method. I hope it is ok for 1.6.0 in this form.

janusw:
Does anyone understand the device-test failures? They don't sound like they are related to my changes ....

Xamarin.Essentials/Types/Location.shared.cs
```diff
@@ -104,9 +112,11 @@ public Location(Location point)
 
         public override string ToString() =>
             $"{nameof(Latitude)}: {Latitude}, {nameof(Longitude)}: {Longitude}, " +
-            $"{nameof(Altitude)}: {Altitude ?? 0}, {nameof(Accuracy)}: {Accuracy ?? 0}, " +
-            $"{nameof(VerticalAccuracy)}: {VerticalAccuracy ?? 0}, " +
-            $"{nameof(Speed)}: {Speed ?? 0}, {nameof(Course)}: {Course ?? 0}, " +
+            $"{nameof(Altitude)}: " + Altitude != null ? $"{Altitude}" : string.Empty + ", " +
```
pictos, line 115: I think that you can can simplify here and other places that use this null check.
```suggestion
            $"{nameof(Altitude)}: " + Altitude  ?? string.Empty + ", " +
```

janusw: Thanks for the suggestion, @pictos, but this does not seem to work. For your code above, I get the error:

```
Error CS0019: Operator '??' cannot be applied to operands of type 'string' and 'bool' (CS0019) (Xamarin.Essentials)
```

This is a bit strange, and probably due to a missing parenthesis. If I add parentheses like this:

```
$"{nameof(Altitude)}: " + (Altitude ?? string.Empty) + ", " +
```

then I get:

```
Error CS0019: Operator '??' cannot be applied to operands of type 'double?' and 'string' (CS0019) (Xamarin.Essentials)
```

This is the same error I already saw for James' suggestion above (which was similar to yours).

pictos: Yeah, this is odd, I'll take a look into the code to see if we are missing something. Before I suggest it, I [tried here](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQBADGQQMwAEA3mXS3Qfo8697oRXQAc6AXjoAiKAHswcOgDcIAG3RwxAbi7cWABQQBLTDG0AKAJQbSW1kNGZ0ixRas79hk+c1bPPACx1dBkZmIgB83s68AJzGQgD8seJ6UHR2DmIelqwAvmRZQA=)

pictos: @janusw I see the problem, it's because the `Altitude` is `double` and `string.Empty` is a `string`... To make it works you can use this approach if it's ok to use `null` as a value. You can play with this [here](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQBADGQQMwAEA3mXS3Qfo869wG4QJ0IdALx1McAO50AMgHsAxhBgBLWZgAUFAHQUkhHQEou3Fk1ImLdAIIAbFTHQATOCLHobN4yYC+Abi/cAay4hACc6hBaACqyAMowCMqYAObqBgb+5qzeZDmkZF649HCY6GB0ACLKsBCY8nAAqpjKMFBeZpZ0ANLKNrJgcPAIUEhBLACyvXBtWSx5hcWl5bb2TnAASnAAZogl9SHtY3RNUAAOcPLKW8pwjq66R1GICBBJroSjsyYAoh7Kp1BZMo7qJ8J9OgBxOBAkF0GjgyyxdAILYQequAAsXnmXyKbA4ckUKjUhy+3DxhKUqg0RjJrA6lhxnQpCipanUjlk6GANhcNipDmcSDonO5vLofRSLTWtM6DM6LGkArWrn5q2cmQVrDkUsFLlEkuS0o1R24UWUAxqYFOrgqSjg5oGAHktlsoIMtA0YPIAHKyCSaxkFOksFlE6kcrk8vnKoUiqPiw3GuDCu3wR1wF1uwZ0FSWmAQa2yyzyhVK9X6iWxuCBrU6o161VqBtrWsKjNWm2iPPTAvWtvcJmWMNsjSi6NVivC8eJ5vJ6cJlwQOzJ4sWUudcvJ1XVgeWevbg1zvV7iwrQ+CFcn02sDt9ruVe0ZrPumCe71+gNHIcWEfEjSUv+dCnECmAwGuJgbpYVx0OoIFJDAIiiKUHgQVqLAwAAFgg/piJI1gIMkZQlDAPruDY3wAB71Kc/7qJghbQlscGgeBGTBuhMjVq48FgVoW7XiGJgHo2oi8W+ImtjeLB3oWD5pg6FpwO+vr+qeJjnqJwGsVomlSUJ3BWPI8jImiACePE6UZJkvPIZnqdwABqiAqIoNjWaZdmWQhWjOQgrnLh5tn2dJdCxOctzeXx4VwLcDmsAAwlywyVuJWhJci7rxSwACSUAAGI4WA4wKAA1gACjhPDAogUVvnlhX9CV8gVVVNUIOpP4mHiCnPq6r50LJ1qMHQySDL4dCvhNXXkvQM4uAJKoMKN42TatM3BHNi4yMeS0rTAE1TXQG2hltYpwAA/NYV57WNB1rfdJ1sGd0ZXUF5kjXdh3rRxw4vbyV1+QF7nGZ5FnLV9D3Tb9v7/ZdYURXcEOrUdT14vNV0ZSln0oz9+RCXiwCyLINh0A1RXNa1sjVc4AjI/dqMw919B6c4mw7AgexwCE10Vuzux1NzhA4wzeNHHiIQAGzxuddAJcuJlqnAVQ1ILkay2qybxPwMALhru3ONr/nCoB1ISqy/7fJgjiptUBaC00LRQHQWBOwYIgAHyhfLNiK/aKv2/U6ia3qRu6+buprGHwp9OGahW44/HVjHFvUgn/EGymLvNK07H48y9BSzLE4+378AB7UQem2o5tx5g0fF4m1YJ3rE5JnqLeVHbleNDnzuu7nnvSaX7j+93aux6OYdJ1OtdTwW/kZ5HcYh2snft2v1vCgPUB5+LheENL81ywro/l+PVepzXk//g31eYHPltb13qv1I7rTZ27Q8GYlp9KxXE8r71wXhJZOj9qTT0kivIB6dFrQLrrAzO28+57wJgfI+20R7/wvnAdQoVj6r0NiAhEWoCGZ2jvg7ahC4At0ofrZeNCt6hQAW/Pun9c5HCghYKAEgWjyEwrBHeaESyhW4Iod0L9A69ydloHofQBhDCgCAURw4ADsdAKDZRMOIlwLDpGtC0JMXkSiVG/nUZo0xIptgQHcKgSx3AsI4SkOIKQVhCLETAk6dAMAXTrFqGNKiNE6IMQGLIZiQi86cTyJ0NG9BqbPBqmwQgFBBpxASEkVI7thBex/iwAAJAAIgYCEpi6g4FwAMN4EAjByneGFMUxiYSymZ0qdUhgUC4B1LoAUugABqUKhSGmhOYqzCpVTGCjIutENJiQUhpDoBdK6IRtDfGtDAMy3gem9O6cKApbZvBAA)

`$"{nameof(Altitude)}: {Altitude ?? null}" + ", " +`

or 

` $"{nameof(Altitude)}: {Altitude.ToString() ?? string.Empty}" + ", "`
or 
` $"{nameof(Altitude)}: {Altitude?.ToString() ?? string.Empty}" + ", "`

janusw: Ok, your last variant works well. But it turns out one can also just use ...

```
$"{nameof(Altitude)}: {Altitude}, "
```

... which has the desired effect of printing an empty string if Altitude is null. :)

janusw:
Alright, rebased once more. `Location.ToString` became a bit simpler now ...

Xamarin.Essentials/Types/Location.shared.cs
pictos: Now the ToString override looks amazing.

janusw:
@jamesmontemagno @mattleibow Hey guys, I'd need another review here. Could someone add the 1.6.0 tag (hoping this is ok for release 1.6)?

janusw:
Ping! @mattleibow: Is there any chance to still get this into 1.6.0? It's really a minor modification and has been lingering here for a while.
Thanks,
Janus

mattleibow:
I think we can! I rebased and now we shall see. If all is green, in it goes!

janusw:
> I think we can!

Cool :D

> I rebased and now we shall see. If all is green, in it goes!

Looks like some checks still have not completed (after a week)?!? Probably something went wrong there ...?

mattleibow (approved):

janusw:
Thanks a lot for finally merging this one! I was holding my breath here for quite some time ;)

mattleibow:
Heh. Sorry it took so long, Thanks for hanging in there.

PR #1220, opened 2020-04-13
Mitigate MD5 usage

Redth:
### Description of Change ###

In the very early versions of Caboodle we did not create separate shared preferences on android for secure storage usage, and instead we chose to store secure storage encrypted values directly in files.  We used the key as the filename, which meant we needed to ensure the filename was valid.  To do so, we introduced an MD5 hash of the key to use as a filename.

Fast forward a bit, and as this detail was forgotten and we switched to using shared preferences to store the secure values, we kept the md5 hash code of the key, which was unnecessary.

Present day, we need to remove MD5 usage (even though it was never used for security reasons).  Originally we were going to replace the usage with CRC64, but after some more discovery @mattleibow and I looked back at the history to see why MD5 was only used for android and nowhere else (not even in Preferences for android which used the same shared preferences native API's for storage).  After some detective work, we re-learned the origin story of this MD5 usage and have determined it's not necessary to use a hash at all anymore, so CRC64 isn't needed.

We still need have the problem which is we don't know if some existing values were saved with the `md5` hash, so we have to for some time enable a mode where we can look to see if the `md5` hash version of the key exists, and if so, save it under the new unhashed key.  Eventually this check will be removed as existing data should have been sufficiently migrated to the unhashed key.

### API Changes ###

Added: 
 
- `bool SecureStorage.LegacyKeyHashFallback { get; set; }`

### Behavioral Changes ###

On Android, we have switched to hashing the key with `crc64` instead of `md5`.

If `SecureStorage.LegacyKeyHashFallback` is `true`:

Any time `GetAsync` is called, if the (now, `crc64`) key does not exist, the old `md5` key is also checked.  If that exists, the value is saved under the new `crc64` key, and the old `md5` keyed entry is removed.

In `SetAsync` and `RemoveAsync` calls, the existence of old values stored with the `md5` is checked for, and those are removed if they exist as well.

### PR Checklist ###


Xamarin.Essentials/AssemblyInfo/AssemblyInfo.shared.cs
```diff
@@ -5,3 +5,4 @@
 [assembly: InternalsVisibleTo("XamarinEssentialsDeviceTestsUWP")]
 [assembly: InternalsVisibleTo("XamarinEssentialsDeviceTestsShared")]
 [assembly: InternalsVisibleTo("XamarinEssentialsDeviceTestsiOS")]
+[assembly: InternalsVisibleTo("Xamarin.Forms.Core")]
```
mattleibow, line 8: Not sure about this... Do we need to? Can they just pull in their own from source?

Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
```diff
@@ -14,14 +14,42 @@ namespace Xamarin.Essentials
...
+                    // If previously stored with md5 key, save with crc key
+                    var md5Key = Utils.Md5Hash(key);
+                    if (Preferences.ContainsKey(md5Key, Alias))
+                    {
+                        var v = Preferences.Get(md5Key, defaultEncStr, Alias);
+                        Preferences.Set(keyHash, v, Alias);
+
+                        try
+                        {
+                            Preferences.Remove(md5Key, Alias);
```
mattleibow, line 43: Do we want to read and write this many times in the get? It is not bad, but what is the issue with just reading the new, and then falling back to the old? Let the "set" methods clean up.

Xamarin.Essentials/AssemblyInfo/AssemblyInfo.shared.cs
Redth: Yeah likely they will keep doing this for now.

Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
Redth: My concern is if you never Set the value again you never get migrated.  Trying to be proactive and get data moved over for when we eventually drop this.

Xamarin.Essentials/AssemblyInfo/AssemblyInfo.shared.cs
Redth: Removed.

Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
Redth: We decided to keep this to get data migrating sooner.

Xamarin.Essentials/SecureStorage/SecureStorage.android.cs
```diff
@@ -14,14 +14,43 @@ namespace Xamarin.Essentials
...
-            var encStr = Preferences.Get(Utils.Md5Hash(key), defaultEncStr, Alias);
+            string encStr = default;
+            var foundLegacyValue = false;
+
+            if (LegacyKeyHashFallback)
+            {
+                // If not found, could have been previously stored with md5 key
+                if (!Preferences.ContainsKey(key, Alias))
+                {
+                    // If previously stored with md5 key, save with crc key
```
mattleibow, line 33:
```suggestion
                    // If previously stored with md5 key, save with new key
```

PR #1194, opened 2020-03-27
Added "SaveDialog" support to FilePicker class.

Peckmore:
### Description of Change ###

- Implemented `PickFileToSaveAsync` for iOS, Android, UWP, macOS, and throw `NotImplementedInReferenceAssemblyException` for Tizen, .Net Standard, tvOS, watchOS.
- Added `SuggestedFileName` to `PickOptions` to suggest a default file name when saving.
- Added `OpenWriteStreamAsync` to `FilePickerResult` to open a file for writing.
- Changed the iOS picker to _Open_ a file instead of _Import_ a file to prevent read stream locks when trying to open a write stream.
- Added Android support for `Intent.ActionOpenDocument`.
- Added `PickMultipleFilesAsync()` overload to call without specifying `PickOptions` (for consistency with `PickFileAsync()`).

### Bugs Fixed ###

- Related to issue #251, #975, #613, #790

### API Changes ###

Added: 
 
**FilePicker**
- `public static Task<FilePickerResult> PickFileToSaveAsync()`
- `public static Task<FilePickerResult> PickFileToSaveAsync(PickOptions options)`
- `public static Task<IEnumerable<FilePickerResult>> PickMultipleFilesAsync()`

**PickOptions**
- `public string SuggestedFileName { get; set; }`

**FilePickerResult**
- `public Task<Stream> OpenWriteStreamAsync()`

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Has samples (if omitted, state reason in description)
- [X] Rebased on top of master at time of PR
- [X] Changes adhere to coding standard
- [ ] Updated documentation ([see walkthrough](https://github.com/xamarin/Essentials/wiki/Documenting-your-code-with-mdoc))

PR #1221, opened 2020-04-14
File picker improvements

vividos:
### Description of Change ###

This is a PR with improvements of the FilePicker implementation on the 'dev/file-picker' branch. Some of the changesets are ports from [jfversluis' FilePicker implementation](https://github.com/jfversluis/FilePicker-Plugin-for-Xamarin-and-Windows).

### Bugs Fixed ###

- 0d24658 fixes Android permission handling; one user reported that using `ActionOpenDocument` and calling `TakePersistableUriPermission` fixes a permission issue on newer Android version
- 7e6237f  uses the MIME type when the file name doesn't contain a file extension yet. This is for users that determine what to do with the picked file based on the file extension and the content provider doesn't provide the filename.
- e84334d: When using multi-file picking, Android returns data in `ClipData`, except for when the user only selects one file (duh!)
- f378202 adds the change from #1194 to use UIDocumentPickerMode.Open on iOS

### API Changes ###

None

### Behavioral Changes ###

The returned picked file patjs should now be accessible even after app restarts.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.android.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.android.cs
@@ -91,11 +91,10 @@ namespace Xamarin.Essentials
                 var result = await IntermediateActivity.StartAsync(pickerIntent, requestCodeFilePicker);
 
                 var resultList = new List<FilePickerResult>();
-                for (var index = 0; index < result.ClipData.ItemCount; index++)
-                {
-                    var data = result.ClipData.GetItemAt(index);
 
-                    var contentUri = data.Uri;
+                if (result.ClipData == null)
+                {
+                    var contentUri = result.Data;
 
                     if (Build.VERSION.SdkInt >= BuildVersionCodes.Kitkat)
                     {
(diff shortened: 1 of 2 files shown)


vividos:
The build currently fails due to this error message:

`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(515,3): Error XA0121: Assembly 'Xamarin.Android.Support.CustomTabs' is using '[assembly: Java.Interop.JavaLibraryReferenceAttribute]', which is no longer supported. Use a newer version of this NuGet package or notify the library author.`

I think this is fixed by switching the base branch to 1.6.0, since there the NuGet packages were updated already.

Peckmore:
@vividos @Redth Thanks for taking the tweak from #1194 to use UIDocumentPickerMode.Open on iOS. :) Now that we are using the Open mode you might also want to take the changes to the FilePickerResult class from FilePicker.ios.cs from the same PR (if the whole PR isn't accepted).

Previously, when using UIDocumentPickerMode.Import, the FilePickerResult class would immediately open a stream to the picked file to prevent the OS from removing it (because Import creates a copy of the picked file). Now that we are using UIDocumentPickerMode.Open we no longer need to worry about the OS removing the file, so we can create the stream on demand when the user calls PlatformOpenReadStreamAsync().

vividos:
@Peckmore I'll check out the changes. I think there's more to come for file picker, including refactoring and a macOS and Tizen implementation.

Peckmore:
@vividos Cool. 👍 I've got a mac implementation ready to commit, including save dialog - would it be worth adding that to my PR #1194?

vividos:
Why not? Although I think the Essentials devs would be glad if the changes would be in several commits, in order to easier review the changes.

Issue #1238, opened 2020-04-23
[Bug] iOS Crash in Permissions.LocationWhenInUse on WiFi iPad Air


Magendanz:
### Description

Since Xamarin.Essentials  v1.5 I've been unable to pass App Store review for my iOS release because it's consistently crashed on launch for the iPad Air 3rd Gen (Wi-Fi Only).  Unfortunately, I don't have this device, and so I've been unable to reproduce the bug myself.

### Steps to Reproduce

1. Submit app that uses Geolocation to Apple App Store for review.
2. Wait a couple of days.
3. They send you back crash logs (with no symbols), which only happen on iPad11,3.

### Expected Behavior

My app was passing review prior to upgrading to Xamarin.Essentials v1.5, and the other changes made between releases were minor and unrelated to geolocation or permissions.

### Actual Behavior

While the crash logs Apple provided were pretty useless, I was eventually able to collect a crash report from App Center:

```
Xamarin Exception Stack:
System.NullReferenceException: Object reference not set to an instance of an object
  at Xamarin.Essentials.Permissions+LocationWhenInUse+<>c__DisplayClass6_0.<RequestLocationAsync>g__LocationAuthCallback|0 (System.Object sender, CoreLocation.CLAuthorizationChangedEventArgs e) <0x105d1e7bc + 0x00164> in <68271ee3b0404411a54069b48d3f21d4#9ef9d2b073382932d73031f45271bd71>:0
  at CoreLocation.CLLocationManager+_CLLocationManagerDelegate.AuthorizationChanged (CoreLocation.CLLocationManager manager, CoreLocation.CLAuthorizationStatus status) <0x105b0dd94 + 0x0005f> in <698888335fd24bdca1317448cda6f409#9ef9d2b073382932d73031f45271bd71>:0
  at (wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
  at UIKit.UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate) <0x105b19a50 + 0x0003f> in <698888335fd24bdca1317448cda6f409#9ef9d2b073382932d73031f45271bd71>:0
  at UIKit.UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName) <0x105b199a8 + 0x00053> in <698888335fd24bdca1317448cda6f409#9ef9d2b073382932d73031f45271bd71>:0
[...]

PR #1245, opened 2020-04-24
GH-1238: Fix permissions not throwing exception

jamesmontemagno:
We used to do this before using the new api, so need ot do checks.
Also put in some null checks on locaiton manager for good measuere

### Bugs Fixed ###

- Related to issue #1238

### Behavioral Changes ###

Brings it back to expected behavior.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
@@ -161,17 +161,21 @@ namespace Xamarin.Essentials
                                     try
                                     {
                                         // Wait for a timeout to see if the check is complete
-                                        if (!tcs.Task.IsCompleted)
+                                        if (tcs != null && !tcs.Task.IsCompleted)
                                         {
                                             locationManager.AuthorizationChanged -= LocationAuthCallback;
                                             tcs.TrySetResult(GetLocationStatus(whenInUse));
-                                            locationManager?.Dispose();
-                                            locationManager = null;
                                         }
                                     }
                                     catch (Exception ex)
                                     {
                                         Debug.WriteLine($"Exception processing location permission: {ex.Message}");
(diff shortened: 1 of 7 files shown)


Redth:
@jamesmontemagno maybe we should add an internal method that handles the throwing like we used to have instead?

jamesmontemagno:
yes, let me do that :)

Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
```diff
@@ -146,30 +150,44 @@ void LocationAuthCallback(object sender, CLAuthorizationChangedEventArgs e)
                     if (e.Status == CLAuthorizationStatus.NotDetermined)
                         return;
 
-                    if (previousState == CLAuthorizationStatus.AuthorizedWhenInUse && !whenInUse)
+                    try
```
mattleibow, line 153: This double exception swallow is a bit scary... What is this for? Can we pass it to the tcs?

jamesmontemagno: Yeah, I can do that especially if the tcs isn't null :)

Basically ... if you look at #1238 who knows that is happening so wanted to just catch it all!

Xamarin.Essentials/Permissions/Permissions.shared.cs
```diff
@@ -16,6 +16,15 @@ internal static void EnsureDeclared<TPermission>()
             where TPermission : BasePermission, new() =>
                 new TPermission().EnsureDeclared();
 
+        internal static async Task RequestAndVerifyAsync<TPermission>()
```
mattleibow, line 19: Can we change this to be `EnsureGrantedAsync` or something else Ensure-y to follow the pattern of check/request/ensure? I am never a fan of "and" in method names.

PR #1250, opened 2020-04-28
Fixes #1247 Must set SourceRect in iOS 13

### LINKED ISSUE #1247: [Bug] Share Text does not work as expected on iPadOS 13.4
### Description
Siimilar to #979, but appears to have returned.

Share Text for iPadOS 13.4 does not bring up "pop-over" window UNLESS  app is displayed with certain split screen modes.

It appears to work fine on iPhone and iPadOS 12.x

### Steps to Reproduce
1. Attached is a sample program, which provides four buttons for the various ways that a Share for text can be done, including using PresentationSourceBounds.
2. Click each button to test.  Note that a click that does not work will need to "cleared" by clicking somewhere else before clicking next button.

### Expected Behavior
Pop-Over (activity controller) allows shared text to be sent to various apps

### Actual Behavior
No Pop-Over on iPad with 13.4, UNLESS in split screen mode and narrow (1/2 on iPad Air, 1/3 on iPad Pro 12.9").
Note that the device "thinks" there is a pop-over, and with one click in "full screen" mode, then going to split-screen, a pop-over appears.
[ShareTest.zip](https://github.com/xamarin/Essentials/files/4533704/ShareTest.zip)

### Basic Information
- Version with issue: 1.5.3.1, XF 4.5.0.657
- Last known good version: unknown
- IDE:  16.5.4
- Platform Target Frameworks: 
  - iOS:  13.4
- Affected Devices:
   iPad Air, iPad Pro 12.9

### Reproduction Link

[ShareTest.zip](https://github.com/xamarin/Essentials/files/4533710/ShareTest.zip)

jamesmontemagno:
### Description of Change ###

iPadOS really wants you to set a sourcerect :( it will default to 0,0,0,0, but developers should specify a correct area to open it.

### Bugs Fixed ###

- Related to issue #1247

### API Changes ###
None

### Behavioral Changes ###

Will now open in the same place on iOS 13 as iOS 12
### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Share/Share.ios.cs
+++ b/Xamarin.Essentials/Share/Share.ios.cs
@@ -29,7 +29,7 @@ namespace Xamarin.Essentials
             {
                 activityController.PopoverPresentationController.SourceView = vc.View;
 
-                if (request.PresentationSourceBounds != Rectangle.Empty)
+                if (request.PresentationSourceBounds != Rectangle.Empty || Platform.HasOSVersion(13, 0))
                     activityController.PopoverPresentationController.SourceRect = request.PresentationSourceBounds.ToPlatformRectangle();
             }
 


mattleibow:
Does this help? Do we just have to set the position, even if it is 0? It will set it if it is not empty.

jamesmontemagno:
Yeah, you just always need to set it... i guess if they set it to null it will throw.

The default is Empty so it will set it to empty on iOS 13 which will pop it up there. Tested it all out.

anders9ustafsson:
Has this fix really been transferred to 1.6? I have just now installed Essentials 1.6.1, and the share UI does not pop-up on an iPad with OS 14.3. Only if I keep the workaround proposed [here](https://github.com/xamarin/Essentials/issues/1247#issuecomment-620737045) the share UI pops up.

mattleibow:
@anders9ustafsson that is not a workaround because you must provide the source rectangle yourself. iOS does not know where to put the popup.

anders9ustafsson:
@mattleibow You mean that this PR does not provide the sufficient fix for my problem, or? The source code that I referenced above is sufficient for displaying the Share dialog on iPad with both Essentials 1.5.3.2 and 1.6.1. However, I was under the impression from the discussion in the associated issue #1247 that I should be able to display the Share dialog on iPad even without the extra code (the "workaround"). But as far as I can see, I still need the extra code for the Share dialog to consistently show up on all devices.

Issue #1225, opened 2020-04-15
[Feature Request] Use Fallback MIME type if not found on files.


flish:
### Description

Sending an email with a single attachment using await Email.ComposeAsync(message); on iOS crashes with a null reference as per below stack trace, attachment is a text file with file name debug.log and is less than 0.5MB in size.  Shared code in Android project works fine.

Using Xamarin.Essentials 1.5.2 nuget with Xamarin Forms 4.4.0991640

### Steps to Reproduce

Code example

```
var message = new EmailMessage
{
    Subject = "test with attach",
    Body = "test email body<br/>",
    To = "test@test.com"
};

message.BodyFormat = EmailBodyFormat.PlainText;
if (File.Exists(attachmentPath))
{
    message.Attachments.Add(new EmailAttachment(attachmentPath));
}

await Email.ComposeAsync(message);
```

### Expected Behavior

Popup iOS mail app with precomposed email and attachment

### Actual Behavior

Unhandled Exception;

```
NullReferenceException: Object reference not set to an instance of an object
  File "d:\a\1\s\Xamarin.Essentials\FileSystem\FileSystem.ios.tvos.watchos.cs", line 53, col 13, in PlatformGetContentType
    System.String PlatformGetContentType(System.String)
  File "d:\a\1\s\Xamarin.Essentials\FileSystem\FileSystem.shared.cs", line 69, col 17, in GetContentType
    System.String GetContentType()
  File "d:\a\1\s\Xamarin.Essentials\FileSystem\FileSystem.shared.cs", line 55, col 20, in get_ContentType
    System.String get_ContentType()
  File "d:\a\1\s\Xamarin.Essentials\Email\Email.ios.cs", line 57, col 21, in ComposeWithMailCompose
    System.Threading.Tasks.Task ComposeWithMailCompose(Xamarin.Essentials.EmailMessage)
  File "d:\a\1\s\Xamarin.Essentials\Email\Email.ios.cs", line 29, col 17, in PlatformComposeAsync
    System.Threading.Tasks.Task PlatformComposeAsync(Xamarin.Essentials.EmailMessage)
  File "d:\a\1\s\Xamarin.Essentials\Email\Email.shared.cs", line 23, col 13, in ComposeAsync
    System.Threading.Tasks.Task ComposeAsync(Xamarin.Essentials.EmailMessage)
[...]

PR #1262, opened 2020-05-05
Flish/#1225 default mime type for attachments

flish:
### Description of Change ###

- Fixed Mimetypes being null and throwing a null exception if mimetype not found on iOS / TVOS
- default Mimetype if not found (null) to be "application/octet-stream" as suggested

No new tests or documentation changes required, sample / test app can be provided.  Simple test is to try an attach a file with .log or .db extension to an email, which are not dound as mimetypes so woudl not be attached (exception)

```
var message = new EmailMessage
{
	Subject = "TEST", Body = "BODY", To = new List<string> {"help@xyroh.com"}
};

message.BodyFormat = EmailBodyFormat.PlainText;
if (File.Exists(attachmentPath))
{
	Console.WriteLine("FILE " + attachmentPath + " FOUND");
	message.Attachments.Add(
		new EmailAttachment(attachmentPath)); //works as long as extension is txt (mime types!)
}

try
{
	foreach (var attach in message.Attachments)
	{
		Console.WriteLine("Type " + attach.ContentType.ToString());
	}
	await Email.ComposeAsync(message);
}
```

### Bugs Fixed ###

- Related to issue #1225 

### API Changes ###

None

### Behavioral Changes ###

If you were checking for null as as the ContentType (on non iOS platforms) then you will now always get a result, iOS / tvOS was throwing an exception anyway

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
@@ -50,14 +50,7 @@ namespace Xamarin.Essentials
 
             var id = UTType.CreatePreferredIdentifier(UTType.TagClassFilenameExtension, extension, null);
             var mimeTypes = UTType.CopyAllTags(id, UTType.TagClassMIMEType);
-            if (mimeTypes != null && mimeTypes.Length > 0)
-            {
-                return mimeTypes[0];
-            }
-            else
-            {
-                return null;
-            }
+            return mimeTypes?.Length > 0 ? mimeTypes[0] : null;
         }
 
         internal void PlatformInit(FileBase file)
--- a/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
@@ -50,7 +50,14 @@ namespace Xamarin.Essentials
 
             var id = UTType.CreatePreferredIdentifier(UTType.TagClassFilenameExtension, extension, null);
             var mimeTypes = UTType.CopyAllTags(id, UTType.TagClassMIMEType);
-            return mimeTypes.Length > 0 ? mimeTypes[0] : null;
+            if (mimeTypes != null && mimeTypes.Length > 0)
+            {
+                return mimeTypes[0];
+            }
+            else
+            {
+                return null;
+            }
         }
 
         internal void PlatformInit(FileBase file)
--- a/Xamarin.Essentials/FileSystem/FileSystem.shared.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.shared.cs
@@ -73,7 +73,9 @@ namespace Xamarin.Essentials
 
             // we haven't been able to determine this
             // leave it up to the sender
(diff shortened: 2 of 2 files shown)


Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
```diff
@@ -50,7 +50,14 @@ internal static string PlatformGetContentType(string extension)
 
             var id = UTType.CreatePreferredIdentifier(UTType.TagClassFilenameExtension, extension, null);
             var mimeTypes = UTType.CopyAllTags(id, UTType.TagClassMIMEType);
-            return mimeTypes.Length > 0 ? mimeTypes[0] : null;
+            if (mimeTypes != null && mimeTypes.Length > 0)
```
jamesmontemagno, line 53: return (mimesTypes?.Length ?? 0) > 0 ? mimeTypes[0] : null;

Xamarin.Essentials/FileSystem/FileSystem.shared.cs
```diff
@@ -73,7 +73,9 @@ internal string GetContentType()
 
             // we haven't been able to determine this
             // leave it up to the sender
-            return null;
+            // return null;
+            // Instead of returning null return a catch all mime-type - see #1225
```
jamesmontemagno, line 77: don't like to put in issue numbers in the code, as we have blame to help us out and link back.

Do we want it to be octet-stream? does everythign support that?

flish: Fair enough ref comments,  octet-stream was suggested in the comments and is as generic as it comes, there are no real application handlers for it so gets attached / downloaded and up to the end user to handle

Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
flish: I’m old school readable code, like my whitespace :-] Will tweak in the am if no more suggestions

Xamarin.Essentials/FileSystem/FileSystem.shared.cs
jamesmontemagno: I think if we just test it out on iOS/Android that then would be good... just want to does work... we can just document it and recommend people set it manually.

Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.cs
mattleibow: You can also just do

return mimesTypes?.Length > 0 ? mimeTypes[0] : null;

Null never matches int, so will always be false.

flish:
@jamesmontemagno  @mattleibow - just pushed the changes, there's a Xam forms sample project at https://xyroh-my.sharepoint.com/:u:/p/af/ETfQysgpf15KksFR8uRVeWQBZJPioTui9Ijv2E-bkA1Ifw?e=jCEe1B if you wnat, but can confirm the octet-stream mime type works fine as a default on iOS and Android, just up to the individual device to do something specific with it

Issue #1267, opened 2020-05-10
[Enhancement] Return information from provider using access token.


solomonfried:
## Summary
The WebAuthenticator returns an access token and expiration date to the client. In practically all cases, the client will then call the underlying provider to get the Email, and possibly other information (Name, Image, etc...)

It would be great if the WebAuthenticator provided this service for the major providers.

## API Changes
Add Provider Enum

```
public enum AuthenticatorProvider{
  Apple,
  Google,
  Facebook
  ….
}
```

Add static call to WebAuthenticor.
`public static Dictionary<string, string> WebAuthenticator.GetUserInformation(string accessToken, AuthenticatorProvider provider);
`

e.g.

```
token = await AppleSignInAuthenticator.AuthenticateAsync();
userinfo = WebAuthenticator.GetUserInformation(token.AccessToken, AuthenticatorProvider.Apple); 

```
.

solomonfried:
I did manage to get the email information from the server call
`Request.HttpContext.AuthenticateAsync(scheme);`
and return it to the client.

If you look at the WebAuth controller example here...
[https://github.com/xamarin/Essentials/blob/develop/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs](https://github.com/xamarin/Essentials/blob/develop/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs)

I modified the code in the Get as follows:
```
if (!auth.Succeeded
                    || auth.Principal == null
                    || !auth.Principal.Identities.Any(id => id.IsAuthenticated)
                    || string.IsNullOrEmpty(auth.Properties.GetTokenValue("access_token")))
                {
                    // Not authenticated, challenge
                    await Request.HttpContext.ChallengeAsync(scheme);
                }
                else
                {
                   // ADDED THE FOLLOWING
                    var claims = auth.Principal.Identities.FirstOrDefault()?.Claims;
                    string email = string.Empty;
                    if (claims != null)
                        email = claims.FirstOrDefault(c =>
                                c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress")
                            ?.Value;
                    // Get parameters to send back to the callback
                    var qs = new Dictionary<string, string>
                    {
                        {"access_token", auth.Properties.GetTokenValue("access_token")},
                        {"refresh_token", auth.Properties.GetTokenValue("refresh_token") ?? string.Empty},
[...]

PR #1287, opened 2020-05-22
Return user's email as part of the properties.

solomonfried:
get the email from the Claims (not sure in what situation multiple identities would be returned, but I am taking the Claims from the first one,  then add the email to the returning Dictionary.

On the Client the user can retrieve the email with....

WebAuthenticatorResult result = await WebAuthenticator.AuthenticateAsync(authUrl, callbackUrl);
if (result.Properties.Contains("email"))
    string email = Uri.UnescapeDataString(result.Properties["email"]);

### Description of Change ###

Return email with auth properties.

### DIFF
--- a/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs
+++ b/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs
@@ -33,7 +33,7 @@ namespace Sample.Server.WebAuthenticator
             {
                var claims = auth.Principal.Identities.FirstOrDefault()?.Claims;
                     string email = string.Empty;
-                    email = claims?.FirstOrDefault(c => c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress")?.Value;
+                    email = claims?.FirstOrDefault(c => c.Type == System.Security.Claims.ClaimTypes.Email)?.Value;
                 // Get parameters to send back to the callback
                 var qs = new Dictionary<string, string>
             {
--- a/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs
+++ b/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs
@@ -31,12 +31,16 @@ namespace Sample.Server.WebAuthenticator
             }
             else
             {
+               var claims = auth.Principal.Identities.FirstOrDefault()?.Claims;
+                    string email = string.Empty;
+                    email = claims?.FirstOrDefault(c => c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress")?.Value;
                 // Get parameters to send back to the callback
                 var qs = new Dictionary<string, string>
             {
                 { "access_token", auth.Properties.GetTokenValue("access_token") },
                 { "refresh_token", auth.Properties.GetTokenValue("refresh_token") ?? string.Empty },
-                { "expires", (auth.Properties.ExpiresUtc?.ToUnixTimeSeconds() ?? -1).ToString() }
+                { "expires", (auth.Properties.ExpiresUtc?.ToUnixTimeSeconds() ?? -1).ToString() },
+                {"email", email}
             };
 
                 // Build the result url


solomonfried:
I have tried it and I always get the email back.

Of course, Apple being Apple, they make things a bit more complicated. According to Apple, they only return the scope information during initial authorization. (see here https://forums.developer.apple.com/thread/119826 )
They claim that the “id_token” JWT in the returned information can be Decoded to find the email.

I have not figured out how to get to the id_token using WebAuthenticator and I also can’t find a JWT decoder that works in Xamarin.

This is the information from apple.


“The scopes are only shared once, during the initial authorization.   Once the SIWA API returns the information, it is up to the client application to store it within their sytems [sic].   For follow up authorizations, the "user" identifier returned as part of SIWA API should be used as a means of fetching the previously shared information from your system.

Again I do not see any “user” property in the returned information, but the returned claims do contain a “nameidentifier” http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier


So on the server when I get an Apple auth request and it returns with an email, I save the nameidentifier and email in a database. Then when on subsequent requests the email is not provided, I look it up by nameidentifier and return the email.

From my testing it seems like I only get the blank emails when I decline the Apple Touch ID. When I select the “Use Different Apple ID” option, it returns the email (or the Apple private email)


If I use the AppleSignInAuthenticator.AuthenticateAsync, thereby bypassing my server code, then the returned WebauthenticationResult will contain a “user_id” which does match the nameidentifier in the claims on the server.

So in that event, I will need to call my service from the client to save/retrieve the stored emails. It also returns the “id_token” JWT token, so if I figure out how to decode that, I will not need to make a service call.

Better yet!! If you guys know how to decode the id_token, then you can decode it in the  AppleSignInAuthenticator.AuthenticateAsync call and return it in the WebauthenticationResult !!


Bottom line is, that the code works.  It will return an email (or a blank if no email is found). But users will need to deal with Apple’s decision to not always return an email and not provide a service endpoint to retrieve an email from an AccessToken as Google, Facebook, and almost all other providers do.

solomonfried:
One addition to the email below.
I tried parsing the id_token that I read would contain the email. But it does not (probably only on the first authorization to the app)
It returns the following

{
    "iss": "https://appleid.apple.com",
    "aud": "com.pnqsys.classtars",
    "exp": "1590296205",
    "iat": "1590295605",
    "sub": "001191.dfe582ecaasasasasa7863d5bc4d64.1533",
    "c_hash": "euvDLdEe8AvKvwh95VYT2w",
    "auth_time": "1590295605",
    "nonce_supported": "true"
}


From: Solomon Fried
Sent: Sunday, May 24, 2020 12:34 AM
To: xamarin/Essentials <reply@reply.github.com>; xamarin/Essentials <Essentials@noreply.github.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [xamarin/Essentials] Return user's email as part of the properties. (#1287)

I have tried it and I always get the email back.

Of course, Apple being Apple, they make things a bit more complicated. According to Apple, they only return the scope information during initial authorization. (see here https://forums.developer.apple.com/thread/119826 )
They claim that the “id_token” JWT in the returned information can be Decoded to find the email.

I have not figured out how to get to the id_token using WebAuthenticator and I also can’t find a JWT decoder that works in Xamarin.

This is the information from apple.


“The scopes are only shared once, during the initial authorization.   Once the SIWA API returns the information, it is up to the client application to store it within their sytems [sic].   For follow up authorizations, the "user" identifier returned as part of SIWA API should be used as a means of fetching the previously shared information from your system.

Again I do not see any “user” property in the returned information, but the returned claims do contain a “nameidentifier” http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier


So on the server when I get an Apple auth request and it returns with an email, I save the nameidentifier and email in a database. Then when on subsequent requests the email is not provided, I look it up by nameidentifier and return the email.

From my testing it seems like I only get the blank emails when I decline the Apple Touch ID. When I select the “Use Different Apple ID” option, it returns the email (or the Apple private email)


If I use the AppleSignInAuthenticator.AuthenticateAsync, thereby bypassing my server code, then the returned WebauthenticationResult will contain a “user_id” which does match the nameidentifier in the claims on the server.

So in that event, I will need to call my service from the client to save/retrieve the stored emails. It also returns the “id_token” JWT token, so if I figure out how to decode that, I will not need to make a service call.

Better yet!! If you guys know how to decode the id_token, then you can decode it in the  AppleSignInAuthenticator.AuthenticateAsync call and return it in the WebauthenticationResult !!


Bottom line is, that the code works.  It will return an email (or a blank if no email is found). But users will need to deal with Apple’s decision to not always return an email and not provide a service endpoint to retrieve an email from an AccessToken as Google, Facebook, and almost all other providers do.

Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs
```diff
@@ -31,12 +31,16 @@ public async Task Get([FromRoute]string scheme)
             }
             else
             {
+               var claims = auth.Principal.Identities.FirstOrDefault()?.Claims;
+                    string email = string.Empty;
+                    email = claims?.FirstOrDefault(c => c.Type == "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress")?.Value;
```
Redth, line 36: We should really use the constant for this:

`System.Security.Claims.ClaimTypes.Email` as seen here: https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimtypes.email

solomonfried: Agreed. Was not aware of the ClassTypes Constants

jamesmontemagno:
Does this close #1267?

solomonfried:
I closed the issue.

I had updated the Git documentation to show how to always return an email

https://github.com/xamarin/Essentials/blob/develop/Samples/Sample.Server.WebAuthenticator/Controllers/MobileAuthController.cs

PR #1290, opened 2020-05-27
GTK Clipboard support

Bobface:
Based on https://github.com/stavroskasidis/XamarinClipboardPlugin/commit/b213a88d3d9d85d3600cd5084a53c670f51f0fb7

### Description of Change ###

Implements Clipboard support for GTK

### PR Checklist ###


jfversluis:
Thank you so much for your time and effort on this! As is probably clear at this time is that we're not going to add new platforms that we will support with Xamarin.Essentials. While this is a small feature, this also opens the door for adding potentially more which is not something we want to do anymore in this phase.

We appreciate the time you took to do this!

PR #1239, opened 2020-04-23
Add .NETStandard2.0 Connectivity

knocte:

PR #1291, opened 2020-05-27
Add .NETStandard2.0 Connectivity

Bobface:
Supersedes #1239 

### Description of Change ###

Adds .NETStandard2.0 Connectivity


jfversluis:
I think it's safe to say that we won't be merging this for Xamarin.Essentials anymore at this point. Thanks for all the time and effort spend on this!

Issue #1295, opened 2020-06-02
[Bug] iOS: WebAuthenticator leaks on cancel


MKuckert:
### Description

The WebAuthenticator leaks the `ASWebAuthenticationSession` or `SFAuthenticationSession` on iOS if the login attempt is being cancelled.

### Steps to Reproduce

Steps for iOS 12+

1. Call `WebAuthenticator.AuthenticateAsync`
2. Cancel the opening authentication view controller
3. Examine the field `WebAuthenticator.was`. It's non-null and still holds the `ASWebAuthenticationSession`

### Expected Behavior

The session is disposed in all cases.

### Actual Behavior

The session is only disposed in the successful case. The session leaks for the cancelled or failing cases.

### Basic Information

- Version with issue: latest
- Last known good version: none (known to me)
- IDE: independent of any IDE
- Platform Target Frameworks: 
  - iOS:  13
- Nuget Packages: none. Reproducible with the sample app
- Affected Devices: every iOS device

### Reproduction Link

One can reproduce with the sample app.

mattleibow:
This was fixed in #1296

PR #1296, opened 2020-06-02
iOS: Fixes leaks in WebAuthenticator

MKuckert:
The WebAuthenticator leaks the `ASWebAuthenticationSession` or `SFAuthenticationSession` if the login attempt is being cancelled.

### Description of Change ###

Surrounds the awaited task with a `try/finally` to dispose the authentication session even for the cancelled or failed cases.

Adds no tests as this is not automatically reproducible (would require ui test)

### Bugs Fixed ###

- Related to issue #1295
### API Changes ###

None

### Behavioral Changes ###

The `was` and `sf` fields are now disposed and set to null in all cases.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/WebAuthenticator/WebAuthenticator.ios.tvos.cs
+++ b/Xamarin.Essentials/WebAuthenticator/WebAuthenticator.ios.tvos.cs
@@ -67,21 +67,21 @@ namespace Xamarin.Essentials
                         void_objc_msgSend_IntPtr(was.Handle, ObjCRuntime.Selector.GetHandle("setPresentationContextProvider:"), ctx.Handle);
                     }
 
-                    was.Start();
-                    var result = await tcsResponse.Task;
-                    was?.Dispose();
-                    was = null;
-                    return result;
+                    using (was)
+                    {
+                        was.Start();
+                        return await tcsResponse.Task;
+                    }
                 }
 
                 if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
                 {
                     sf = new SFAuthenticationSession(new NSUrl(url.OriginalString), scheme, AuthSessionCallback);
-                    sf.Start();
-                    var result = await tcsResponse.Task;
-                    sf?.Dispose();
-                    sf = null;
-                    return result;
+                    using (sf)
+                    {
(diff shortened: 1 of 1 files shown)


Xamarin.Essentials/WebAuthenticator/WebAuthenticator.ios.tvos.cs
```diff
@@ -68,20 +68,32 @@ void AuthSessionCallback(NSUrl cbUrl, NSError error)
                     }
 
                     was.Start();
-                    var result = await tcsResponse.Task;
-                    was?.Dispose();
-                    was = null;
-                    return result;
+                    try
+                    {
+                        var result = await tcsResponse.Task;
```
lindexi, line 73: Why not use using

MKuckert: Oh yes  🤦‍♂️. That would be a lot better.

MKuckert: Hmm, but a `using` wouldn't set the value to `null` again, which seems to be a good idea. So I'd stick to the current solution.

jamesmontemagno: not sure if it needs to be nulled out or not. I assume after dispose it will get gced up and then we will reassign.

MKuckert: So @jamesmontemagno you'd prefer a using instead and drop the null-assignment?

jamesmontemagno: I think should be fine with using @Redth

MKuckert: Changed the implementation to `using` now and force-pushed the branch to drop the previous commit.

MKuckert: /cc @jamesmontemagno @Redth

PR #1333, opened 2020-07-20
added HapticFeedback macOS

dimonovdd:
I made an implementation hapticFeedback for mac OS (https://github.com/xamarin/Essentials/pull/1138).
HapticFeedbackType.Click is not required on macOS, because it execute automatically when the touchpad is clicked


Xamarin.Essentials/HapticFeedback/HapticFeedback.macos.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Xamarin.Essentials
+{
+    public static partial class HapticFeedback
+    {
+        internal static bool IsSupported => true;
+
+        static void PlatformPerform(HapticFeedbackType type)
+        {
+            if (type == HapticFeedbackType.LongPress)
+                NSHapticFeedbackManager.DefaultPerformer.PerformFeedback(NSHapticFeedbackPattern.Alignment, NSHapticFeedbackPerformanceTime.Default);
```
mattleibow, line 14: Is alignment the correct pattern? What is that? How does it compare to generic?

dimonovdd: @mattleibow 
I didn't quite understand the question.
In total there are three types of the haptic feedback on MacOS:
- alignment
- Generic ?
- LevelChange ?
As I could tactually feel it, it seemed...

mattleibow: That is what I mean. You picked alignment. Is that the correct one for long press? Why are the others not the right one?

mattleibow: The reason I say this is because the docs say:
> A haptic feedback pattern to be used in response to the alignment of an object the user is dragging around.

dimonovdd: It may be replaced by "generic"

PR #1335, opened 2020-07-22
[Tizen] Update the FilePicker implementation

rookiejava:
### Description of Change ###

This is a fix for requesting permission in order for the file picker to work properly in Tizen.
As of Tizen 4.0 and later, media storage privilege (`http://tizen.org/privilege/mediastorage`) should be treated as user privacy privilege.
 - http://tizen.org/privilege/mediastorage
   - Allows the application to read and write files in media folders.
 - See [here ](https://www.tizen.org/privilege)full privilege.

### Bugs Fixed ###

None

### API Changes ###

None

### Behavioral Changes ###

In tizen, FilePicker requesting the user's consent to use.

### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.tizen.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.tizen.cs
@@ -11,10 +11,12 @@ namespace Xamarin.Essentials
 {
     public static partial class FilePicker
     {
-        static Task<IEnumerable<FilePickerResult>> PlatformPickAsync(PickOptions options, bool allowMultiple = false)
+        static List<FilePickerResult> resultFiles = new List<FilePickerResult>();
+
+        static async Task<IEnumerable<FilePickerResult>> PlatformPickAsync(PickOptions options, bool allowMultiple = false)
         {
             Permissions.EnsureDeclared<Permissions.LaunchApp>();
-            Permissions.EnsureDeclared<Permissions.StorageRead>();
+            await Permissions.EnsureGrantedAsync<Permissions.StorageRead>();
 
             var tcs = new TaskCompletionSource<IEnumerable<FilePickerResult>>();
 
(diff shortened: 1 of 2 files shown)


Xamarin.Essentials/FilePicker/FilePicker.tizen.cs
```diff
@@ -11,10 +11,12 @@ namespace Xamarin.Essentials
 {
     public static partial class FilePicker
     {
-        static Task<IEnumerable<FilePickerResult>> PlatformPickAsync(PickOptions options, bool allowMultiple = false)
+        static List<FilePickerResult> resultFiles = new List<FilePickerResult>();
```
mattleibow, line 14: Why was this moved out of the scope? I also see another variable that is unused.

https://github.com/xamarin/Essentials/blob/f8045e0b835e8ce30eab712a9456a72161da6acb/Xamarin.Essentials/FilePicker/FilePicker.tizen.cs#L31

Does it not work to use that and then
`return await tcs.Task`

I also think what is happening is that files are being added to the list, but never cleared out...

Xamarin.Essentials/Permissions/Permissions.tizen.cs
```diff
@@ -43,7 +42,8 @@ async Task<PermissionStatus> CheckPrivilegeAsync(bool ask)
 
                 foreach (var (tizenPrivilege, isRuntime) in tizenPrivileges)
                 {
-                    if (PrivacyPrivilegeManager.CheckPermission(tizenPrivilege) == CheckResult.Ask)
+                    var checkResult = PrivacyPrivilegeManager.CheckPermission(tizenPrivilege);
```
mattleibow, line 45: Nice

Xamarin.Essentials/FilePicker/FilePicker.tizen.cs
rookiejava: Thanks, @mattleibow and @redth! 😉

Issue #1130, opened 2020-02-27
[Spec] Added send sms silently API for Android and UWP


soroshsabz:
# Send SMS silently

I think is very good if we can send SMS background and silently when user permit our. this API useful for any app that need send SMS automatically.

It is also available at https://github.com/cjlotz/Xamarin.Plugins

# API

### Properties

| API | Description |
| ------------- | ------------- |
| CanComposeSilently | Gets or sets support of this feature in specific platform. |

### Methods

| API | Description |
| ------------- | ------------- |
| ComposeSilentlyAsync | Send SMS silently. |

# Platform Compatibility
- Target Frameworks: 
  - iOS: Not supported
  - Android: Supported 
  - UWP: Supported 
  - Tizen: Supported
# Backward Compatibility
There is no backward compatibility issue

# Difficulty : Low

soroshsabz:
see cjlotz#47

mattleibow:
As mentioned in https://github.com/xamarin/Essentials/pull/1338#issuecomment-673699506, we will not be adding the ability to send sms messages directly.

PR #1338, opened 2020-07-31
Add ability to send SMS in background. (see #1130)

soroshsabz:
### Description of Change ###

This PR complete some of requirements in #1130 
In this PR we have support send SMS silently in Android and UWP.

### API Changes ###

Added: 
 
- `static Task ComposeInBackgroundAsync(SmsMessage message)`
- `static bool IsComposeInBackgroundSupported`

Changed:

 - None

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Sms/Sms.android.cs
+++ b/Xamarin.Essentials/Sms/Sms.android.cs
@@ -4,7 +4,7 @@ using System.Threading.Tasks;
 using Android.Content;
 using Android.OS;
 using Android.Provider;
-
+using Android.Telephony;
 using AndroidUri = Android.Net.Uri;
 
 namespace Xamarin.Essentials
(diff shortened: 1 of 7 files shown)


soroshsabz:
I don't think these failed test because of my changes

soroshsabz:
@jamesmontemagno can you review this PR?

Thanks a lot

mattleibow:
Thanks for this PR. I have just been thinking that this might not be something that we would actually want to encourage developers to do.

Even though permissions may be granted and the user aware, this action is not always seen as desirable. And, the use case is pretty much limited to a custom SMS app.

But, you may be using this for some reason? 

@Redth @jamesmontemagno thoughts?

AmrAlSayed0:
Shouldn't ```IsComposeInBackgroundNotSupported``` just return false instead of throwing?

Redth:
Thanks for the contribution, we appreciate the interest! 

Unfortunately this functionality doesn't really fit in well with the spirit of Essentials, which is to enable common use cases and scenarios with a cross platform API which works on the majority of platforms supported.  Given that this only works on Android and the other platforms don't support this behaviour, and that the use case for this API would be rather specific, we feel it's best to leave up to individual app developers needing the functionality to implement in a dependency service or with some other pattern.

Thanks again!

PR #1340, opened 2020-08-02
Fix crash of clipboard sample when text is null

bares43:
### Description of Change ###
This fixes crash of clipboard sample when text is null.

### API Changes ###
None

### Behavioral Changes ###

Sample app doesn't crash anymore.

### PR Checklist ###

### DIFF
--- a/Samples/Samples/ViewModel/ClipboardViewModel.cs
+++ b/Samples/Samples/ViewModel/ClipboardViewModel.cs
@@ -59,7 +59,16 @@ namespace Samples.ViewModel
             LastCopied = $"Last copied text at {DateTime.UtcNow:T}";
         }
 
-        async void OnCopy() => await Clipboard.SetTextAsync(FieldValue);
+        async void OnCopy()
+        {
+            if (FieldValue == null)
+            {
+                await DisplayAlertAsync($"Text is null");
+                return;
+            }
+
+            await Clipboard.SetTextAsync(FieldValue);
+        }
 
         async void OnPaste()
         {


jamesmontemagno:
I am thinking null == string.empty :)

bares43:
Thanks for your review and reply.

Actually, I'm not sure what you mean by empty string. Do you mean add here https://github.com/xamarin/Essentials/blob/develop/Xamarin.Essentials/Clipboard/Clipboard.shared.cs#L9m something like this?
`
if (text == null)
{
    text = string.Empty;
}
`

In that case user would end up with empty string in the clipboard. [http://prntscr.com/tumh6s](http://prntscr.com/tumh6s)

Maybe it would be better to call return, when string is null. Like this:
`
if (text == null)
{
    return Task.FromResult<object>(null);
}
`

What do you think?

mattleibow:
Setting an empty or null might be a way to clear the clipboard... I think that might be a nice feature.

jamesmontemagno:
If you set string.Empty it will clear it, so probably if you pass in null it should just set it to string.Empty for you... or throw an exception lol

mattleibow:
Maybe for the sample, pass in "" if it is null, but for the actual API, throw?

jamesmontemagno:
i think converting it to "" is ideal

bares43:
Ok, so as I wrote above, I will just set `string.Empty` if value is `null` in `Clipboard.shared.cs`.

bares43:
I guess that since this PR is already merged, I don't have to change anything. Or do you think I should create another PR and add `string.Empty` to `Clipboard.share.cs` whent it's `null`? @jamesmontemagno @mattleibow

mattleibow:
Could you create a new PR for that? It would be awesome!

bares43:
> Could you create a new PR for that? It would be awesome! 

Sure :)

bares43:
New PR is here #1345

PR #1345, opened 2020-08-12
Clipboard: set string.Empty when value is null

bares43:
### Description of Change ###

As was discussed in #1340:  set `string.Empty` when value is `null` (in Clipboard). I also reverted null check in sample app, because it doesn't make any sense anymore.

### API Changes ###

None

### Behavioral Changes ###

Set empty string to value when it's null.

### PR Checklist ###


Xamarin.Essentials/Clipboard/Clipboard.shared.cs
```diff
@@ -6,7 +6,14 @@ namespace Xamarin.Essentials
     public static partial class Clipboard
     {
         public static Task SetTextAsync(string text)
-            => PlatformSetTextAsync(text);
+        {
+            if (text == null)
```
jamesmontemagno, line 10: should be able to do -> 

text ??= string.Empty;

mattleibow: Or even `return PlatformSetTextAsync(text ?? string.Empty);`

bares43: Thanks, it would be better this way :) Sometimes I almost forget what amazing features has C# to offer :D

mattleibow (approved):
Love it!

bares43:
I'm happy you like it 😀 do you think it's ready to merge? 😀

PR #1349, opened 2020-08-16
ContactAPI

pictos:
### Description of Change ###

This PR implements the Contacts API.

### Bugs Fixed ###

- Related to issue #299

### API Changes ###

Create a new folder `Contacts`.

Added: 
- `Task<PhoneContact> PickContactAsync();`
 
```csharp
public static class Contact.Shared {
  `public static Task<PhoneContact> PickContactAsync();`
}
```
**Android**
```csharp
public static class Contact.Android {
static Activity Activity => Platform.GetCurrentActivity(true);
static Task<PhoneContact> PlataformPickContactAsync();
internal static PhoneContact PlataformGetContacts(Net.Uri contactUri);
static ContactType GetContactType(string type);
}
```

**iOS**

```csharp
public static class Contact.iOS{
internal static Action<PhoneContact> CallBack { get; set; }
static Task<PhoneContact> PlataformPickContactAsync();
internal static PhoneContact GetContact(CNContact contact);
}
```
```csharp
public static class Contact.UWP{
static async Task<PhoneContact> PlataformPickContactAsync();
}
```
### Behavioral Changes ###
Now we can get the contacts info in the device and add new contacts.

ezgif com-optimize

### PR Checklist ###


Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -0,0 +1,171 @@
...
+
+namespace Xamarin.Essentials
+{
+    public static partial class Contacts
+    {
+        static async Task<Contact> PlatformPickContactAsync()
+        {
+            using var intent = new Intent(Intent.ActionPick);
+            intent.SetType(ContactsContract.CommonDataKinds.Phone.ContentType);
+            var result = await IntermediateActivity.StartAsync(intent, 101).ConfigureAwait(false);
```
mattleibow, line 17: This code needs to be moved to that central location... Not sure if that change was merged yet though.

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            return null;
+        }
+
+        internal static Contact PlatformGetContacts(Net.Uri contactUri)
+        {
+            if (contactUri == null)
+                return default;
+
+            using var context = Platform.GetCurrentActivity(true).ContentResolver;
```
mattleibow, line 30: Does this have to be the activity? Does the AppContext not work?

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -0,0 +1,171 @@
...
+            var bDate = string.Empty;
+
+            if (cur.MoveToFirst())
+            {
+                var typeOfContact = cur.GetString(cur.GetColumnIndex(ContactsContract.CommonDataKinds.Phone.InterfaceConsts.Type));
+
+                var name = cur.GetString(cur.GetColumnIndex(ContactsContract.Contacts.InterfaceConsts.DisplayName));
+                string id;
+
+                id = cur.GetString(cur.GetColumnIndex(ContactsContract.CommonDataKinds.Email.InterfaceConsts.ContactId));
```
mattleibow, line 44: Maybe condense this into a single line.

```suggestion
                var id = cur.GetString(cur.GetColumnIndex(ContactsContract.CommonDataKinds.Email.InterfaceConsts.ContactId));
```

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -0,0 +1,171 @@
...
+                        var phone = cursor.GetString(cursor.GetColumnIndex(projection[0]));
+                        var phoneType = cursor.GetString(cursor.GetColumnIndex(projection[1]));
+
+                        var contactType = GetPhoneContactType(phoneType);
+
+                        phones.Add(new ContactPhone(phone, contactType));
+                    }
+                    while (cursor.MoveToNext());
+                }
+                cursor.Close();
```
mattleibow, line 73: In terms of code follow-ability, it is fine, but maybe this and the emails chunks can be moved to a static local method of say GetPhones() so that code reads a bit better?

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -0,0 +1,171 @@
...
+                }
+
+                cursor.Close();
+
+                projection = new string[3]
+                {
+                    ContactsContract.CommonDataKinds.StructuredPostal.Street,
+                    ContactsContract.CommonDataKinds.StructuredPostal.City,
+                    ContactsContract.CommonDataKinds.StructuredPostal.Postcode
+                };
```
mattleibow, line 100: I see a set-up for an address, but is this used?

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Xamarin.Essentials
+{
+    public static partial class Contacts
+    {
+        static Task<Contact> PlatformPickContactAsync()
+        {
+            var uiView = Platform.GetCurrentViewController();
+            if (uiView == null)
+                throw new ArgumentNullException($"The View Controller can't be null.");
```
mattleibow, line 18: Maybe use a more "helpful" message for the developer, like "Unable to find the current view controller." so that at least it makes more sense as to _why_ it crashed.

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            var source = new TaskCompletionSource<Contact>();
+
+            using var picker = new CNContactPickerViewController
```
mattleibow, line 22: How does this work with the GC? I recall somewhere that iOS may collect this instance, and thus close the popup.
Do we need to keep a reference to it at all?

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                foreach (var item in contact.EmailAddresses)
+                    emails.Add(new ContactEmail(item?.Value?.ToString(), contactType));
+
+                var name = string.Empty;
+
+                if (string.IsNullOrEmpty(contact.MiddleName))
+                    name = $"{contact.GivenName} {contact.FamilyName}";
+                else
+                    name = $"{contact.GivenName} {contact.MiddleName} {contact.FamilyName}";
```
mattleibow, line 56: Not so happy with this... Is it possible to have both the "display name" and each name part? Not sure how Android stores the parts, but is it possible to get it?

@Redth @pictos I am more thinking that the app displays "Last, First" or something. I think having both will add more value.

Xamarin.Essentials/Contacts/Contacts.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+    public static partial class Contacts
+    {
+        public static async Task<Contact> PickContactAsync()
+        {
+            if (DeviceInfo.Platform != DevicePlatform.iOS)
+            {
+                var permissionStatus = await Permissions.RequestAsync<Permissions.ContactsRead>();
+                if (permissionStatus != PermissionStatus.Granted)
+                    return null;
+            }
```
mattleibow, line 17: What happens on iOS when you call this?

DO we maybe want to throw if there is no permissions? I think we usually use `EnsureGrantedAsync` instead and the dev can catch and handle. I think null means canceled?

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -0,0 +1,130 @@
...
+                            {
+                                var first = recordName.Get<string>(TizenName.First) ?? string.Empty;
+                                var last = recordName.Get<string>(TizenName.Last) ?? string.Empty;
+
+                                name = $"{first} {last}".Trim();
+                            }
+
+                            var emailCount = record.GetChildRecordCount(TizenContact.Email);
+                            if (emailCount > 0)
+                            {
```
mattleibow, line 60: this if is _technically_ not necessary... But no real issue.

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -0,0 +1,130 @@
...
+                                for (var i = 0; i < emailCount; i++)
+                                {
+                                    var item = record.GetChildRecord(TizenContact.Email, i);
+                                    var addr = item.Get<string>(TizenEmail.Address);
+                                    var type = item.Get<int>(TizenEmail.Type);
+
+                                    var emailType = TizenEmail.Types.Other;
+                                    if (Enum.IsDefined(typeof(TizenEmail.Types), type))
+                                        emailType = (TizenEmail.Types)type;
```
mattleibow, line 69: Took me a few reads to see that this was basically saying use other if the type is invalid. But, since this is an enum, what are the possible values? Maybe let the `GetContactType` actually do this? If `type` is int, then you can still just cast it to an enum.

```csharp
enum Test {
  A = 0,
  B = 1,
}

int v = 1000;
Test res = (Test)v;
```

that is still valid C# code and then the `_ => ` in the switch will catch it.

Xamarin.Essentials/Contacts/Contacts.uwp.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            try
+            {
+                var contactSelected = await contactPicker.PickContactAsync();
+
+                if (contactSelected == null)
+                    return null;
+
+                var contactManager = await ContactManager.RequestStoreAsync();
+                var contact = await contactManager.FindContactsAsync(contactSelected.Name);
```
mattleibow, line 24: Why do we need the ContactManager? We already have the contact.

Looking at the example in the docs, it appears they just use the contact, or optionally reduce the results by setting the fields: https://docs.microsoft.com/en-us/windows/uwp/contacts-and-calendar/selecting-contacts#complete-example-single-contact

mattleibow (commented):
Overall, looks very nice.

I just got a suggestion for the structure of the Android code. A question about the GC on iOS. Tizen has a one-liner change. 

UWP seems a bit different, I got a comment which will explain a bit more. 

But, I think the API is nice, the flow is pretty decent and just questions/tweaks.

Xamarin.Essentials/Contacts/Contacts.ios.cs
Redth: Yeah I think display name is more about how you have setup the contacts app on the phone to display the names, and it isn't _really_ appropriate to use that here.  I think this is fine how it is, although part of me wonders if we should break up the Name for the contact into First/Middle/Last properties...

Xamarin.Essentials/Contacts/Contacts.shared.cs
Redth: I was thinking of denying permissions as being cancelled here.  I guess we should review what we do for things like the MediaPicker when permission is denied...

Xamarin.Essentials/Contacts/Contacts.android.cs
Redth: This wasn't / isn't merged yet but making the change now

Xamarin.Essentials/Contacts/Contacts.shared.cs
Redth: iOS BTW does not require permissions for the picker: https://developer.apple.com/documentation/contactsui/cncontactpickerviewcontroller

Looking at MediaPicker, we do throw if not granted (EnsureDeclared) will change this for consistency.

Xamarin.Essentials/Contacts/Contacts.ios.cs
mattleibow: That might be nice. They may want to show initials in a circle as well.

Xamarin.Essentials/Contacts/Contacts.uwp.cs
Redth: Tested works fine without the manager and finding it... changed.

Xamarin.Essentials/Contacts/Contacts.ios.cs
Redth: Leaving as single Name for now.  We can always add First/Last later if we want, and still keep Name as a combined full name field without breaking API.

Xamarin.Essentials/Contacts/Contacts.android.cs
Redth: @pictos implemented this.

PR #1362, opened 2020-08-25
Adding an IsSupported property

mattleibow:
### Description of Change ###

Adding an `IsSupported` property to `MediaPicker` and `Screenshot`.

I also moved the check for Android to the actual inline logic because this is a window flag that can change at any time, so rather check when using it. This is also not available when navigating, so any view setup logic will be incorrect.

### Bugs Fixed ###

- Related to issue #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Screenshot/Screenshot.android.cs
+++ b/Xamarin.Essentials/Screenshot/Screenshot.android.cs
@@ -14,7 +14,7 @@ namespace Xamarin.Essentials
         static Task<ScreenshotResult> PlatformCaptureAsync()
         {
             if (Platform.WindowManager?.DefaultDisplay?.Flags.HasFlag(DisplayFlags.Secure) == true)
-                throw new FeatureNotSupportedException("Unable to take a screenshot of a secure window.");
+                throw new UnauthorizedAccessException("Unable to take a screenshot of a secure window.");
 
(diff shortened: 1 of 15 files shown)


Xamarin.Essentials/Screenshot/Screenshot.android.cs
```diff
@@ -8,11 +8,14 @@ namespace Xamarin.Essentials
...
     {
-        static bool PlatformCanCapture =>
-            Platform.WindowManager.DefaultDisplay?.Flags.HasFlag(DisplayFlags.Secure) == false;
+        static bool PlatformIsCaptureSupported =>
+            true;
 
         static Task<ScreenshotResult> PlatformCaptureAsync()
         {
+            if (Platform.WindowManager?.DefaultDisplay?.Flags.HasFlag(DisplayFlags.Secure) == true)
+                throw new FeatureNotSupportedException("Unable to take a screenshot of a secure window.");
```
Redth, line 17: Would this make sense to be something like `UnauthorizedAccessException` instead?

mattleibow: Oh yeah. That is much better.

PR #1367, opened 2020-09-01
Make forms app static

Redth:
### Description of Change ###

Stop recreating the app instance if the xamarin app's runtime is still active.

### DIFF
--- a/Samples/Samples.Android/MainActivity.cs
+++ b/Samples/Samples.Android/MainActivity.cs
@@ -14,6 +14,8 @@ namespace Samples.Droid
         Categories = new[] { Intent.CategoryDefault })]
     public class MainActivity : Xamarin.Forms.Platform.Android.FormsAppCompatActivity
     {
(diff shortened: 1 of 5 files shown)


jamesmontemagno:
Interesting.... does this seem to be an issue in some places?

Redth:
@jamesmontemagno so on Android, if you hit 'back' enough, your activity will be finished and destroyed, but the xamarin runtime will still be alive.  What happens when you go back into the app is, you will get a new App() created even though you had a perfectly good one around before when you hit the back button.  Using a static instance will reuse/share this Forms app instance.  Think about the case where maybe you have a tabview and you were on the second tab, when you come back and the activity is recreated your App would get recreated and you'd lose your 'state' in a sense.  With the static app instance you can retain this 'state'.

Hopefully that makes sense.  This is probably largely unnecessary on the other platforms.

jamesmontemagno:
fancy! <3

Issue #1376, opened 2020-09-08
On iOS Launcher.OpenAsync presents Open-In/ShareSheet Menu but closes prematurely or is unresponsive after clicking on one of the options


fadihrezq:
When trying to open a file I downloaded into my app's Documents folder, using the Launcher.OpenAsync, it presents the "Open In Menu" (ShareSheet menu) but clicking on any of the options does not do anything or sometimes just closes the menu.

Launcher.OpenAsync works normally in Android but this issue appears in iOS.

After several days of trying different things I finally created my own DependencyInjection iOS implementation of the Launcher.OpenAsync code (copied from the repo here) and after tinkering around with it, I found that if I keep a reference of the UIDocumentInteractionController and a reference of the Launcher implementation, then it works.

My guess is that the UIDocumentInteractionController still needs to be alive when the user clicks the icon of whatever app needs to handle the file. In the Xamarin.Essentials Launcher code, the UIDocumentInteractionController is instantiated as a local variable in the OpenAsync method and after it presents the "Open In Menu",  the function returns and the interaction controller goes out of scope. So when the user tries to interact with the menu, the interaction controller reference is no longer valid and this is what causes the issue.

Here is a sample of my code:

In the iOS implementation:
```
public class IOSShareSheetManager : IShareSheetManager
{
     UIDocumentInteractionController controller = null;
     public Task Open(OpenFileRequest request)
     {
           var fileUrl = NSUrl.FromFilename(request.File.FullPath);

           controller = UIDocumentInteractionController.FromUrl(fileUrl);
           controller.Uti = request.File.ContentType;

          ....

          controller.PresentOpenInMenu(rect.Value, vc.View, true);

          return Task.CompletedTask;
     }
}
```

shared code usage:
```
if (shareSheetManager == null)
     shareSheetManager = DependencyService.Get<IShareSheetManager>();
await shareSheetManager.Open(fileRequest);
```

I also believe this also may be related to the following issues: 

 #https://github.com/xamarin/Essentials/issues/1166
 #https://github.com/xamarin/Essentials/issues/1348

I am using:
Xamarin.Forms v4.7
Xamarin.Essentials 1.5.3.2

Redth:
Should be fixed in #1380

Aguirrehub25:
you can implement the classes in visual studio 2019 ? 
I have the same problem displaying images on ios

Issue #1373, opened 2020-09-03
[Bug] Permissions.RequestAsync<Permissions.ContactsRead>() raises System.ObjectDisposedException


nbsoftware:
Permissions.RequestAsync<Permissions.ContactsRead> was working just fine until iOS 13.6

I have upgraded XCode to 11.7, built my app against latest 13.7 SDK and I got the exception System.ObjectDisposedException below, either on the Simulator or a physical device upgraded to 13.7

Tested again on an iOS 13.6 Simulator, works just fine as expected. 

```
System.ObjectDisposedException: Cannot access a disposed object.
  at AddressBook.ABAddressBook.AssertValid () [0x00012] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.22.1.8/src/Xamarin.iOS/AddressBook/ABAddressBook.cs:171 
  at AddressBook.ABAddressBook.get_Handle () [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.22.1.8/src/Xamarin.iOS/AddressBook/ABAddressBook.cs:176 
  at AddressBook.ABAddressBook.RequestAccess (System.Action`2[T1,T2] onCompleted) [0x00027] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.22.1.8/src/Xamarin.iOS/AddressBook/ABAddressBook.cs:204 
  at Xamarin.Essentials.Permissions+ContactsRead.RequestAddressBookPermission () [0x00016] in d:\a\1\s\Xamarin.Essentials\Permissions\Permissions.ios.cs:112 
  at Xamarin.Essentials.Permissions+ContactsRead.RequestAsync () [0x0001d] in d:\a\1\s\Xamarin.Essentials\Permissions\Permissions.ios.cs:91 
  at Xamarin.Essentials.Permissions.RequestAsync[TPermission] () [0x00000] in d:\a\1\s\Xamarin.Essentials\Permissions\Permissions.shared.cs:13 
  at fipacapp.Misc.PermissionsHelpers+<>c.<CheckAddressBookPermissions>b__1_0 () [0x0000f] 
...
```

Not too sure if this is a Xamarin.Essentials bug or a Xamarin.iOS one... so I'll post it to Xamarin.iOS too

### Basic Information

```
Visual Studio Community 2019 for Mac (Preview)
Version 8.8 Preview (8.8 build 1532)
Installation UUID: be4088b1-069a-44b1-bca1-a8bc61ab68dc
[...]

PR #1380, opened 2020-09-14
iOS Keep things in scope

Redth:
### Description of Change ###

There are a couple of issues related to what I believe could be do to something going out of scope and being GC'd before the native action is complete.

This PR just keeps some instances around (and attempts to clean them up when finished).

### Bugs Fixed ###

- #1376
- #1373


mattleibow (approved):
LGTM

mattleibow:
Had to update the iOS simulator version because the bots are updated

Issue #1383, opened 2020-09-17
[Bug] problem with Colors on macOS


tipa:
First of all, thanks for finally adding macOS support - I've been waiting for it for months!!
I installed v1.6.0-pre1

Unfortunately there seems to be some issues when working with `Color`, it seems to fail to resolve the System.Drawing.Color type.
For example, this cannot compile:`NSColor.Red.ToSystemColor();`
Same problem with `ColorConverters.FromHex("#fff");`

`.../ViewController.cs(13,13): Error CS0012: The type 'Color' is defined in an assembly that is not referenced. You must add a reference to assembly 'OpenTK, Version=0.0.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065'. (CS0012) (Test2)
`

Redth:
If you add a reference to OpenTK it should work.

We need to look at seeing if we can have the nuget pull this in.

mattleibow:
I thought this was moved to another assembly in VS2019...

mattleibow:
@tipa let me know if adding a reference does actually fix it. Make sure to use the local opentk reference in the references window, and not the NuGet.

tipa:
Yes this does fix it - thanks!
By the way, looking at the .csproj file I see that `PackageReference` is being used to include `System.Numerics.Vectors` on iOS, tvOS, watchOS but `Reference` is being used for macOS. Not sure if that's of importance...

tipa:
I found another problem. This code

`NSColor.ControlAccentColor.ToSystenmColor()`

throws `NSInvalidArgumentException: *** -getRed:green:blue:alpha: not valid for the NSColor Catalog color: System controlAccentColor; need to first convert colorspace.`

I think this line needs to be added in the `ToSystemColor` method:
`color = color.UsingColorSpace(NSColorSpace.GenericRGBColorSpace);`

Redth:
I'm not too up on color spaces, but I bet @mattleibow is with SkiaSharp...  Is this the right thing to do here do you think @mattleibow  ?  GenericRGBColorSpace?

mattleibow:
Should it not be the sRGB color space?

tipa:
Good question. I am no color expert either. Not sure why I chose to use `NSColorSpace.GenericRGBColorSpace`, I think sRGB would work, too

mattleibow:
Thanks for reporting. Pushed to NuGet. Let us know how it goes!

tipa:
[...]

PR #1385, opened 2020-09-17
Include the OpenTK reference in macOS for System.Drawing.Color

Redth:
### Description of Change ###

Adds a reference to OpenTK in the macOS project which should make it through to the nuget package.

### Bugs Fixed ###

- #1383


mattleibow (approved):
I believe this is the correct option.

PR #1405, opened 2020-09-30
[Contacts] Adding GetAll and Refactoring on Contacts

### LINKED ISSUE #1403: [Enhancement] Contacts Get All
There is already a method for getting one item from the contact list ([Pull Request](https://github.com/xamarin/Essentials/pull/1349))
But there is no method for getting all the contact list
I have drafts that will help solve this issue on Android, iOS and UWP

 **API**
`Task<List<Contact>> Contacts.GetAll()`

**Difficulty : [low]**

dimonovdd:
### Description of Change ###
There is already a method for getting one item from the contact list ([Pull Request](https://github.com/xamarin/Essentials/pull/1349))

- These changes implement getting all contacts for the following platforms: (Android, iOS, UWP, MacOS, Tizen)
- removed ContactType
- and small fixes

### Issue Closed ###
- Fixed #1403

### API Changes ###
**Added:** 
```csharp
//cancellationToken is only used on UWP
public static Task<IEnumerable<Contact>> Contacts.GetAllAsync(CancellationToken cancellationToken = default)
```

**Removed:**
```csharp
enum ContactType
```

**Changed:**
```csharp
    public class Contact
    {

        public string Id { get; private set; }

        public string DisplayName { get; private set; }

        public string NamePrefix { get; set; }

        public string GivenName { get; set; }

        public string MiddleName { get; set; }

        public string FamilyName { get; set; }

        public string NameSuffix { get; set; }

        public List<ContactPhone> Phones { get; set; }

        public List<ContactEmail> Emails { get; set; }

        public override string ToString() => DisplayName;
    }
```
```csharp
    public class ContactEmail
    {
        public string EmailAddress { get; set; }

        public override string ToString() => EmailAddress;
    }
```
```csharp
    public class ContactPhone
    {
        public string PhoneNumber { get; set; }

        public override string ToString() => PhoneNumber;
    }
```

### PR Checklist ###

- [x] Has tests 
- [x] Has samples 
- [x] Updated documentation


pictos:
@dimonovdd maybe this can help you
https://github.com/xamarin/Essentials/blob/809fbb64b0d4d9b513bc2bc22ba26ea1298651c3/Xamarin.Essentials/Contacts/Contact.android.cs#L44

dimonovdd:
> @dimonovdd maybe this can help you
> https://github.com/xamarin/Essentials/blob/809fbb64b0d4d9b513bc2bc22ba26ea1298651c3/Xamarin.Essentials/Contacts/Contact.android.cs#L44

I have my old code, but on Android it doesn't match yours. I don't like duplicating code.

I still want to make an implementation for MacOS. 
Help is most needed with Tizen

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -22,6 +22,29 @@ static async Task<Contact> PlatformPickContactAsync()
             return null;
         }
 
+        static async Task<IEnumerable<Contact>> PlatformGetAllAsync()
+        {
+            await Task.CompletedTask;
+            return GetAll();
```
mattleibow, line 28: You can also do `Task.FromResult(...)` I think. Then no need for the async/await part at all.

dimonovdd: this is a draft

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -27,6 +29,46 @@ static Task<Contact> PlatformPickContactAsync()
...
+            using (var store = new CNContactStore())
+            {
+                var containers = store.GetContainers(null, out var error);
+
+                foreach (var container in containers)
+                {
+                    try
+                    {
+                        using var pred = CNContact.GetPredicateForContactsInContainer(container.Identifier);
+                        results.AddRange(store.GetUnifiedContacts(pred, keys, out error)?.Select(a => GetContact(a)));
```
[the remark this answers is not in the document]
dimonovdd: We need to see if this is possible on iOS, UWP, MacOS and Tizen.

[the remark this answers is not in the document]
dimonovdd: @mattleibow 
First we need to write a draft implementation for all platforms

Xamarin.Essentials/Contacts/Contacts.android.cs
mattleibow: Oh yeah. Just throwing comments out for discussion here. Thanks for getting started on this.

Xamarin.Essentials/Contacts/Contacts.ios.cs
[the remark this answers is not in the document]
mattleibow: Of course. Just wondering if it will even be possible... I know UWP likes to have awaits everywhere. That one might be the real one to cause any issues...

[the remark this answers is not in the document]
dimonovdd: There are TasCompletionSource for emergencies

dimonovdd:
Help me!
When I include a "Contacts.macos.cs" file in a project, Visual Studio (Mac OS, Windows) perceives it as a "Contacts.netstandard.tvos.watchos.cs" file

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -124,5 +127,10 @@
   <ItemGroup>
     <None Remove="mdoc.targets" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="Contacts\Contacts.netstandard.tvos.watchos.cs">
+      <SubType></SubType>
+    </Compile>
+  </ItemGroup>
```
mattleibow, line 134: VS borked the csproj :(

mattleibow: Usually when working with files with multi-targeting, I do it in VSCode / explorer. VS can't figure things out correctly with all the conditions.

mattleibow: Check out the diff and then just revert the csproj.

dimonovdd: I delete it and it reappears

mattleibow: Make sure to close VS. There are 2 changes to revert in there.

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -54,6 +54,9 @@
     <PackageReference Include="Xamarin.Build.TypeRedirector" Version="0.1.2-preview" PrivateAssets="all" />
     <Compile Include="**\*.shared.cs" />
     <Compile Include="**\*.shared.*.cs" />
+    <Compile Include="Contacts\Contacts.macos.cs">
+      <SubType></SubType>
+    </Compile>
```
mattleibow, line 59: Here is bad boy part 2

dimonovdd: I removed this lines on VS Windows, compiled Android and transferred project to my MacBook and it started working. Magic

Thanks

dimonovdd:
@mattleibow 
I made drafts for Android, iOS, UWP and MacOS. 
I could use some help with Tizen

mattleibow:
Awesome! I'll take a look! @rookiejava or @sung-su can probably help out with Tizen until I get to that!

dimonovdd:
@mattleibow just don't forget that this is still a draft, and it has ugly code

pictos:
I just tested here, and we need to think in a pagination thing. On my Android device take a while to get all contacts, something like 4 seconds (I have 141 contacts). And the searching isn't in a background thread, so the UI is blocked during the `GetAllContacts` processing.
If you have a mobile device with something like 500 contacts (and it's easy to find) the experience wouldn't be good, even if we are processing in a background thread.

mattleibow:
No worries. Until this is out of draft, it will remain a live mind dump 😁

dimonovdd:
@mattleibow @pictos 
Apparently, Android is the only place where we can use `yield`
And UWP in any case wants us to `await`
I'm afraid that the best we can do for all the platforms is to return `Task<IEnumerable<Contact>`
Nevertheless this is better than nothing
Maybe anybody else has ideas?

pictos:
@dimonovdd I need to play around, but I think that you can use [IAsyncEnumerable<T>](https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8)

dimonovdd:
> @dimonovdd I need to play around, but I think that you can use [IAsyncEnumerable](https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8)

I'll try this on the weekend

mattleibow:
You can use a yield I think if you create a new method that is just IEnumerable. Die we have async per item, or just the initial connection.

The task can be fromresult (enummethod)

Will iasyncenumerable restrict platforms? Is it supported on UWP and netstandard2.0

dimonovdd:
> You can use a yield I think if you create a new method that is just IEnumerable

@mattleibow what will happen to UWP ?

dimonovdd:
@pictos  > but I think that you can use [IAsyncEnumerable](https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8)

@mattleibow > Will iasyncenumerable restrict platforms? Is it supported on UWP

Yes, UWP out of the box does not support IAsyncEnumerable

dimonovdd:
I tried several different options. But UWP breaks everything.
There were no similar tasks in Essentials?

pictos:
@dimonovdd can you push your uwp code? Even if it's broken, that whey we can see what's going on.

dimonovdd:
> @dimonovdd can you push your uwp code? Even if it's broken, that whey we can see what's going on.

You can view [this file](https://github.com/xamarin/Essentials/blob/d9eb03c7572dc2f91bf5c20256c26338d335889f/Xamarin.Essentials/Contacts/Contacts.uwp.cs). The PlatformGetAllAsync method on UWP must be asynchronous.

mattleibow:
I see... Is there a nuget package for iasyncenumerable? Does that work on UWP?

Also, I am just thinking, do we want a flat list or some paging, some more advanced lookup? Maybe have an offset/count type of lookup as well?

If I load a page for the user, and then want to say get a bit, I will have to basically load all the items before.

pictos:
> I see... Is there a nuget package for iasyncenumerable? Does that work on UWP?

@mattleibow I found this one, I didn't have a chance to test yet, maybe today night...
https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/

Samples/Samples/ViewModel/ContactsViewModel.cs
```diff
@@ -91,29 +91,38 @@ async void OnGetContact()
...
+                if (tasks != null)
                 {
-                    foreach (var a in collect)
-                        MainThread.BeginInvokeOnMainThread(() => ContactsList.Add(a));
-                });
+                    foreach (var task in tasks)
+                    {
+                        Task.Run(async () =>
+                        {
+                             foreach (var contact in await task)
```
dimonovdd, line 115: @mattleibow @pictos 
I found a way for all platforms, but its use doesn't look very nice

mattleibow: That is not very nice indeed...

What if instead of an IEnumerable result, we return some contacts result object?

Then this can have the actual work/paging/enumeration.

Thus, the contacts API really returns the iterator object. With think like hasmore and getnextasync and stuff like that?

mattleibow: Maybe we can even have the iasyncenumerable on the playforms that support it... Everything bu UWP maybe

dimonovdd: we only need IAsyncEnumerable for uwp. But IAsyncEnumerable only works with C# version 8.0

dimonovdd: and I don't think giving up UWP is a good way

mattleibow: Oh yeah. I mean can we have a class that conditionally implements the interface? I have used it in the past, but I am not sure if it supports an actual implementation.

Then we can either use await foreach on supporting platforms, and some nice GetNextAsync members on UWP.

pictos:
> > I see... Is there a nuget package for iasyncenumerable? Does that work on UWP?
> 
> @mattleibow I found this one, I didn't have a chance to test yet, maybe today night...
> https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/

I don't know if I missed something. But adding this package I was able to use IAsyncEnumarable in UWP. At least the compiler doesn't complain. But we will need to add one more ItemGroup in csproj to add this package for the .netStandard2.0. 
Maybe doing this can be easier than implement it from zero.

I did a silly implementation, just to see if the compiler will be happy.
image

mattleibow:
Ah, that is nice then.

dimonovdd:
@pictos This is great, but I was hoping we could do it without the extra dependency

dimonovdd:
@pictos what did you do with the Tizen and .NETStandard 1.0?

image

pictos:
As I said, for `.netstandard1.0` we can't support. Now Tizen I didn't notice that

dimonovdd:
@maybe something is wrong with my VS, 
I'll try to check Tizen tomorrow.

pictos:
@dimonovdd, to be honest, I didn't run the Tizen project. But if it's a blocker I'll try to find my old implementation.

sung-su:
`IAsyncEnumerable<T>` is available from `.NET Standard 2.1` or `.Net Core 3.0`.
There are supported from `tizen60` tfm on the `Tizen` platform.

dimonovdd:
@pictos thank you very much
do you think it's better to use 
```csharp
await Task.CompletedTask;
```
or 
```csharp
yield return await Task.FromResult(contact);
```
in the PlatformGetAllAsync method on ios, macos, android, tizen

@sung-su Please check if this works on tizen

sung-su:
@dimonovdd `IAsyncEnumerable<Contact> PlatformGetAllAsync()` method works well on `Tizen`.

dimonovdd:
@pictos @mattleibow 
I noticed that the `ContactType` property in the `Contact` class is only used on iOS and macOS. 
The `Contact Type` property in the `ContactEmail` and `ContactPhone` classes on ios and macOS is defined incorrectly

image
image

dimonovdd:
I made a [draft](https://github.com/dimonovdd/Essentials/blob/featureCotactsTypeConverter/Xamarin.Essentials/Types/Contact.shared.cs). you can view it.

dimonovdd:
Do you have any thoughts about this pull request? I would like to finish it, but I don't know in which direction to move

Xamarin.Essentials/Contacts/Contacts.macos.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Contacts;
+
+namespace Xamarin.Essentials
+{
+    public static partial class Contacts
+    {
+        static Task<Contact> PlatformPickContactAsync() => throw ExceptionUtils.NotSupportedOrImplementedException;
```
mattleibow, line 11: Maybe this file should be merged with ios? It is just the picker that macos does not have, but the rest of the read is using the same APIs. And probably the write.

We can `#if __IOS__` the picker code...

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -36,56 +38,64 @@ static async Task<Contact> PlatformPickContactAsync()
...
+
+        static Contact ToContact(ContactsRecord contactsRecord)
+        {
+            var name = string.Empty;
+            var record = contactsRecord.GetChildRecord(TizenContact.Name, 0);
+            if (record != null)
+            {
+                var first = record.Get<string>(TizenName.First) ?? string.Empty;
+                var last = record.Get<string>(TizenName.Last) ?? string.Empty;
+                name = $"{first} {last}".Trim();
```
mattleibow, line 71: Can we maybe do something different?

```
if (first)
  if (last)
    first + last
  else 
    first
else
  if (last)
    last
```

The we don't have to allocate strings and then trim. Not perfect, but .NET needs a way to do this...

mattleibow (commented):
Looking at the code, I think it is good in the sense the logic seems right.
I just have a few nitpicks with somethings, but mostly good.

One thing I do want to talk about is the IAsyncEnumerable and the cancellation process. For most platforms, this is synchronous, but that doesn't mean we can do it better. I see there is an attribute `[EnumeratorCancellation]` that we use as a parameter of the method:

```
async IAsyncEnumerable<T> DoThing(..., [EnumeratorCancellation] CancellationToken cancellationToken = default)
```

We can use that to actually provide a way to stop requesting contacts in an async way or to be consistent when something is running and then it is cancelled.

Se more: https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -17,100 +18,99 @@ static async Task<Contact> PlatformPickContactAsync()
...
+            var name = cursor.GetString(cursor.GetColumnIndex(ContactsContract.Contacts.InterfaceConsts.DisplayName));
+            var idQ = new string[1] { cursor.GetString(cursor.GetColumnIndex(idKey)) };
+            var phones = GetNumbers(context, idQ)?.Select(a => new ContactPhone(a.data, GetPhoneContactType(a.type)))?.ToList();
+            var emails = GetEmails(context, idQ)?.Select(a => new ContactEmail(a.data, GetEmailContactType(a.type)))?.ToList();
+            ContactType type;
+
+            if (typeKey != null)
+                type = GetPhoneContactType(cursor.GetString(cursor.GetColumnIndex(typeKey)));
+            else
+                type = phones?.FirstOrDefault()?.ContactType ?? ContactType.Unknown;
```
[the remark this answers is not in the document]
dimonovdd: this is changed in the new branch

DeviceTests/DeviceTests.Shared/Contacts_Tests.cs
```diff
@@ -0,0 +1,27 @@
...
+            {
+                await Permissions.RequestAsync<Permissions.ContactsRead>();
+            });
+
+            var list = new List<Xamarin.Essentials.Contact>();
+            var contacts = Xamarin.Essentials.Contacts.GetAllAsync();
+
+            await foreach (var contact in contacts)
+                list?.Add(contact);
+        }
```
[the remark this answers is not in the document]
dimonovdd: I don't think it's doing any harm

Samples/Samples/ViewModel/ContactsViewModel.cs
```diff
@@ -80,5 +90,35 @@ async void OnGetContact()
...
+            IsBusy = true;
+            try
+            {
+                var contacts = Contacts.GetAllAsync();
+
+                _ = Task.Run(async () =>
+                      {
+                          await foreach (var contact in contacts)
+                              MainThread.BeginInvokeOnMainThread(() => ContactsList.Add(contact));
+                      });
```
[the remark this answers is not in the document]
dimonovdd: this is the only way we get each contact in turn on the screen

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -19,44 +20,64 @@ static Task<Contact> PlatformPickContactAsync()
...
+                var contactType = ToPhoneContact(contact.ContactType);
+                var phones = contact.PhoneNumbers?.Select(item => new ContactPhone(item?.Value?.StringValue, contactType))?.ToList();
+                var emails = contact.EmailAddresses?.Select(item => new ContactEmail(item?.Value?.ToString(), contactType))?.ToList();
+                var name = $"{contact.NamePrefix} {contact.GivenName} {contact.MiddleName} {contact.FamilyName} {contact.NameSuffix}"
+                    .Replace("  ", " ").TrimEnd();
 
                 return new Contact(name, phones, emails, contactType);
             }
-            catch (Exception)
+            catch (Exception ex)
```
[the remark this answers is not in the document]
dimonovdd: I also like `using`

Can we use this with `foreach`?

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -17,100 +18,99 @@ static async Task<Contact> PlatformPickContactAsync()
...
 
-        internal static Contact PlatformGetContacts(Net.Uri contactUri)
+        static async IAsyncEnumerable<Contact> PlatformGetAllAsync()
         {
-            if (contactUri == null)
-                return default;
-
+            await Task.CompletedTask;
             using var context = Platform.AppContext.ContentResolver;
+            using var cursor = context.Query(ContactsContract.Contacts.ContentUri, null, null, null, null);
```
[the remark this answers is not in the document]
dimonovdd: this is done on the method `ReadCursorItems`

Xamarin.Essentials/Contacts/Contacts.android.cs
```diff
@@ -17,100 +18,99 @@ static async Task<Contact> PlatformPickContactAsync()
...
             return null;
         }
 
-        internal static Contact PlatformGetContacts(Net.Uri contactUri)
+        static async IAsyncEnumerable<Contact> PlatformGetAllAsync()
         {
-            if (contactUri == null)
-                return default;
-
+            await Task.CompletedTask;
```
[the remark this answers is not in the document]
dimonovdd: `yield return await Task.FromResult(contact)`

Xamarin.Essentials/Contacts/Contacts.ios.cs
```diff
@@ -19,44 +20,64 @@ static Task<Contact> PlatformPickContactAsync()
...
-                    name = $"{contact.GivenName} {contact.FamilyName}";
-                else
-                    name = $"{contact.GivenName} {contact.MiddleName} {contact.FamilyName}";
+            try
+            {
+                var contactType = ToPhoneContact(contact.ContactType);
+                var phones = contact.PhoneNumbers?.Select(item => new ContactPhone(item?.Value?.StringValue, contactType))?.ToList();
+                var emails = contact.EmailAddresses?.Select(item => new ContactEmail(item?.Value?.ToString(), contactType))?.ToList();
+                var name = $"{contact.NamePrefix} {contact.GivenName} {contact.MiddleName} {contact.FamilyName} {contact.NameSuffix}"
+                    .Replace("  ", " ").TrimEnd();
```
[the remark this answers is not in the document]
dimonovdd: I think it's senseless

Xamarin.Essentials/Contacts/Contacts.tizen.cs
dimonovdd:
```csharp
if (!string.IsNullOrWhiteSpace(first))
{
     if (!string.IsNullOrWhiteSpace(last))
         name = first + last;
     else
        name = first;
}
else
{
    if (!string.IsNullOrWhiteSpace(last))
        name = last;
}
```

it's not pretty

pictos: maybe... @dimonovdd 
```csharp
var isFirstValid = !string.IsNullOrWhiteSpace(first);
var isLastValid = !string.IsNullOrWhiteSpace(last);
if (isFirstValid)
{
      name = isLastValid ? first + last : first;
}
else
{
    if (isLastValid)
        name = last;
}
```

dimonovdd:
> async IAsyncEnumerable<T> DoThing(..., [EnumeratorCancellation] CancellationToken cancellationToken = default)

@mattleibow How do we use `CancellationToken` on iOS, macOS, Tizen, Droid?

Xamarin.Essentials/Contacts/Contacts.tizen.cs
mattleibow: We need the space, but maybe we could use string.Join...

mattleibow:
@dimonovdd we could actually use it in the loop: `CancellationToken.ThrowIfCancellationRequested()`

This will be useful if we - just like in the sample - are iterating on a different thread.

Xamarin.Essentials/Contacts/Contacts.tizen.cs
dimonovdd:
```csharp
string[] names = { contact.NamePrefix, contact.GivenName, contact.MiddleName, contact.FamilyName, contact.NameSuffix };
string displayName = string.Join(' ', names.Where(a => !string.IsNullOrWhiteSpace(a)));
```

dimonovdd:
@mattleibow  and I thought a lot

There were a few questions:

1. Can  [`ContactEmail`](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Types/Contact.shared.cs#L23) and [`ContactPhone`](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Types/Contact.shared.cs#L36) be structure?
@Redth [corrected this in the old PR](https://github.com/xamarin/Essentials/pull/1349/commits/6147e6c2275c9960e2110f2c9bdaa3bddb770678#diff-8e70416a843b8a75e17dd34230b83b6b059de855e6a187305621c6a96ff8bc2bL88)

2. [`ContactType`](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Contacts/Contacts.shared.cs#L29) has nothing to do with the "contact".  We need a correct name.

3. [`IAsuncEnumerable` ](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Contacts/Contacts.shared.cs#L19) is a bad way in our case.
`GetAllAsync` never actually uses the `async` on each item even in UWP
We have several options:

    a) [`Task<IEnumerable>`](https://github.com/xamarin/Essentials/blob/a860ba2ace28cf21f332321a269b41854b0831fe/Xamarin.Essentials/Contacts/Contacts.uwp.cs#L22) with using a nested function

    b) [`ContactsStore`](https://github.com/dimonovdd/Essentials/blob/a0bdaf577212b71ae6b838d2882a20b44a505420/Xamarin.Essentials/Contacts/Contacts.uwp.cs#L23) class with [`GetAll method`](https://github.com/dimonovdd/Essentials/blob/a0bdaf577212b71ae6b838d2882a20b44a505420/Xamarin.Essentials/Contacts/Contacts.uwp.cs#L82)

    c) Replace `ContactsStore` on `ContactsCollection : IEnumerable<Contact>`

4. Is there a easy way on iOS to actually get the count without iterating? 
[stackoverflow](https://stackoverflow.com/questions/36695582/cncontactstore-count-of-contact-records/43506808#43506808)

Maybe I forgot something or wrote too little.
@mattleibow  please add information if necessary

mattleibow:
@dimonovdd I think that is good, thanks.

1. @Redth  was changing the main `Contact` type to a class and I think I also suggested to change email/phone. I am in two minds about those. I think there is no harm in it being a struct, but maybe we need to sort of future proof it and use a class. Maybe we want to add more methods later. It is immutable, so a struct might just work.

2. I can't think of anything better. Everyone just uses type and some use EmailType and PhoneType, but that is not better. Maybe there is a word that the contacts people use to talk about this "property" of contacts?

3. IAsyncEnumerable is a waste, and I think I am leaning towards a) as it is simple, but then it is not very future proof. What if we want more things? For example count (as mentioned in 4.)? Since the "get all" is always going to be an expensive task, it might be worth having a new type. Going with b) is pretty nice as we can basically have that do all the work, and then cache or whatever. However we can actually do the same thing with c) and it make it simple to use. We can add properties to that collection (such as Count) as well as use LINQ. LINQ might be slow, so we could also add our own methods to it, such as Count() and do the work.

dimonovdd:
@mattleibow 

5. Rename[ `Contact.Name`](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Types/Contact.shared.cs#L16) to `Contact.DisplayName`

6. Change the implementation of getting a displayName on iOS and Tizen, if possible to do it better

7. we need to add getting Identifier on [Contact](https://github.com/dimonovdd/Essentials/blob/c79f4cb551ea65588408af889a3d8f4840b1b9b5/Xamarin.Essentials/Types/Contact.shared.cs#L7)

mattleibow:
@dimonovdd maybe we can add the various name properties to the contact - like first, last, middle - and then make "Name" be "Display Name". If the OS doesn't have a feature for display name, then we can do our own combination.

mattleibow:
@dimonovdd I think for the `Contacts.GetAllAsync` we can use ` Task<IEnumerable<Contact>>` as that is what the method says - get all the contacts.

If we ever need more things, then there will be no issue to add a `Contacts.GetContactStoreAsync` and then go from there.

dimonovdd:
@sung-su @mattleibow 
is there a MiddleName at Tizen?
image

mattleibow (commented):
This is just a very quick look. I'll do a more in depth one tomorrow as it is quite late. But... very nice so far. I think we are very close. The enums are looking like we may have done it wrong with just the Work/Personal... Can we tell the difference with Fax vs Phone? Landline vs Mobile? 

I saw you added a few more items, but it might be worth having a look at all the options by the platform and just do several separate ones. Before we make any changes, we probably need to list all the available options in a table and determine where it fits.

Samples/Samples.Android/Samples.Android.csproj
```diff
@@ -124,4 +124,9 @@
     <AndroidAsset Include="Assets\FileSystemTemplate.txt" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
+  <ProjectExtensions>
+    <VisualStudio>
+      <UserProperties TriggeredFromHotReload="False" />
+    </VisualStudio>
+  </ProjectExtensions>
```
mattleibow, line 131: Let me ask the forms team why this is now appearing in the project...

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -93,24 +98,30 @@
...
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
     <Reference Include="System.Numerics" />
     <Reference Include="OpenTK-1.0" />
   </ItemGroup>
   <ItemGroup Condition=" $(TargetFramework.StartsWith('Xamarin.WatchOS')) ">
     <Compile Include="**\*.watchos.cs" />
     <Compile Include="**\*.watchos.*.cs" />
     <PackageReference Include="System.Numerics.Vectors" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.1" />
```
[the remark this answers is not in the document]
dimonovdd: Yes I remember that. I'll delete it at the end

Xamarin.Essentials/Contacts/Contacts.shared.cs
```diff
@@ -12,12 +11,23 @@ public static async Task<Contact> PickContactAsync()
...
                 await Permissions.EnsureGrantedAsync<Permissions.ContactsRead>();
-
             return await PlatformPickContactAsync();
         }
+
+        public static Task<IEnumerable<Contact>> GetAllAsync(CancellationToken cancellationToken = default)
+            => PlatformGetAllAsync(cancellationToken);
+    }
+
+    public enum ContactPhoneType
```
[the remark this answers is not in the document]
dimonovdd: Yes I also wanted to make a table

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -13,10 +13,13 @@ namespace Xamarin.Essentials
...
     public static partial class Contacts
     {
+        static ContactsManager manager = new ContactsManager();
+
         static async Task<Contact> PlatformPickContactAsync()
         {
             Permissions.EnsureDeclared<Permissions.ContactsRead>();
-            await Permissions.EnsureGrantedAsync<Permissions.ContactsRead>();
+            Permissions.EnsureDeclared<Permissions.LaunchApp>();
+            await Permissions.RequestAsync<Permissions.ContactsRead>();
```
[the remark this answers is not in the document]
dimonovdd: I didn't pay much attention to Tizen

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -36,81 +39,93 @@ static async Task<Contact> PlatformPickContactAsync()
...
+            {
+                var emailRecord = contactsRecord.GetChildRecord(TizenContact.Email, i);
+                var addr = emailRecord.Get<string>(TizenEmail.Address);
+                var type = (TizenEmail.Types)emailRecord.Get<int>(TizenEmail.Type);
+
+                emails.Add(new ContactEmail(addr, GetContactType(type)));
+            }
+
+            return new Contact(
+                    record?.Get<string>(TizenName.ContactId),
```
sung-su, line 94: record?.Get<int>(TizenName.ContactId).ToString(),

Xamarin.Essentials/Contacts/Contacts.tizen.cs
```diff
@@ -36,81 +39,93 @@ static async Task<Contact> PlatformPickContactAsync()
...
+                    record?.Get<string>(TizenName.Prefix),
+                    record?.Get<string>(TizenName.First),
+                    record?.Get<string>(TizenName.Addition),
```
sung-su, line 97: Right, `TizenName.Addition` is the middle name.

Xamarin.Essentials/Contacts/Contacts.tizen.cs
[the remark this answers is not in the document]
sung-su: Please change`EnsureGrantedAsync`. Thank you.

Xamarin.Essentials/Contacts/Contacts.tizen.cs
dimonovdd: `TizenName.PhoneticMiddle` and  `TizenName.Addition`  threw me off my train of thought

dimonovdd:
Apart from the enumerated types from the table, the platforms can also return custom (user's) types that users can input manually

I suggest not adding EmailType and PhoneType to the release 1.6.0. Indeed, that's very confusing. I got my brain exploded while drawing up the correspondence table of the types. We can provide an incomplete API for now. It's better than doing it poorly. We'll add the types when we implement the functionality for adding contacts or a little later


image
image
[Original Table](https://yadi.sk/i/lQTyy7x3cFGXKQ)


customTypes_1

mattleibow:
@dimonovdd I think your point of the number of different types, the custom types and what we need to do for writing contacts needs to be done correctly.

After chatting with you for a bit, I think we might find some solution in having a stringly-typed enum might work. This will allow us to also write custom types.

Another side of my brain is saying that this might not be necessary since we are going for a 80/20 rule here and maybe we don't need to support custom types. However, this is very dependent on what 80% of people actually want.

With stringly-typed enums, we could also have "extension methods" that are able to "reduce" the actual number. For example, we could have a member that is `.IsWork()` or `.IsFax()` so that we don't really depend on the actual value, but that is just used to preserve the saved type. We will most likely use just the extensions to influence layout.

So... Leave comments!

dimonovdd:
> Another side of my brain is saying that this might not be necessary since we are going for a 80/20 rule here and maybe we don't need to support custom types. However, this is very dependent on what 80% of people actually want.

@mattleibow  As I have said, we'll do that as part of the 1.7.0 release

DeviceTests/DeviceTests.Shared/Contacts_Tests.cs
```diff
@@ -0,0 +1,27 @@
...
+            await MainThread.InvokeOnMainThreadAsync(async () =>
+            {
+                await Permissions.RequestAsync<Permissions.ContactsRead>();
+            });
+
+            var list = new List<Xamarin.Essentials.Contact>();
+            var contacts = Xamarin.Essentials.Contacts.GetAllAsync();
+
+            await foreach (var contact in contacts)
+                list?.Add(contact);
```
mattleibow, line 24: This needs to be changed for the new way of getting contacts.

Xamarin.Essentials/Contacts/Contacts.shared.cs
```diff
@@ -12,12 +11,23 @@ public static async Task<Contact> PickContactAsync()
...
+    {
+        Unknown = 0,
+        Personal = 1,
+        Work = 2,
+        Main = 3,
+        Mobile = 4
     }
 
-    public enum ContactType
+    public enum ContactEmailType
```
mattleibow, line 30: As discussed in the comments, we can remove this and the properties on the type for now as we decide on how to actually implement this properly. This is mainly because each platform is very different, and some platforms even support arbitrary user values.

This may mean we need a more complex type if we want to support this. 

We also probably need an issue for this.

Xamarin.Essentials/Types/Contact.shared.cs
```diff
@@ -1,55 +1,95 @@
...
-        internal ContactEmail(string email, ContactType contactType)
+        public ContactEmail(string value, ContactEmailType type)
         {
-            EmailAddress = email;
-            ContactType = contactType;
+            Value = value;
+            Type = type;
         }
+
+        public string Value { get; set; }
```
mattleibow, line 74: We should probably make this `EmailAddress` so that we can be visually explicit and more intuitive. The word "value" is very abstract.

Xamarin.Essentials/Types/Contact.shared.cs
```diff
@@ -1,55 +1,95 @@
...
-        internal ContactPhone(string phoneNumber, ContactType contactType)
+        public ContactPhone(string value, ContactPhoneType type)
         {
-            PhoneNumber = phoneNumber;
-            ContactType = contactType;
+            Value = value;
+            Type = type;
         }
+
+        public string Value { get; set; }
```
mattleibow, line 91: Similar to `ContactEmail`, we should go with `PhoneNumber` instead.

Xamarin.Essentials/Types/Contact.shared.cs
```diff
@@ -1,55 +1,95 @@
...
         {
-            EmailAddress = email;
-            ContactType = contactType;
+            Value = value;
+            Type = type;
         }
+
+        public string Value { get; set; }
+
+        public ContactEmailType Type { get; set; }
```
mattleibow, line 76: As noted in a comment above, removing this for v1.6 is probably best and we can add it back via demand or when doing writing.

mattleibow (commented):
Just a few changes as discussed in various chats and mediums:
 - remove the "contact type" concept for now
 - rename the `Value` properties to be more explicit

dimonovdd:
@mattleibow 
I saved the changes related to types in this branch
https://github.com/dimonovdd/Essentials/tree/draftContactsTypes

dimonovdd:
@mattleibow 
I also updated the [documentation](https://github.com/MicrosoftDocs/xamarin-docs/pull/3152)

PR #1406, opened 2020-10-01
[Calendar] Write features

nickrandolph:
### Description of Change ###

Adds the ability to write (create, update, delete) calendars and events

### Bugs Fixed ###

- Related to issue #

### API Changes ###

Added: 
 
Changed:

 
```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Calendars/Calendars.shared.cs
+++ b/Xamarin.Essentials/Calendars/Calendars.shared.cs
@@ -47,6 +47,24 @@ namespace Xamarin.Essentials
             return await PlatformGetEventAsync(eventId);
         }
 
+        public static Task<CalendarEvent> GetEventInstanceAsync(string eventId, DateTimeOffset instanceDate) => PlatformGetEventInstanceAsync(eventId, instanceDate);
+
+        public static Task<string> CreateCalendarEvent(CalendarEvent newEvent) => PlatformCreateCalendarEvent(newEvent);
+
+        public static Task<bool> UpdateCalendarEvent(CalendarEvent eventToUpdate) => PlatformUpdateCalendarEvent(eventToUpdate);
+
+        public static Task<bool> SetEventRecurrenceEndDate(string eventId, DateTimeOffset recurrenceEndDate) => PlatformSetEventRecurrenceEndDate(eventId, recurrenceEndDate);
+
+        public static Task<bool> DeleteCalendarEventInstanceByDate(string eventId, string calendarId, DateTimeOffset dateOfInstanceUtc) => PlatformDeleteCalendarEventInstanceByDate(eventId, calendarId, dateOfInstanceUtc);
+
+        public static Task<bool> DeleteCalendarEvent(string eventId, string calendarId) => PlatformDeleteCalendarEvent(eventId, calendarId);
+
+        public static Task<string> CreateCalendar(Calendar newCalendar) => PlatformCreateCalendar(newCalendar);
+
+        public static Task<bool> AddAttendeeToEvent(CalendarEventAttendee newAttendee, string eventId) => PlatformAddAttendeeToEvent(newAttendee, eventId);
+
+        public static Task<bool> RemoveAttendeeFromEvent(CalendarEventAttendee newAttendee, string eventId) => PlatformRemoveAttendeeFromEvent(newAttendee, eventId);
+
         static ArgumentException InvalidCalendar(string calendarId) =>
             new ArgumentOutOfRangeException($"No calendar exists with the ID '{calendarId}'.", nameof(calendarId));
 
--- a/Xamarin.Essentials/Calendars/Calendars.uwp.cs
+++ b/Xamarin.Essentials/Calendars/Calendars.uwp.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.Appointments;
 
@@ -9,14 +10,24 @@ namespace Xamarin.Essentials
     public static partial class Calendars
     {
         static Task<AppointmentStore> uwpAppointmentStore;
+        static AppointmentStoreAccessType lastRequestType;
 
-        static Task<AppointmentStore> GetInstanceAsync() =>
-            uwpAppointmentStore ??= AppointmentManager.RequestStoreAsync(AppointmentStoreAccessType.AllCalendarsReadOnly).AsTask();
+        static Task<AppointmentStore> GetInstanceAsync(AppointmentStoreAccessType type = AppointmentStoreAccessType.AppCalendarsReadWrite)
+        {
+            if (uwpAppointmentStore == null || lastRequestType != type)
+            {
+                uwpAppointmentStore = AppointmentManager.RequestStoreAsync(type).AsTask();
+                lastRequestType = type;
+            }
+
+            return uwpAppointmentStore;
+        }
 
         static async Task<IEnumerable<Calendar>> PlatformGetCalendarsAsync()
         {
-            var instance = await GetInstanceAsync().ConfigureAwait(false);
+            await Permissions.RequestAsync<Permissions.CalendarRead>();
 
+            var instance = await GetInstanceAsync(AppointmentStoreAccessType.AllCalendarsReadWrite).ConfigureAwait(false);
             var calendars = await instance.FindAppointmentCalendarsAsync(FindAppointmentCalendarsOptions.IncludeHidden).AsTask().ConfigureAwait(false);
 
             return ToCalendars(calendars).ToList();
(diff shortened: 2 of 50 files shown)


nickrandolph:
@mattleibow I've created this draft PR so that you can start to take a look through the code and suggest changes. I still need to review all the write apis to make them consistent with the changes you made to the read apis. 
We also need to tidy up the UI - The current UI is what we had in the write branch. I would like to use the UI you tidied up as the basis for the UI. Perhaps you could look at applying your changes?
I also need to review all the test cases as I think some of them are failing - for UWP some of the failing test cases aren't related to calendar.
Feel free to push changes directly to my branch but please let me know if you do so that I know to sync before I make further changes

Xamarin.Essentials/Types/Calendar.shared.cs
```diff
@@ -43,32 +47,138 @@ public CalendarEvent(string id, string calendarId, string title)
...
+    public enum IterationOffset
+    {
+        First = 0,
+        Second = 1,
+        Third = 2,
+        Fourth = 3,
+        Last = 4
+    }
+
+#endif
```
mattleibow, line 176: Because this is shared code, and these are constants, we cannot have different values. If I write netstandard and pick `First`, it will use the value `0` - which does not match anything on Android.

Xamarin.Essentials/Types/PlatformExtensions/CalendarExtensions.ios.macos.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Foundation;
+
+namespace Xamarin.Essentials
+{
+    public static partial class CalendarExtensions
+    {
+        // https://developer.apple.com/documentation/foundation/nsdate
+        // NSDate minimum date is 2001/01/01
+        static DateTime iosNSDateTimeSystemZeroPoint = new DateTime(2001, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
```
mattleibow, line 10: Is there a reason to use Apple's special date as opposed to the Unix time? It also makes conversion easier. But this is just me being a bit OCD.

Xamarin.Essentials/Types/Calendar.shared.cs
```diff
@@ -43,32 +47,138 @@ public CalendarEvent(string id, string calendarId, string title)
...
+        public uint? TotalOccurrences { get; set; }
+
+        public uint Interval { get; set; }
+
+        public DateTimeOffset? EndDate { get; set; }
+
+        public RecurrenceFrequency? Frequency { get; set; }
+
+        // Only allow event to occur on these days [not available for daily]
+        public List<CalendarDayOfWeek> DaysOfTheWeek { get; set; }
```
mattleibow, line 105: This is a list, and the enum is a flag.

But this is also a opportunity here... @jamesmontemagno @Redth ... Do we like/prefer enum flags vs list of classes? We can also do that Stringly typed constants. Like we do for the OS. Then we can even add methods to it. Maybe `Weekday.ToString() == "Monday"` and then `Weekday.ToShortString() == "Mon"` or something?

Xamarin.Essentials/Calendars/Calendars.shared.cs
```diff
@@ -47,6 +47,24 @@ public static async Task<CalendarEvent> GetEventAsync(string eventId)
             return await PlatformGetEventAsync(eventId);
         }
 
+        public static Task<CalendarEvent> GetEventInstanceAsync(string eventId, DateTimeOffset instanceDate) => PlatformGetEventInstanceAsync(eventId, instanceDate);
+
+        public static Task<string> CreateCalendarEvent(CalendarEvent newEvent) => PlatformCreateCalendarEvent(newEvent);
```
mattleibow, line 52: Maybe these names can just be `CreateEvent` as we already know about the calendar part.

Xamarin.Essentials/Calendars/Calendars.shared.cs
```diff
@@ -47,6 +47,24 @@ public static async Task<CalendarEvent> GetEventAsync(string eventId)
...
+
+        public static Task<bool> UpdateCalendarEvent(CalendarEvent eventToUpdate) => PlatformUpdateCalendarEvent(eventToUpdate);
+
+        public static Task<bool> SetEventRecurrenceEndDate(string eventId, DateTimeOffset recurrenceEndDate) => PlatformSetEventRecurrenceEndDate(eventId, recurrenceEndDate);
+
+        public static Task<bool> DeleteCalendarEventInstanceByDate(string eventId, string calendarId, DateTimeOffset dateOfInstanceUtc) => PlatformDeleteCalendarEventInstanceByDate(eventId, calendarId, dateOfInstanceUtc);
```
mattleibow, line 58: To fit with more .NET-y things, maybe a better name would be `DeleteEventInstanceAsync` and then the parameters can overload to be the ID and the instance date. This way, the popup in intellisense groups related actions/object together.

Xamarin.Essentials/Calendars/Calendars.shared.cs
```diff
@@ -47,6 +47,24 @@ public static async Task<CalendarEvent> GetEventAsync(string eventId)
             return await PlatformGetEventAsync(eventId);
         }
 
+        public static Task<CalendarEvent> GetEventInstanceAsync(string eventId, DateTimeOffset instanceDate) => PlatformGetEventInstanceAsync(eventId, instanceDate);
+
+        public static Task<string> CreateCalendarEvent(CalendarEvent newEvent) => PlatformCreateCalendarEvent(newEvent);
+
+        public static Task<bool> UpdateCalendarEvent(CalendarEvent eventToUpdate) => PlatformUpdateCalendarEvent(eventToUpdate);
+
+        public static Task<bool> SetEventRecurrenceEndDate(string eventId, DateTimeOffset recurrenceEndDate) => PlatformSetEventRecurrenceEndDate(eventId, recurrenceEndDate);
```
mattleibow, line 56: As I look at this `SetEventRecurrenceEndDate`, and the `AddAttendeeToEvent` methods... This feels a bit different to use than .NET.

Maybe we should start thinking about making the calendar object a bit smarter? Or something. I like the idea of Calendar being a POCO, but we are doing a lot of things here that maybe should be on the object instead? I don't know... thoughts are happening in my head.

@Redth @jamesmontemagno this might require better brains than mine.

Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs
```diff
@@ -51,7 +51,7 @@ internal static async Task PlatformSpeakAsync(string text, SpeechOptions options
 
             void TryCancel()
             {
-                ss.StopSpeaking(NSSpeechBoundary.hWord);
+                ss.StopSpeaking(NSSpeechBoundary.Word);
```
mattleibow, line 54: Did they fix this bug?

Xamarin.Essentials/Calendars/Calendars.uwp.cs
```diff
@@ -9,24 +10,34 @@ namespace Xamarin.Essentials
     public static partial class Calendars
     {
         static Task<AppointmentStore> uwpAppointmentStore;
+        static AppointmentStoreAccessType lastRequestType;
 
-        static Task<AppointmentStore> GetInstanceAsync() =>
-            uwpAppointmentStore ??= AppointmentManager.RequestStoreAsync(AppointmentStoreAccessType.AllCalendarsReadOnly).AsTask();
+        static Task<AppointmentStore> GetInstanceAsync(AppointmentStoreAccessType type = AppointmentStoreAccessType.AppCalendarsReadWrite)
```
mattleibow, line 15: Maybe we can have multiple fields for this? Do the stores communicate? Like if I add a new calendar, will it appear in the other store? Is this expensive? Should we even be caching this?

Xamarin.Essentials/Calendars/Calendars.uwp.cs
```diff
@@ -9,24 +10,34 @@ namespace Xamarin.Essentials
...
+                lastRequestType = type;
+            }
+
+            return uwpAppointmentStore;
+        }
 
         static async Task<IEnumerable<Calendar>> PlatformGetCalendarsAsync()
         {
-            var instance = await GetInstanceAsync().ConfigureAwait(false);
+            await Permissions.RequestAsync<Permissions.CalendarRead>();
```
mattleibow, line 28: Can we move the permissions to the shared code?

mattleibow (commented):
Thanks for this PR! So many features!

I had a quick look, and I am going more for comments on the API structure for now. I am sure the implementation is great, so we can go over that later. I saw your last PR, and that was pretty good.

I sort of brain dumped as I went along.

Let me know your thoughts on my comments.

Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs
mattleibow: I think this is literally a big in the API: https://docs.microsoft.com/en-us/dotnet/api/appkit.nsspeechboundary?view=xamarin-mac-sdk-14

PR #1412, opened 2020-10-04
[FilePicker] Add support for selecting PDF files (Issue 1404)

### LINKED ISSUE #1404: [Enhancement] Add FilePickerFileType.Pdf to FilePicker
## Summary
Add option for picking PDFs on all supported platforms. I realize we could use a FilePickerFileType for all possible filetypes (or maybe a parser), but PDFs is such a common format that i think it justifies the addition.

## API Changes
To make it simpler for developers to pick PDFs we add `FilePickerFileType.Pdf` to the FilePicker API for all supported platforms.

## Intended Use Case
Picking PDFs

RodgerLeblanc:
### Description of Change ###

Add option for picking PDFs on all supported platforms. I realize we could use a FilePickerFileType for all possible filetypes (or maybe a parser), but PDFs is such a common format that i think it justifies the addition.

### Bugs Fixed ###

- Fixes #1404 

### API Changes ###

Added: 
 
- `FilePickerFileType.Pdf` enum type
- `FilePickerFileType.Jpeg` enum type

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.android.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.android.cs
@@ -86,5 +86,11 @@ namespace Xamarin.Essentials
             {
                 { DevicePlatform.Android, new[] { "video/*" } }
             });
+
+        public static FilePickerFileType PlatformPdfFileType() =>
+            new FilePickerFileType(new Dictionary<DevicePlatform, IEnumerable<string>>
+            {
+                { DevicePlatform.Android, new[] { "application/pdf" } }
+            });
     }
 }
--- a/Xamarin.Essentials/FilePicker/FilePicker.android.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.android.cs
@@ -81,6 +81,12 @@ namespace Xamarin.Essentials
                 { DevicePlatform.Android, new[] { "image/png" } }
             });
 
+        static FilePickerFileType PlatformJpegFileType() =>
+            new FilePickerFileType(new Dictionary<DevicePlatform, IEnumerable<string>>
+            {
+                { DevicePlatform.Android, new[] { "image/jpeg" } }
+            });
+
         static FilePickerFileType PlatformVideoFileType() =>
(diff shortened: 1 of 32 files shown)


Xamarin.Essentials/FilePicker/FilePicker.android.cs
```diff
@@ -86,5 +86,11 @@ public partial class FilePickerFileType
             {
                 { DevicePlatform.Android, new[] { "video/*" } }
             });
+
+        public static FilePickerFileType PlatformPdfFileType() =>
```
mattleibow, line 90: Could we make these internal? For all the platforms. I know that you just copied the surrounding code, but I think this should have not been public since the correct access is the public properties.

docs/en/Xamarin.Essentials/FilePickerFileType.xml
```diff
@@ -87,6 +87,23 @@
...
+      <AssemblyInfo>
+        <AssemblyName>Xamarin.Essentials</AssemblyName>
+        <AssemblyVersion>1.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>Xamarin.Essentials.FilePickerFileType</ReturnType>
+      </ReturnValue>
+      <Docs>
+        <summary>To be added.</summary>
+        <remarks>To be added.</remarks>
```
mattleibow, line 104:
```suggestion
        <summary>PDF file type (*.pdf).</summary>
        <remarks></remarks>
```

mattleibow (commented):
Looking really good!
I got a suggestion for docs and also wondering if you could fix the visibility. I think we missed that and it is redundant.

Xamarin.Essentials/FilePicker/FilePicker.android.cs
RodgerLeblanc: Great suggestion, I've made the method `private` instead of `internal` as it is only used in the `partial` class.

mattleibow (dismissed):
Very nice! Thanks for the PR.

Issue #963, opened 2019-10-30
[Enhancement] Share feature throws different exceptions across platforms.


Mrnikbobjeff:
### Description
The only platform to validate share request data is Tizen, all others will throw nullreferenceexceptions if passed null via invalid constructed sharerequests. We should validate the share request equally across all platforms

### Expected Behavior
All platforms throw ArgumentNullException
### Actual Behavior
NullReferenceException on Android in FileProvider
### Basic Information

- Version with issue: 1.3.1
- Last known good version: None

RodgerLeblanc:
I'll work on this one.

mattleibow:
Thanks @RodgerLeblanc!

RodgerLeblanc:
Can't tell you why, but this repo is ineligible for Hacktoberfest.

![image](https://user-images.githubusercontent.com/6235690/95026212-bf87b800-065d-11eb-93cb-c37628b159b8.png)

This repo has some issue with the `hacktoberfest` label, so you might want to fix this.
You probably can add the `hacktoberfest` tag to the repo to make it eligible manually, or get in touch with Hacktoberfest team to understand why this repo was marked as ineligible.

mattleibow:
@RodgerLeblanc  I added the tag. I see the rules are still changing. Thanks to the "nice" folks in the community.

PR #1413, opened 2020-10-04
Throw ArgumentNullException when passing null to a Share request

RodgerLeblanc:
### Description of Change ###

Ensures that all supported platforms throws the same exception when passing invalid argument to a Share request.
No sample added as this behavior can be validated by running the samples and sending a Share request without any arguments.

### Bugs Fixed ###

- Related to issue #963 

### API Changes ###

None.

### Behavioral Changes ###

All platforms will now throw an `ArgumentNullException` when passing `null` as parameter to Share API. Previously, the Tizen platform was throwing an `ArgumentNullException` and the other platforms would throw a `NullReferenceException` when trying to access a property of a `null` object.

### PR Checklist ###

### DIFF
--- a/DeviceTests/DeviceTests.Shared/Share_Tests.cs
+++ b/DeviceTests/DeviceTests.Shared/Share_Tests.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.Essentials;
+using Xunit;
+
+namespace DeviceTests
+{
+    public class Share_Tests
+    {
+        [Fact]
+        public async Task Share_NullText()
+        {
+            string text = null;
+            await Assert.ThrowsAsync<ArgumentNullException>(() => Share.RequestAsync(text));
+        }
+
+        [Fact]
+        public async Task Share_NullRequest()
+        {
+            ShareTextRequest request = null;
+            await Assert.ThrowsAsync<ArgumentNullException>(() => Share.RequestAsync(request));
+        }
+    }
+}
--- a/DeviceTests/DeviceTests.Shared/Share_Tests.cs
+++ b/DeviceTests/DeviceTests.Shared/Share_Tests.cs
@@ -8,16 +8,47 @@ namespace DeviceTests
     public class Share_Tests
     {
         [Fact]
-        public async Task Share_NullText()
+        public async Task Share_ShareTextRequestWithInvalidTextAndUri()
         {
-            string text = null;
-            await Assert.ThrowsAsync<ArgumentNullException>(() => Share.RequestAsync(text));
+            var request = new ShareTextRequest
+            {
+                Text = null,
+                Uri = null
+            };
+            await Assert.ThrowsAsync<ArgumentException>(() => Share.RequestAsync(request));
         }
 
         [Fact]
-        public async Task Share_NullRequest()
+        public async Task Share_NullShareTextRequest()
         {
             ShareTextRequest request = null;
+            await Assert.ThrowsAsync<ArgumentNullException>(() => Share.RequestAsync(request));
+        }
+
+        [Fact]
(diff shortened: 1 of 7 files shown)


Xamarin.Essentials/Share/Share.android.cs
```diff
@@ -9,6 +9,12 @@ public static partial class Share
     {
         static Task PlatformRequestAsync(ShareTextRequest request)
         {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+
+            if (string.IsNullOrEmpty(request.Text) && string.IsNullOrEmpty(request.Uri))
+                throw new ArgumentNullException(nameof(request.Text));
+
```
mattleibow, line 17: Maybe we can move this to the shard layer? Then we don't need to duplicate the code in every platform.

DeviceTests/DeviceTests.Shared/Share_Tests.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        [Fact]
+        public async Task Share_NullText()
+        {
+            string text = null;
+            await Assert.ThrowsAsync<ArgumentNullException>(() => Share.RequestAsync(text));
+        }
+
+        [Fact]
+        public async Task Share_NullRequest()
```
Mrnikbobjeff, line 18: Consider adding tests  which actually validate the behaviour noticed in the ticket, i.e. share requests which have nested properties that are null.

Xamarin.Essentials/Share/Share.ios.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
```
Mrnikbobjeff, line 1: Can be reverted

Xamarin.Essentials/Share/Share.macos.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
```
Mrnikbobjeff, line 1: Can be reverted

Xamarin.Essentials/Share/Share.shared.cs
```diff
@@ -15,11 +14,28 @@ public static partial class Share
...
 
-        public static Task RequestAsync(ShareTextRequest request) =>
-            PlatformRequestAsync(request);
+        public static Task RequestAsync(ShareTextRequest request)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+
+            if (string.IsNullOrEmpty(request.Text) && string.IsNullOrEmpty(request.Uri))
+                throw new ArgumentNullException(nameof(request.Text));
```
Mrnikbobjeff, line 23: This will possibly confuse people if they pass a sharerequest with a null uri, they would get the wrong nameof expression. Splitting this into two if statements removes that problem

Mrnikbobjeff (changes requested):
Looking good just some minor nitpicks. The Exception nameof parameter is the most important change IMHO which needs to be adressed

Xamarin.Essentials/Share/Share.shared.cs
mattleibow: Good catch.

Maybe this should not be an `ArgumentNullException` since the argument was not null. Maybe a `ArgumentException` with a better message? "Both the Text and Uri were invalid. Make sure to xyz." or something?

Mrnikbobjeff: In an ideal world I would love to prevent this from occuring at all instead of having a developer have  to investigate which parameters are valid. I know it would be a breaking change, but changing shareRequest to have readonly properties and having only constructors exposed which set required fields. This is the solution imho, as being able to construct objects which always fail if not all parameters are set is not very elegant

mattleibow (approved):
I think I quite like this.

PR #1416, opened 2020-10-05
Moved over brightness implementations, initial testing and also added…

Mrnikbobjeff:
… tizen implementation

### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Types/Brightness.shared.cs
+++ b/Xamarin.Essentials/Types/Brightness.shared.cs
@@ -1,0 +1,75 @@
+﻿using System;
+#if WINDOWS_UWP
+using UwpBrightness = Windows.Graphics.Display.BrightnessOverride;
+#endif
+
+namespace Xamarin.Essentials
+{
+    [Preserve(AllMembers = true)]
+    public readonly partial struct BrightnessOverride : IDisposable, IEquatable<BrightnessOverride>
+    {
+        internal BrightnessOverride(Brightness oldBrightness, Brightness appliedBrightness)
+        {
+            OldBrightness = oldBrightness;
+            AppliedBrightness = appliedBrightness;
+#if WINDOWS_UWP
+            @override = null;
+#endif
+        }
+
+#if WINDOWS_UWP
+        readonly UwpBrightness @override;
+
+        public BrightnessOverride(Brightness oldBrightness, Brightness newBrightness, UwpBrightness @override)
+            : this(oldBrightness, newBrightness)
+        {
+            this.@override = @override;
+        }
+
+        public void Dispose()
+        {
+            @override.StopOverride();
+        }
+
+        public bool Equals(BrightnessOverride other)
+        {
+            return OldBrightness.Equals(other.OldBrightness) && AppliedBrightness.Equals(other.AppliedBrightness) && @override.Equals(other.@override);
+        }
+#endif
+
+        public Brightness OldBrightness { get; }
+
+        public Brightness AppliedBrightness { get; }
+
+#if !WINDOWS_UWP
+        public void Dispose()
+        {
+            AppInfo.SetBrightness(OldBrightness);
+        }
+#endif
+
+#if !WINDOWS_UWP
+        public bool Equals(BrightnessOverride other)
+        {
+            return OldBrightness.Equals(other.OldBrightness) && AppliedBrightness.Equals(other.AppliedBrightness);
+        }
+#endif
+    }
+
+    [Preserve(AllMembers = true)]
+    public readonly struct Brightness
+    {
+        public double Value { get; }
+
+        public Brightness(double value)
+        {
+            const double minValue = 0d;
+#if __ANDROID__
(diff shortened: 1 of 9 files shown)


Mrnikbobjeff:
This is still missing some things: I checked around and Tizen does not have a concept of resetting the device display brightness. Also, Tizen is the only os exposing multiple displays for brightness, so i currently only use the main display. I also still have to test ios, but due to my setup this always takes longer. @mattleibow @Redth

Xamarin.Essentials/Types/Brightness.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+#if !WINDOWS_UWP
+        public bool Equals(BrightnessOverride other)
+        {
+            return OldBrightness.Equals(other.OldBrightness) && AppliedBrightness.Equals(other.AppliedBrightness);
+        }
+#endif
+    }
+
+    [Preserve(AllMembers = true)]
+    public readonly struct Brightness
```
dimonovdd, line 60: Why was a class created to describe a single property?

Xamarin.Essentials/Types/Brightness.shared.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        public double Value { get; }
+
+        public Brightness(double value)
+        {
+            const double minValue = 0d;
+#if __ANDROID__
+            const double minValue = -1d;
```
dimonovdd, line 68: It would be more convenient to use the range from 0 to 1 on Android too

Xamarin.Essentials/Types/Brightness.shared.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+#if WINDOWS_UWP
+using UwpBrightness = Windows.Graphics.Display.BrightnessOverride;
+#endif
+
+namespace Xamarin.Essentials
+{
+    [Preserve(AllMembers = true)]
```
dimonovdd, line 8: I think this attribute is not needed

Xamarin.Essentials/Types/Brightness.shared.cs
mattleibow: All constants must be the same across platforms. The value is actually copied to the caller, so if an assembly is compiled for one platform (including netstandard) then it will remain at the value and not actually change. Same for enum values.

Xamarin.Essentials/Types/Brightness.shared.cs
mattleibow: Yep. Not anymore. I think we needed it in the past due to bad things or something.

mattleibow:
Should this maybe be on `Display`? Even though we are per-app, it is actually changing the properties of the display, not the app.

Xamarin.Essentials/Types/Brightness.shared.cs
dimonovdd: `relative_value = (value - low) / (high - low)`
`scaled_value = low_2 + (high_2 - low_2) * relative_value`
Is it possible to use this formula to convert the range for Android?
-1 on Android and 0 on iOS have same behavior?

Xamarin.Essentials/Types/Brightness.shared.cs
Mrnikbobjeff: Check all other instances of sensor readings, such as barometer.

Xamarin.Essentials/Types/Brightness.shared.cs
Mrnikbobjeff: No they do not, which is why I have this check. Android also offers the same range as iOS for display brightness, but if you set a negative value the override ends and the is supplied brightness is chosen

Mrnikbobjeff:
> Should this maybe be on `Display`? Even though we are per-app, it is actually changing the properties of the display, not the app.

I thought about moving it to a  brightness class entirely as the behaviour is slightly different across platforms, making neither app not display the perfect fit

mattleibow:
I was just typing that this could work. Then I remembered the wake lock we have, which is in display... I think we need to be consistent.

However, there are enough members on this API that it could stand alone...

Mrnikbobjeff:
Once support for macos and win32 gets merged we can support them as well. [MacOs](https://github.com/nriley/brightness/blob/master/brightness.c)
Win32 support would work with [ddc/ci](https://en.wikipedia.org/wiki/Display_Data_Channel#DDC/CI)

jfversluis:
Thanks for all your hard work on this @Mrnikbobjeff! Unfortunately we won't be able to make this happen anymore for Xamarin.Essentials. If you're still interested in this, make sure to open an issue for .NET MAUI to discuss possibilities there

Issue #1429, opened 2020-10-08
[Bug] macOS: TextToSpeech NSSpeechBoundary.hWord obsolete


Suplanus:
### Steps to Reproduce

1. Build Xamarin.Essentials under macOS

### Expected Behavior
Success

### Actual Behavior
Failed:
```
Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs(33,33): Error CS0618: 'NSSpeechBoundary.hWord' is obsolete: 'Use 'Word' instead.' (CS0618) (Xamarin.Essentials)
```

### Basic Information

- Version with issue: 1.6.0-pre2
- Affected Devices: macOS

mattleibow:
Closing this as this requires that everyone use the very latest code to compile. As this is not too critical right now, it should be fine to leave it as is.
I actually merged this, but then all of CI failed because it was on the stable versions only.

Suplanus:
OK. No problem. But why you close it? The open issue would be a good reminder for a todo 🙃

mattleibow:
Seems legit

jamesmontemagno:
I have updated it with some new tags

PR #1430, opened 2020-10-08
macOS: TextToSpeech replaced NSSpeechBoundary

Suplanus:
### Description of Change ###

`hWord` is obsolete, replaced it with `Word`

### Bugs Fixed ###

- Related to issue #1429

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs
+++ b/Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs
@@ -51,7 +51,7 @@ namespace Xamarin.Essentials
 
             void TryCancel()
             {
-                ss.StopSpeaking(NSSpeechBoundary.hWord);
+                ss.StopSpeaking(NSSpeechBoundary.Word);
                 tcs.TrySetResult(true);
             }
 


Xamarin.Essentials/TextToSpeech/TextToSpeech.macos.cs
```diff
@@ -51,7 +51,7 @@ internal static async Task PlatformSpeakAsync(string text, SpeechOptions options
 
             void TryCancel()
             {
-                ss.StopSpeaking(NSSpeechBoundary.hWord);
+                ss.StopSpeaking(NSSpeechBoundary.Word);
```
jamesmontemagno, line 54: This is a tricky one as this means anyone that hasn't updated to the latest version this may cause issues... @mattleibow ....

mattleibow: Yeah. I was thinking that.

But, this only applies to those building this source code. As enums are constants, only the value is used in IL.

It should be OK.

mattleibow: But, I think we should test to confirm my logic here. It should be a matter of just downloading this built assembly and viewing the IL.

I'll have a look to see.

mattleibow: I'm setting up a VS2017 VM right now and I will check this.

PR #1433, opened 2020-10-08
macOS: Screenshot

Suplanus:
### Description of Change ###

**Screenshot for macOS**
I thought it would be easy... hours later after testing with `CGWindowListCreateImage` and much more (`CGImage.ScreenImage`), this is the only way I can find out.

**Documentation**
By default only the app windows are visible. But macOS is asking to set the permission in the settings.
The security setting have to be enabled to get all apps on screen.

image

image

### PR Checklist ###


Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                case ScreenshotFormat.Png:
+                    utType = UTType.PNG;
+                    break;
+            }
+            var fileURL = new NSUrl(filename, false);
+
+            using (var dataConsumer = new CGDataConsumer(fileURL))
+            {
+                var imageDestination = CGImageDestination.Create(dataConsumer, utType, 1);
```
mattleibow, line 77: Maybe you can save to an `NSMutableData` and then when the stream is requested, just return the `NSData.AsStream` result?

Saving to a file is not _bad_, but no need to do that if we can avoid it. I personally don't like putting things in any files that the user can see. But again, no real issue.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,83 @@
...
+    public static partial class Screenshot
+    {
+        internal static bool PlatformIsCaptureSupported => true;
+
+        const string appServicesPath =
+    "/System/Library/Frameworks/ApplicationServices.framework/Versions/Current/ApplicationServices";
+
+        // https://developer.apple.com/documentation/coregraphics/1454595-cgdisplaycreateimage
+        [DllImport(appServicesPath, EntryPoint = "CGDisplayCreateImage")]
+        static extern /* CGImageRef */ IntPtr CGDisplayCreateImage(int displayId);
```
mattleibow, line 24: It might be worth adding a small helper class in the Platform file so we don't mix functionality with implementation: https://github.com/xamarin/Essentials/blob/main/Xamarin.Essentials/Platform/Platform.macos.cs#L23

You can probably also return the CGImage from there as well so this file only works with concrete types.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        static Task<ScreenshotResult> PlatformCaptureAsync()
+        {
+            using var pool = new NSAutoreleasePool();
+            var location = NSApplication.SharedApplication.KeyWindow.Frame.Location;
+            var screens = new List<NSScreen>(NSScreen.Screens);
+            var screen = screens.First(obj => obj.Frame.Contains(location));
```
mattleibow, line 31: This might thro a confusing exception if no screen was found. Maybe do a FirstOrDefault and then throw an invalid op ex with a useful message... 

This also brings up a though for me. Is it possible that a view can be off all the screens? What are the cases where this could actually fail? Does macOS support having no screens?

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,83 @@
...
+    "/System/Library/Frameworks/ApplicationServices.framework/Versions/Current/ApplicationServices";
+
+        // https://developer.apple.com/documentation/coregraphics/1454595-cgdisplaycreateimage
+        [DllImport(appServicesPath, EntryPoint = "CGDisplayCreateImage")]
+        static extern /* CGImageRef */ IntPtr CGDisplayCreateImage(int displayId);
+
+        static Task<ScreenshotResult> PlatformCaptureAsync()
+        {
+            using var pool = new NSAutoreleasePool();
+            var location = NSApplication.SharedApplication.KeyWindow.Frame.Location;
```
mattleibow, line 29: Can the app call with without a window? Like the AppDelegate ctor? Maybe this should use `?.` just to be safe. 

Not sure what we want to do if there is no window or no screen... Maybe an InvalidOpEx with a "helpful" message.

mattleibow (commented):
Thanks for this PR! I was wanting to get this in, but didn't manage to.

This is all vey nice, just have a few questions about corner cases. Also a suggestion to try for a memory stream.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: >Do we need to maybe add some permissions API to have a popup for the user, or will macOS automatically do this for us?

I added some description in the PR. In short: Its not needed but if you want a screenshot of other apps you have to enable in security settings.

>I personally don't like putting things in any files that the user can see.

Same for me. I will try.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: >Does macOS support having no screens?

Good question... I had `FirstOrDefault()`, but then I thought its not possible to have.
But it could be:
You have a mac mini, start the app, unplug the screen and click...
OK, i will catch that 🥳

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: In my understanding Xamarin.macOS and Xamarin.Forms need a window. So I think this could not be null.

Suplanus: I implement a fallback with `CurrentMouseLocation`.

Suplanus:
I committed the changes. Looks much better, thanks @mattleibow 💖
Perhaps you can have a second look at it.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            cgImage = image;
+            Width = (int)cgImage.Width;
+            Height = (int)cgImage.Height;
+        }
+
+        internal Task<Stream> PlatformOpenReadAsync(ScreenshotFormat format)
+        {
+            var utType = UTType.JPEG;
+            switch (format)
```
Mrnikbobjeff, line 41: A switch with a single case can be rewritten as an if

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var image = ApplicationServices.GetScreenshot();
+            var result = new ScreenshotResult(image);
+            return Task.FromResult(result);
+        }
+    }
+
+    public partial class ScreenshotResult
+    {
+        readonly CGImage cgImage;
```
Mrnikbobjeff, line 29: This will have to be disposed or risk being pushed on the finalization queue

Xamarin.Essentials/Platform/Platform.macos.cs
```diff
@@ -20,6 +23,46 @@ internal static NSWindow GetCurrentWindow(bool throwIfNull = true)
         }
     }
 
+    internal static class ApplicationServices
+    {
+        const string appServicesPath = "/System/Library/Frameworks/ApplicationServices.framework/Versions/Current/ApplicationServices";
+
+        // https://developer.apple.com/documentation/coregraphics/1454595-cgdisplaycreateimage
+        [DllImport(appServicesPath, EntryPoint = "CGDisplayCreateImage")]
+        static extern /* CGImageRef */ IntPtr CGDisplayCreateImage(int displayId);
```
Mrnikbobjeff, line 32: Can we use something derived from SafeHandle here? Would be preferable to oblique IntPtrs and ease freeing of resources

Mrnikbobjeff (commented):
This will leak unless we adress the freeing issue I believe

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: I would have used `if` but I used the style from iOS:
https://github.com/xamarin/Essentials/blob/0dba844e8085f7cb36d62115b005620d541e0fbd/Xamarin.Essentials/Screenshot/Screenshot.ios.tvos.cs#L35

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: Sorry, but have no idea how to do that. `UiImage` is also not disposed in iOS:
https://github.com/xamarin/Essentials/blob/0dba844e8085f7cb36d62115b005620d541e0fbd/Xamarin.Essentials/Screenshot/Screenshot.ios.tvos.cs#L23

Xamarin.Essentials/Platform/Platform.macos.cs
Suplanus: I think this should be discussed in separate issue. `Platform.IOKit` would also be affected.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: The difference here is that they actually use two cases, the two swtich statements are not similar. Yours is a overly complicated if block, they have an if/else block

Suplanus: Thats correct. In my reading style I prefer to use the full statement, because the default knows nobody. And wie would be safe if there will be a new type. Would be this OK?

```csharp
NSString utType;
switch (format)
{
    case ScreenshotFormat.Png:
        utType = UTType.PNG;
        break;
    case ScreenshotFormat.Jpeg:
        utType = UTType.JPEG;
        break;
    default:
        throw new NotImplementedException("The ScreenshotFormat is not supported");
}
```

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: That is what I realised as well, which is kind of a bummer. UIImage does not have to be disposed, CGImage has to be though. If you compare the docs [UIImage](https://docs.microsoft.com/en-us/dotnet/api/uikit.uiimage?view=xamarin-ios-sdk-12#methods) [CGImage](https://docs.microsoft.com/en-us/dotnet/api/coregraphics.cgimage?view=xamarin-ios-sdk-12#methods) you will see that while both expose a Dispose method for NSObject, CGImage does not have the same Dipose method as NSObject while UIImage does.

Mrnikbobjeff: Found a solution which I belive is ideal. You can remove the macos specific ScreenshotResult, convert the CIImage to UIImage via the constructor and alleviate the problems

Mrnikbobjeff: https://docs.microsoft.com/en-us/dotnet/api/uikit.uiimage.-ctor?view=xamarin-ios-sdk-12#UIKit_UIImage__ctor_CoreImage_CIImage_

Suplanus: Is this possible in macOS SDK? I think not because I need a reference to UIKit in iOS?

Mrnikbobjeff: You are correct,you can not eliminate the struct but we can eliminate cgimage by moving to nsimage, which will follow the guidance to get away from native objects as soon as possible. NSImage also has a constructor accepting CGImage but also requires a CGSize trivially constructable by CGImage

Suplanus: `CGImageDestination` not accept nsImage so we back again 😔

Suplanus: How do you think about the added finalizer?

Mrnikbobjeff: Check out this gist: https://gist.github.com/Mrnikbobjeff/7fb533642a1b9bfeb7376a79d43da7dc

Mrnikbobjeff: And if you had to implement a finalizer be sure to check out [this](http://joeduffyblog.com/2005/04/08/dg-update-dispose-finalization-and-resource-management/) or take a look at my record audio pr recording class, it has the correct finalization + disposing pattern

Suplanus: Oh nice. Great, thank you. Making PRs is the best way to learn 🦄 
I have committed the changes.

`CGImage.ScreenImage()` is not working correctly. I also added this as comment and in the PR description.

Mrnikbobjeff: I am sorry, I have missed disposing the actual stream in the custom stream class dispose method

Mrnikbobjeff: And what do you mean it does not work correctly?

Suplanus: Oh my... time for weekend...
I added it, perhaps you have a look (again). Thank you!

`CGImage.ScreenImage()` have two problems:

**Not possible to catch entire screen**
image

**Not working if the window is not on the main screen (empty)**
image

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -17,6 +17,7 @@ static Task<ScreenshotResult> PlatformCaptureAsync()
         {
             var image = ApplicationServices.GetScreenshot();
             var result = new ScreenshotResult(new NSImage(image, new CGSize(image.Width, image.Height)));
+            image?.Dispose();
```
Mrnikbobjeff, line 20: Remove this and use 'using var image = '

Mrnikbobjeff (changes requested):

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Suplanus: Have to still get used to it... Thanks. It`s commited.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: The only concern I have is app store releaseability, I do not know the policy for this for the mac app store. I am pretty sure that CIImage methods are allowed so I will follow up on that. Otherwise this looks great :)

Suplanus: Next week I will try to commit Cuterdio to Mac App Store... So I can test.
Thanks for your help!

Mrnikbobjeff: https://github.com/xamarin/xamarin-macios/blob/caba0a66be1490b28f5fb38b17919c3eabebee56/src/CoreGraphics/CGImage.cs#L222

Mrnikbobjeff (dismissed):

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: /* Flags for CGWindowImageOption values.  These may be ORed together. */

enum {

  /* If `CGRectNull' is passed as the screen bounds, then then bounds
     computation includes window frame ornamentation, such as a shadow. */
  kCGWindowImageDefault = 0,

  /* If `CGRectNull' is passed as the screen bounds, then then bounds
     computation excludes window frame ornamentation, such as a shadow. */
  kCGWindowImageBoundsIgnoreFraming = (1 << 0),

  /* Force the created image to be opaque.  Empty areas are white */
  kCGWindowImageShouldBeOpaque = (1 << 1),

  /* Only draw the windows' shadows, not the windows themselves. */
  kCGWindowImageOnlyShadows = (1 << 2),

  /* Return the best image resolution. The screen size may be
     different than the returned image size. */
  kCGWindowImageBestResolution = (1 << 3),

  /* Return the nominal image resolution. The screen size
     equals the returned image size. */
  kCGWindowImageNominalResolution = (1 << 4)
};

/* Create an image containing a composite of the specified set of windows
   contained within a rectangular area. The set of windows is specified
   using options from `CGWindowListOption', along with an optional
   additional window ID.

   The windows list options are:

   --- kCGWindowListOptionAll, kCGWindowListOptionOnScreenOnly: Use all
   on-screen windows in this user session to construct the image. The
   parameter `windowID' should be `kCGNullWindowID'.

   --- kCGWindowListOptionOnScreenAboveWindow: Use all on-screen windows in
   this user session above the window specified by `windowID', ordered from
   front to back, to construct the image. To include the window specified by
   `windowID', add the flag `kCGWindowListOptionIncludingWindow'.

   --- kCGWindowListOptionOnScreenBelowWindow: Use all on-screen windows in
   this user session below the window specified by `windowID', ordered from
   front to back, to construct the image. To include the window specified by
   `windowID', add the flag `kCGWindowListOptionIncludingWindow'.

   --- kCGWindowListOptionIncludingWindow: Use only the window specified by
   `windowID' to construct the image.

   The parameter `screenBounds' specifies the rectangle in screen space
   (origin at the upper-left; y-value increasing downward). Setting
   `screenBounds' to `CGRectInfinite' will include all the windows on the
   entire desktop. Setting `screenBounds' to `CGRectNull' will use the
   bounding box of the specified windows as the screen space rectangle.

   The parameter `imageOptions' allows you to specify whether the window
   frame ornamentation, such as a shadow or similar effect, should be
   included or excluded in the bounds calculation when `CGRectNull' is
   specified for the window bounds.

   If no windows meet the specified criteria, or the windows can't be read,
   then a transparent black image will be returned.

   Any on-screen window with sharing type `kCGWindowSharingNone' will not
   be included in the image.

   This function returns NULL if the caller is not running within a Quartz
   GUI session or the window server is disabled. */

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Xamarin.Essentials
+{
+    public static partial class Screenshot
+    {
+        internal static bool PlatformIsCaptureSupported => true;
+
+        static Task<ScreenshotResult> PlatformCaptureAsync()
+        {
+            using var image = ApplicationServices.GetScreenshot();
+            var result = new ScreenshotResult(new NSImage(image, new CGSize(image.Width, image.Height)));
```
mattleibow, line 19: Maybe we should do as little as possible here and pass CGImage to the result directly. Then do all the work in the get.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            uiImage = image;
+
+            Width = (int)image.Size.Width;
+            Height = (int)image.Size.Height;
+        }
+
+        internal Task<Stream> PlatformOpenReadAsync(ScreenshotFormat format)
+        {
+            return Task.FromResult<Stream>(new ImageTypeStream(uiImage.CGImage, format));
```
mattleibow, line 38: Do we need the stream type? What happens if we cache the data object, and then call as stream multiple times?

The GC will collect the image and data.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            public override long Seek(long offset, SeekOrigin origin) => decoratedStream.Seek(offset, origin);
+
+            public override void SetLength(long value) => decoratedStream.SetLength(value);
+
+            public override void Write(byte[] buffer, int offset, int count) => decoratedStream.Write(buffer, offset, count);
+
+            protected override void Dispose(bool disposing)
+            {
+                nativeImage.Dispose();
```
mattleibow, line 88: Maybe we should disposer in the screenshot result? I think all the other platforms will not exactly leak. The GC will get it.

mattleibow, line 88: Won't this disposer the original image and then future calls will use a bad object?

mattleibow (commented):
Things are looking better! Love the use of no files and all that.

I just got some comments on the lifetime of objects.

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: It should not, the cgimage is generated from the nsimage and seperate, disposal should not result in a bad object

Mrnikbobjeff: Well that would be a breaking change would it not? This way as long as the stream is handled correctly we do not have to add further disposal mechanics

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: Then we would have to dispose ScreenshotResult again, which is less desirable than disposing a stream

Xamarin.Essentials/Screenshot/Screenshot.macos.cs
Mrnikbobjeff: We need it to not have to dispose ScreenshotResult and to be able to call PlatformOpenReadAsync multiple times

Mrnikbobjeff:
This is the only platform where we take an actual screenshot instead of an application screenshot. I think this should work the same across uwp/macos/win32

Suplanus:
Good point!
I was thinking about that too, at writing the implementation. But for example on iOS its not only the app "window". If you are using split-mode on iPad, also other "windows" will be captured.
I doesn't tested UWP. But if the behavior is that only the "window" is captured, I think thats wrong:
1. iOS & Android capture full screen
2. Multi-Window app would not be captured

Second thing I was thinking about is: What about multiple screens? On iOS for example Carplay or Desktop external Display.

In short my suggestion:
- Current implementation should be screenshot (whole screen)
- UWP have to be changed

Mrnikbobjeff:
Android does not take a complete screenshot on my device. It only captures my application
screenshot
Screenshot_20201015-124353

Suplanus:
Oh... my fault. I tested on iOS as well.
There is also only the App visible. I will fix this PR. But also think it should have a other behavoir.
image

Suplanus:
I have updated the code. So we need no interop methods any more. Thanks for all your suggestions!

Suplanus:
Is there a reason why it's not merged?

jfversluis:
While this is supercool! I don't think it's feasible to take this in anymore at this point in the lifespan anymore. Maybe something can be salvaged and put into .NET MAUI? 

A big thank you for all the time and effort here!

Issue #1438, opened 2020-10-09
[Bug] SensorSpeed intervals differ between Android and iOS


janusw:
### Description

Looking at https://github.com/xamarin/Essentials/blob/main/Xamarin.Essentials/Types/SensorSpeed.ios.tvos.watchos.cs, one can see the following SensorSpeed intervals for iOS:

- `SensorSpeed.Default`: 225ms
- `SensorSpeed.UI`: 80ms
- `SensorSpeed.Game`: 40ms
- `SensorSpeed.Fastest`: 20ms

(The source code does not give any hint as to where these numbers come from.)

On Android, in contrast, there are no hard-coded numbers, but instead one relies on `Android.Hardware.SensorDelay`. According to https://developer.android.com/guide/topics/sensors/sensors_overview, the Android intervals are defined like this:

- `SENSOR_DELAY_NORMAL`: 200ms
- `SENSOR_DELAY_UI`: 60ms
- `SENSOR_DELAY_GAME`: 20ms
- `SENSOR_DELAY_FASTEST`: 0ms

Ideally the sensor intervals should be the same on all platforms, I guess. However, as one can see above, they differ considerably between Android and iOS (in particular the smaller intervals).

### Basic Information

- Version with issue: 1.6.0-pre2
- Last known good version: none
- IDE: VS for Mac
- Platform Target Frameworks: 
  - iOS:  14
  - Android: 11

janusw:
After doing some git archeology, I found that the iOS numbers were first introduced with the `OrientationSensor` implementation (#249). Btw, the OrientationSensor is also where I noticed the differences.

The `OrientationSensor` was initially implemented with commit 3432c0e65929a445acaa71f0b9bdf972a8d03ed6, which includes a revealing comment in `OrientationSensor.ios.cs`:

```
        // Timing intervales to match android sensor speeds in seconds
        // https://stackoverflow.com/questions/10044158/android-sensors
```

So, first of all I'm glad that there was the desire to match the Android sensor speeds :)

However, looking at https://stackoverflow.com/questions/10044158/android-sensors, the Xamarin iOS values apparently come from the top-rated answer, which only has approximations based on test results. Also it's from 2013.

The second answer also quotes the numbers from https://developer.android.com/guide/topics/sensors/sensors_overview, which I could even verify by measurement (for the fastest interval I actually get values around 5ms on my Zenfone 6 with Android 10).

[...]

PR #1439, opened 2020-10-09
SensorSpeed: use the same intervals on all platforms

janusw:
### Description of Change ###

* the intervals are matching those of Android, see https://developer.android.com/guide/topics/sensors/sensors_overview
* note that we do not use a zero-interval for SensorSpeed.Fastest (since this might turn off the sensor, in particular on UWP); instead we use 5ms, which is roughly what I have measured on Android 10 for SensorSpeed.Fastest

### Bugs Fixed ###

- Related to issue #1438

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/docs/en/Xamarin.Essentials/SensorSpeed.xml
+++ b/docs/en/Xamarin.Essentials/SensorSpeed.xml
@@ -28,7 +28,7 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>The default sensor speed.</summary>
+        <summary>The default sensor rate (5 Hz, corresponding to an interval of 200 ms).</summary>
       </Docs>
     </Member>
     <Member MemberName="Fastest">
@@ -45,7 +45,7 @@
       </ReturnValue>
       <MemberValue>3</MemberValue>
       <Docs>
-        <summary>Get the sensor data as fast as possible.</summary>
+        <summary>Get the sensor data as fast as possible (the actual rate is platform-dependent).</summary>
       </Docs>
     </Member>
     <Member MemberName="Game">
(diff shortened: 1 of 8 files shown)


docs/en/Xamarin.Essentials/SensorSpeed.xml
```diff
@@ -47,7 +47,7 @@
       </ReturnValue>
       <MemberValue>3</MemberValue>
       <Docs>
-        <summary>Get the sensor data as fast as possible.</summary>
+        <summary>Get the sensor data as fast as possible (the atual rate is platform-dependent).</summary>
```
jamesmontemagno, line 50: "the actual"

janusw: Thanks for catching this one, James! Fixed with b6ea7e6e3beefcf0079bee1512f4565d42efcb1e.

janusw:
@jamesmontemagno @mattleibow Any chance this might be considered for 1.7.0?

janusw:
Rebased to 1.7.0.

Also: Ping! (This has been waiting for 9 months now ...)

jfversluis:
I'm on the fence about this one to still include it in this stage of the Essentials lifetime. Technically it's a breaking change, we need to update docs, etc. As much as I love this to be consistent, I don't think we should do this anymore for Xamarin.Essentials.

I think this would be something nice to include in .NET MAUI though. Any thoughts @jamesmontemagno?

janusw:
> I'm on the fence about this one to still include it in this stage of the Essentials lifetime. Technically it's a breaking change, we need to update docs, etc. As much as I love this to be consistent, I don't think we should do this anymore for Xamarin.Essentials.
> 
> I think this would be something nice to include in .NET MAUI though.

Thanks for the comment, @jfversluis. To be honest, I had basically given up on this PR already :/

I can certainly forward-port it to the Maui repo, if there is a chance of getting it approved there.

jfversluis:
I know @janusw so sorry! We're doing better with .NET MAUI!

Let's not do this for Essentials anymore. It's a breaking change and a bigger one. I see you brought it to the .NET MAUI repo, thank you for that! So let's make sure we add it there.

Thank you for all the patience here, much appreciated!

PR #1440, opened 2020-10-11
Fixed running Sample (All Platforms)

### LINKED ISSUE #1441: [Bug] Samples.Android fails with FeatureNotSupportedException on Android 6
### Description

Stack trace:

```
Xamarin.Essentials.FeatureNotSupportedException: Specified method is not supported.
  at Xamarin.Essentials.AppActions.PlatformSetAsync (System.Collections.Generic.IEnumerable`1[T] actions) [0x0000d] in /Users/janus/github/Essentials/Xamarin.Essentials/AppActions/AppActions.android.cs:30
  at Xamarin.Essentials.AppActions.SetAsync (Xamarin.Essentials.AppAction[] actions) [0x00000] in /Users/janus/github/Essentials/Xamarin.Essentials/AppActions/AppActions.shared.cs:16
  at Samples.App.OnStart () [0x00087] in /Users/janus/github/Essentials/Samples/Samples/App.xaml.cs:51
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__7_0 (System.Object state) [0x00000] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/AsyncMethodBuilder.cs:1021
  at Android.App.SyncContext+<>c__DisplayClass2_0.<Post>b__0 () [0x00000] in <7d2292394f8c488b97f5bc2a0ac0240d>:0
  at Java.Lang.Thread+RunnableImplementor.Run () [0x00008] in <7d2292394f8c488b97f5bc2a0ac0240d>:0
  at Java.Lang.IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this) [0x00008] in <7d2292394f8c488b97f5bc2a0ac0240d>:0
  at at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.1(intptr,intptr)

```

### Steps to Reproduce

1. Open Xamarin.Essentials.sln.
2. Build the project Samples.Android.
3. Run it on an Android-6 device (API 23).

### Expected Behavior

The samples app runs without error.

### Actual Behavior

It fails with an exception, see above.

### Basic Information

- Version with issue: 1.6.0-pre2-1-g0dba844
- Last known good version: ?
- IDE: VS for Mac
- Platform Target Frameworks: 
  - Android: 10
- Affected Devices: Nexus 7

dimonovdd:
The app didn't run due to an "AppActions"

Fixes #1441

### DIFF
--- a/Samples/Samples/App.xaml.cs
+++ b/Samples/Samples/App.xaml.cs
@@ -72,17 +72,11 @@ namespace Samples
         {
             // Don't handle events fired for old application instances
             // and cleanup the old instance's event handler
-            try
+
+            if (Application.Current != this && Application.Current is App app)
             {
-                if (Application.Current != this && Application.Current is App app)
-                {
-                    AppActions.OnAppAction -= app.AppActions_OnAppAction;
-                    return;
-                }
-            }
-            catch (FeatureNotSupportedException ex)
-            {
-                Debug.WriteLine($"{nameof(AppActions)} Exception: {ex}");
+                AppActions.OnAppAction -= app.AppActions_OnAppAction;
+                return;
             }
 
             Device.BeginInvokeOnMainThread(async () =>
--- a/Samples/Samples/App.xaml.cs
+++ b/Samples/Samples/App.xaml.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AppCenter;
@@ -30,8 +31,14 @@ namespace Samples
 
             MainPage = new NavigationPage(new HomePage());
 
-            if (Device.RuntimePlatform != Device.macOS && Device.RuntimePlatform != Device.Tizen)
+            try
+            {
                 AppActions.OnAppAction += AppActions_OnAppAction;
+            }
+            catch (FeatureNotSupportedException ex)
+            {
+                Debug.WriteLine($"{nameof(AppActions)} Exception: {ex}");
+            }
         }
 
         protected override async void OnStart()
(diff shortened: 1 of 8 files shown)


mattleibow (approved):
Thanks for this!

janusw:
Note that there is an alternative (and more general) fix in #1442.

janusw (commented):
This change is not general enough: It fixes the problem on MacOS and Tizen, but not on old Android versions (see #1441).

jamesmontemagno:
This is the recommended route and just need to also check version numbers too before calling the API.

janusw:
> This is the recommended route and just need to also check version numbers too before calling the API.

I really don't think it is a good idea to further extend the explicit check in this PR, since the logic becomes quite involved and it just duplicates what is already encoded in `AppActions.PlatformIsSupported`. I can see two reasonable solutions here:

1. try-catch:
```
    try {
        // do some AppAction stuff here
    } catch (FeatureNotSupportedException) {
    }
```
2. using a public 'IsSupported' method (this is PR #1442)
```
    if (AppActions.IsSupported) {
        // do some AppAction stuff here
    }
```

dimonovdd:
> try {
>         // do some AppAction stuff here
>     } catch (FeatureNotSupportedException) {
>     }

@jamesmontemagno @mattleibow  @janusw 
I think this is the best fix for the problem in this case

Samples/Samples/App.xaml.cs
```diff
@@ -48,19 +56,33 @@ protected override async void OnStart()
...
+                Debug.WriteLine($"{nameof(AppActions)} Exception: {ex}");
+            }
         }
 
         void AppActions_OnAppAction(object sender, AppActionEventArgs e)
         {
             // Don't handle events fired for old application instances
             // and cleanup the old instance's event handler
-            if (Application.Current != this && Application.Current is App app)
+            try
```
mattleibow, line 75: This does not need to be wrapped in a try/catch as the actual registration check will catch it.

mattleibow (commented):
This looks good and I think we can merge this before #1442 so that at least it works for now.

Samples/Samples/App.xaml.cs
dimonovdd: I must be tired.

mattleibow (approved):

dimonovdd:
> /azp run
this failed. 

@mattleibow 
I don't have access to the logs. Is there anything I should fix?

jamesmontemagno (approved):

mattleibow:
Why is CI not starting?

mattleibow:
Thanks for this.

Issue #77, opened 2018-03-13
Feature/Capability Availability/Status API's


Redth:
There's been a lot of discussion around handling checking of device capabilities and feature availability at runtime.  This should be a central place to formulate our strategy around these types of API's.

Keep in mind we want to grow the API surface as conservatively as possible.  Once an API is public in our first stable release, it lives forever.

We also want to remember that the reason for adding any API's is to make developers' experiences delightful in consuming our library.

## Purpose

The purpose of any sort of `Status` or `Availability` query on a feature is to see if it's usable or not when called.  This means in almost every case, the developer doesn't really care about the actual _state_ of availability, but rather that they can _use_ the feature.

There's one exception to this: When something might be unavailable, but _could be made available_ through code (eg: prompting the user, etc).

NOTE: This is _not_ to be confused with missing or lack of granted permissions for a feature.

### Capabilities class

[One strategy](https://github.com/xamarin/Caboodle/issues/36) was to introduce a `Capabilities` class.  I dislike this strategy for a couple of reasons:

 1. It's separated from the actual class being used, and much less discoverable.  If I'm working with Location, if while I'm typing `Location.` I see `Location.Availability`, it might prompt me with the idea that I should or at least _could_ check to see if the feature works at runtime before using it.
 2. Capabilities is in a sense tightly coupled to pretty much every other class.

One suggestion is we could have a property in each class which just calls into the Capabilities class.  I also dislike this idea as it introduces ambiguity (we'll end up with people writing code like `if (Geolocation.IsAvailable && Capabilities.HasGeolocation)` to be safe.  It also means more code paths to test and maintain.

### Exceptions

Regardless of the strategy we use around checking for feature/capability availability, users won't be _required_ to check these before attempting to _use_ them, so we will still need to throw exceptions such as `NotSupportedException` at runtime if a particular API is unavailable at the time of invocation.

So, the exception strategy is to simply omit any API for checking availability and have the developer rely on exception handling to detect when things are not supported or available.  

The benefit to this approach is we can always add other API's for checking availability down the road as this exception work will be done regardless.

Where this approach becomes a bit messy is with statuses where a developer may be able to do something to enable or have the user enable a feature and try again.  For example, this becomes a bit messy:

```csharp
Position p;

try {
    p = await Geolocation.GetPositionAsync ();
} catch (GeolocationDisabledException) {
    // Prompt user to enable geolocation
    await DoSomethingToGetTheUserToEnable ();

    // Try getting location again
    try {
[...]

PR #1442, opened 2020-10-12
make the samples app run on Android 6, MacOS, Tizen

janusw:
### Description of Change ###

* make AppActions.IsSupported public
* use it in the Samples app to get rid of a FeatureNotSupportedException

### Bugs Fixed ###

- Related to issue #1441 

### API Changes ###

Changed:

 - `internal static bool AppActions.IsSupported` => `public static bool AppActions.IsSupported`

### Behavioral Changes ###

None.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/AppActions/AppActions.netstandard.tvos.watchos.macos.tizen.cs
+++ b/Xamarin.Essentials/AppActions/AppActions.netstandard.tvos.watchos.macos.tizen.cs
@@ -6,8 +6,7 @@ namespace Xamarin.Essentials
 {
     public static partial class AppActions
     {
-        internal static bool PlatformIsSupported
-            => throw ExceptionUtils.NotSupportedOrImplementedException;
+        internal static bool PlatformIsSupported => false;
 
         static Task<IEnumerable<AppAction>> PlatformGetAsync() =>
             throw ExceptionUtils.NotSupportedOrImplementedException;
--- a/Samples/Samples/App.xaml.cs
+++ b/Samples/Samples/App.xaml.cs
@@ -48,9 +48,12 @@ namespace Samples
                 typeof(Distribute));
             }
 
-            await AppActions.SetAsync(
-                new AppAction("app_info", "App Info", icon: "app_info_action_icon"),
-                new AppAction("battery_info", "Battery Info"));
+            if (AppActions.IsSupported)
+            {
+                await AppActions.SetAsync(
+                    new AppAction("app_info", "App Info", icon: "app_info_action_icon"),
+                    new AppAction("battery_info", "Battery Info"));
+            }
         }
 
         void AppActions_OnAppAction(object sender, AppActionEventArgs e)
--- a/Xamarin.Essentials/AppActions/AppActions.shared.cs
+++ b/Xamarin.Essentials/AppActions/AppActions.shared.cs
@@ -6,7 +6,7 @@ namespace Xamarin.Essentials
 {
     public static partial class AppActions
     {
-        internal static bool IsSupported
+        public static bool IsSupported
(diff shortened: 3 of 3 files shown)


janusw:
I just noticed that this PR is closely related to #1440 (which has a similar fix for MacOS), so I added another commit which fixes things on MacOS & co properly, so that this PR now includes fixes for Android, MacOS, Tizen etc.

IMHO the approach taken here is preferable over that in #1440 (because it's more general and more elegant) ...

Xamarin.Essentials/AppActions/AppActions.netstandard.tvos.watchos.macos.tizen.cs
```diff
@@ -6,8 +6,7 @@ namespace Xamarin.Essentials
 {
     public static partial class AppActions
     {
-        internal static bool PlatformIsSupported
-            => throw ExceptionUtils.NotSupportedOrImplementedException;
+        internal static bool PlatformIsSupported => false;
```
dimonovdd, line 9: This project uses NotSupportedOrImplementedException everywhere

janusw: Yes, it does, and that's perfectly fine for methods like `AppActions.PlatformGetAsync` or `AppActions.PlatformSetAsync`: If I'm trying to get or set the AppActions on platforms that don't support it, I should clearly get an exception. But `AppActions.PlatformIsSupported` is different: It is asking whether the platform supports AppActions, and I want a yes-or-no answer here, not an exception.

An alternative to making `AppActions.IsSupported` public would be to wrap any usage of `AppActions` in try/catch blocks, but since the `IsSupported` method is already there, I find it easier to use it directly. It produces more readable code (and is probably more performant).

dimonovdd: Look at Vibration.IsSupported, for example. It's like that everywhere

dimonovdd:
I like the approach with the "IsSupported" property.
Then we should have... have this property for each method of their API classes.
Or have a shared public API that determines whether the method is supported.

Sample:
`bool val1 = FilePicker.PickAsyncIsSupported;`
`bool val1 = FilePicker.PickMultipleAsyncIsSupported;`

Or:
`bool val1 = Essentials.IsSupported(nameof(FilePicker), nameof(PickAsync));`
`bool val1 = Essentials.IsSupported(nameof(FilePicker), nameof(PickMultipleAsync));`

jamesmontemagno:
We do no have the concept of IsSupported yet as the library is exception first.

To make the sample run I would recommend just checking the OS and versions it is running.

janusw:
> To make the sample run I would recommend just checking the OS and versions it is running.

Well, that is precisely what `AppActions.PlatformIsSupported` is doing already, nicely abstracted and separated by platform. Do you really want to duplicate this logic in the Samples app? That would be quite cumbersome. What's wrong about simply making `AppActions.IsSupported` public (as proposed above)?

jamesmontemagno:
We just havent' done taht for any API and we would want to create a single supported across all the APIs. We have an open ticket on that.

janusw:
> We just havent' done taht for any API and we would want to create a single supported across all the APIs. We have an open ticket on that.

Could you throw me a link or an issue number? If this PR is accepted, I'd volunteer to take care of generalizing the API as a follow-up ...

mattleibow:
This is the issue with the support API: https://github.com/xamarin/Essentials/issues/77

janusw:
Since #1440 was merged (and #1441 is fixed), I'll close this PR.

PR #1501, opened 2020-11-06
Don't throw if iOS can't pick multiple files

### LINKED ISSUE #1492: [Bug] App crashes on iOS 10 when the internal implementation attempts to set `AllowsMultipleSelection`
### Description
App crashes when calling `PickFileAsync` on iOS 10.

### Steps to Reproduce
1. Call `PickFileAsync(options)` on iOS 10.
2. App crashes with` UIDocumentPickerViewController.set_AllowsMultipleSelection (System.Boolean value)
SIGABRT: This API is not supported on this version of iOS`

### Expected Behavior
We should be wrapping [this line of code](https://github.com/xamarin/Essentials/blob/044c92dc18b44c5f37a0156c5de227a14a4317cd/Xamarin.Essentials/FilePicker/FilePicker.ios.cs#L32) in a version check but we are only checking the version at the beginning of the method when deciding whether to respect the `allowMultiple` parameter.

### Actual Behavior
We should be doing this:
```cs
if (Platform.HasOSVersion(11, 0))
    documentPicker.AllowsMultipleSelection = allowMultiple;
```
### Basic Information

- Version with issue: 1.6.0-pre2
- Platform Target Frameworks: 
  - iOS:  10.3.3 (any pre 11 version)

mattleibow:
### Description of Change ###

On pre-iOS 11, the `UIDocumentPickerViewController.AllowsMultipleSelection` did not exist.

Our previous solution was to throw, but this is inconsistent with all the other platforms. If an app on android can't pick multiple files, then it still works as the multiple flag is almost optional.

### Bugs Fixed ###

- Fixes #1492

### API Changes ###

None.

### Behavioral Changes ###

On older iOS versions, the app no longer throws an exception.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.ios.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.ios.cs
@@ -13,8 +13,7 @@ namespace Xamarin.Essentials
     {
         static Task<IEnumerable<FileResult>> PlatformPickAsync(PickOptions options, bool allowMultiple = false)
         {
-            if (allowMultiple && !Platform.HasOSVersion(11, 0))
-                throw new FeatureNotSupportedException("Multiple file picking is only available on iOS 11 or later.");
+            throw new FeatureNotSupportedException("Multiple file picking is only available on iOS 11 or later.");
 
             var allowedUtis = options?.FileTypes?.Value?.ToArray() ?? new string[]
             {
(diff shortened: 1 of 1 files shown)


jamesmontemagno (approved):

jamesmontemagno:
I should also probably document this :)

PR #1505, opened 2020-11-06
add new property DisplayInfo.RefreshRate

### LINKED ISSUE #1443: [Enhancement] DeviceDisplay/DisplayInfo should report refresh rate
## Summary

[DisplayInfo](https://docs.microsoft.com/de-de/dotnet/api/xamarin.essentials.displayinfo) already contains various information about the display, like width/height, density, orientation, rotation, etc. It should also include the device's screen refresh rate.

## API Changes

The DisplayInfo struct should obtain a new property called `RefreshRate` or similar:

    public readonly struct DisplayInfo : IEquatable<DisplayInfo>
    {
        [...]
        public double RefreshRate { get; }
        [...]
    }

Example usage:

    // Get Metrics
    var mainDisplayInfo = DeviceDisplay.MainDisplayInfo;
    
    // display refresh rate in Hz (i.e. "frames per second")
    var rate = mainDisplayInfo.RefreshRate;

## Intended Use Case

The refresh rate can be useful for example in connection with sensor update rates, which are configurable via `SensorSpeed` (or can be throttled manually). If one knows the screen refresh rate, one can make sharper constraints on the necessary sensor rates. E.g. if my screen updates only with 30 Hz, I probably don't need to sample the OrientationSensor with 50 Hz, because the information can not be displayed that quickly. Or the other way around: If I know that my screen updates with 60 or even 120 Hz, then I might want to increase the sensor updates accordingly, so that I can get a smoother display (e.g. think of an "augmented-reality"-type camera overlay).

## Implementation

Some pointers that might be useful for the implementation:
* Android: https://docs.microsoft.com/en-us/dotnet/api/android.views.display.refreshrate
* iOS: https://developer.apple.com/documentation/uikit/uiscreen/2806814-maximumframespersecond

janusw:
### Description of Change ###

This enhancement was proposed in #1443. It extends the `DisplayInfo` class with information about the display's refresh rate.

The new property has an implementation on Android, iOS and UWP. On MacOS it is always zero at present, because the corresponding Xamarin API is missing on MacOS (see also xamarin/xamarin-macios#9958).

I have tested the implementation on real hardware with Android 10 and iOS14.

### Bugs Fixed ###

- Fixes #1443 

### API Changes ###

Added: 
 
- `public float DisplayInfo.RefreshRate { get; }`
- constructor: `DisplayInfo(double width, double height, double density, DisplayOrientation orientation, DisplayRotation rotation, float rate)`

### Behavioral Changes ###

None.

### PR Checklist ###


Xamarin.Essentials/Types/DisplayInfo.shared.cs
```diff
@@ -5,13 +5,14 @@ namespace Xamarin.Essentials
     [Preserve(AllMembers = true)]
     public readonly struct DisplayInfo : IEquatable<DisplayInfo>
     {
-        public DisplayInfo(double width, double height, double density, DisplayOrientation orientation, DisplayRotation rotation)
+        public DisplayInfo(double width, double height, double density, DisplayOrientation orientation, DisplayRotation rotation, float rate)
```
mattleibow, line 8: We can't break the API, so we need to add an overload with this new parameter.

mattleibow (commented):
Looking really good. We just can't break API.

Xamarin.Essentials/Types/DisplayInfo.shared.cs
janusw: Good point. I did not think about this. In fact I considered the `DisplayInfo` constructor to be more of an 'internal' API, since it is usually not called directly by the end user. In a typical usage scenario, I guess the user would access the `DisplayInfo` only via `DeviceDisplay.MainDisplayInfo`, which calls the constructor internally.

But you are right, in principle the end user can also construct a `DisplayInfo` variable manually. I will fix it like you proposed. Thanks for the review!

janusw: Modified the code according to your suggestion and rebased ...

mattleibow (approved):

mattleibow (dismissed):
I tested on my Razer Phone and got that sweet extra Hz

mattleibow:
CI seems unhappy on some devices. Can't seem to repro it locally :(

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.android.cs
```diff
@@ -34,15 +34,16 @@ static bool PlatformKeepScreenOn
         static DisplayInfo GetMainDisplayInfo()
         {
             using var displayMetrics = new DisplayMetrics();
-            using var display = GetDefaultDisplay();
+            var display = GetDefaultDisplay();
```
mattleibow, line 37: Without this, it appears that Android 21-24 throws a ObjectDisposedException for the Display object

janusw:
> I tested on my Razer Phone and got that sweet extra Hz

Cool :)
I only tested it on standard 60 Hz devices so far ...

janusw:
Thanks a lot for adding a MacOS implementation, @mattleibow! I just ran Samples.Mac on a MacBook with Big Sur, but unfortunately I see a Refresh Rate of zero :/
Did you test it and could you observe a more meaningful value?

mattleibow:
@janusw I think I may have found a solution... But it doesn't work for my screen...

I saw this issue: https://github.com/glfw/glfw/issues/137
And this commit: https://github.com/glfw/glfw/commit/aab08712dd8142b642e2042e7b7ba563acd07a45

Maybe if you tweak some things it works for you?

janusw:
> @janusw I think I may have found a solution... But it doesn't work for my screen...
> 
> I saw this issue: [glfw/glfw#137](https://github.com/glfw/glfw/issues/137)
> And this commit: [glfw/glfw@aab0871](https://github.com/glfw/glfw/commit/aab08712dd8142b642e2042e7b7ba563acd07a45)
> 
> Maybe if you tweak some things it works for you?

Thanks for your efforts here, Matt! I tried your solution, but I still only get a RefreshRate of 0 on the MacBook.

I see `CVDisplayLinkInterop.GetRefreshRate` being called twice. The first time, all components of `period` are zero. The second time, `period.Flags` and `period.TimeScale` have rather odd non-zero values, but `period.TimeValue` is still zero. I have no idea why this is the case.

janusw:
Awesome! Thanks for merging 😃

PR #1508, opened 2020-11-08
Update Samples.iOS app to be compatible with App Store

### LINKED ISSUE #1467: [Bug] Samples.iOS is not fit for TestFlight / App Store
### Description

I recently wanted to make the XE Samples app (for iOS) available to a colleague via TestFlight (for testing/debugging purposes). Unfortunately one gets a number of errors when trying to upload the ipa package to App Store Connect.

### Steps to Reproduce

1. Open the Xamarin.Essentials solution in VS for Mac.
2. Build the Samples.iOS project and archive it for publishing.
3. Upload the ipa package to App Store Connect (e.g. via the Transporter app).

### Expected Behavior

The Samples.iOS app should follow Apple's rules for App Store distribution.

### Actual Behavior

There are several errors when verifying the package. I currently observe 11 errors, among them:
* several complaints about "Missing required icon files"
* several cases of "Invalid Image Path"
* "The following URL schemes found in your app are not in the correct format: [appcenter-APP_SECRET]."
* "Missing App Icon. An app icon measuring 1024 by 1024 pixels in PNG format must be included in the Asset Catalog"
* "Missing Info.plist value. A value for the Info.plist key 'CFBundleIconName' is missing in the bundle"

### Basic Information

- Version with issue: main branch as of today (043e81e53f12015b2ee926da9da37de023830447)
- Last known good version: ?
- IDE: VS for Mac
- Platform Target Frameworks: 
  - iOS:  14

janusw:
### Description of Change ###

This PR fixes issue #1467 by dealing with all problems that are reported when trying to upload the Samples.iOS app to App Store Connect, including:
* several issues with the app icons
* a problem with an URL scheme (related to App Center)
* deprecated API usage (UIWebView)

### Bugs Fixed ###

- Fixes #1467

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###


Samples/Samples.iOS/Info.plist
```diff
@@ -28,39 +28,21 @@
...
-		<string>Default-Portrait</string>
-		<string>Default-Portrait@2x</string>
-		<string>Icon-Small-40</string>
-		<string>Icon-Small-40@2x</string>
-		<string>Icon-Small-40@3x</string>
-		<string>Icon-Small</string>
-		<string>Icon-Small@2x</string>
-		<string>Icon-Small@3x</string>
-	</array>
+	<string>0</string>
```
mattleibow, line 31:
```suggestion
	<string>1</string>
```

Samples/Samples.iOS/Info.plist
```diff
@@ -28,39 +28,21 @@
...
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
 	<key>CFBundleName</key>
 	<string>Xamarin.Essentials</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>1.6.0</string>
```
mattleibow, line 37:
```suggestion
	<string>1.0</string>
```

mattleibow (commented):
Looks good. Maybe we can keep the same version numbers of "1.0" so we don't have to update it for each release.

Also, I see a comment in the issue about the `CFBundleIconName`. And, where are the "several cases of "Invalid Image Path"" issues resolved?

janusw:
> Looks good. Maybe we can keep the same version numbers of "1.0" so we don't have to update it for each release.

Yeah, well, that is kind of optional from my side. I thought it would be better to update it to the current version now, so that it's somewhat less outdated, even if it will not be updated for every release.

> Also, I see a comment in the issue about the `CFBundleIconName`.

I think that was fixed by the migration to the asset catalog (see 544c9709ffe636a79db4f7cc2bdc0369d9bcb6ed), which removed the `CFBundleIconFiles`.

> And, where are the "several cases of "Invalid Image Path"" issues resolved?

In 4166093941376865ae1c7bfd3334fd4020a8ca3f. `Info.plist` contained a few references to icons that don't exit.

Samples/Samples.iOS/Info.plist
janusw: I don't have a strong opinion on this. "0" is a valid build number for iOS builds, at least (in contrast to Android, where this is not the case). If you're not planning to update the version and build number, it has not much meaning anyway.

Samples/Samples.iOS/Info.plist
janusw: I'd prefer `1.6.0` here, but I can live with `1.0` as well.

One could also think about setting up a script that updates all version numbers in the project and config files for a release ...

Issue #1514, opened 2020-11-10
[Bug] MediaPicker CapturePhotoAsync - Orientation of photo rotated 90 degrees


tomaswinston:
### Description
After calling `CapturePhotoAsync` and setting the ImageSource to an Image, the resulting image is flipped 90 degrees (sideways) rather than retaining the orientation that the photo was taken in.
### Steps to Reproduce

1. Call CapturePhotoAsync with device in portrait.
2. Set resulting image to image source of Image element
3. Image is sideways

### Expected Behavior
I would expect that the photo retains the orientation the device was in when the photo was taken.
### Actual Behavior
The photo is 90 degrees rotated
### Basic Information

- Version with issue: 1.6.0-pre
- Last known good version:
- IDE: Visual Studio for Mac
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

AndrewBohdanovich:
any updates?

groege:
As described in #1640: RotateImage flag should be implemented.
As far as I understand the problem ist, that the image is stored ignoring the phones rotation and only the exif data in the picture contains the necessary rotation...
Therefore we need the option RotateImage which should rotate the image according to the exif data and remove the rotation data from exif.
Also it should be possible to retain Exif data as already requested in multiple duplicate issues (see #1640 for an overview)

ColdstarJoey:
Any update on this at all ?

mmunchandersen:
Did anyone develop a media picker that handles the rotation bug?

nicoriff:
Is this issue being treated?. I implemented this library and resulted to be worst than what I was using before.

AndrewBohdanovich:
i'm not sure about actuality of this information https://stackoverflow.com/questions/10850184/ios-image-get-rotated-90-degree-after-saved-as-png-representation-data but if we save photo as PNG on iOS with using UIImagePickerController we reproduce this bug

nicoriff:
No time to lose. I ended up going to Xam.Plugin from @jamesmontemagno. That library is still working fine.

mmunchandersen:
@nicoriff does it work on iOS14?

nicoriff:
@mmunchandersen Yes in iOS 14.4 works great.

teymur371:
It seems I'm not the only one moving back to Xam.Plugin

javanjoel:
[...]

PR #1515, opened 2020-11-11
Correct image rotation when taking a photo on iOS

Confusedfish:
### Description of Change ###

When taking a photo on iOS the resulting image is rotated by 90 degrees. This code change uses the same technique as this RotateImage method https://github.com/jamesmontemagno/MediaPlugin/blob/4b57e71d9b242ec7ae2d2bd4758f77e4535bc011/src/Media.Plugin/iOS/MediaPickerDelegate.cs#L743

No unit tests added as no existing tests for media picker in solution, presumably it is not possible to unit test the action of taking  a photo.

### Bugs Fixed ###

Relates to issue #1514 whereby when taking a photo on iOS the resulting image is incorrecrtly rotated by 90 degrees. 

### API Changes ###

None

### Behavioral Changes ###

Taking a photo may take slightly longer as the image manipulation is applied

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
+++ b/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Threading.Tasks;
+using CoreGraphics;
 using Foundation;
 using MobileCoreServices;
 using Photos;
@@ -116,9 +117,14 @@ namespace Xamarin.Essentials
             if (phAsset == null || assetUrl == null)
             {
                 var img = info.ValueForKey(UIImagePickerController.OriginalImage) as UIImage;
+                var imgUrl = info.ValueForKey(UIImagePickerController.ImageUrl) as NSString;
 
                 if (img != null)
-                    return new UIImageFileResult(img);
+                {
+                    var rotatedImg = CorrectImageRotation(img);
+
+                    return new UIImageFileResult(rotatedImg ?? img);
+                }
             }
 
             if (phAsset == null || assetUrl == null)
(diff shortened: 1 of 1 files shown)


Confusedfish:
Sorry guys. I am trying to help but I am unclear as to why this build is failing and do not have access to the Azure DevOps to investigate the cause.

Confusedfish:
Given the test failure is occuring on other pull requests I am assuming the failure is not related to this pull request and is more likely related to the test environment? The failing Android tests passes locally without issue.

jamesmontemagno:
I am not 100% sure if we want to start rotating images. We need proper exif data and then the library needs to read and rotate. I am very scared of any of my code in my library to be honest with you.

Confusedfish:
Hi @jamesmontemagno. Your solution works lovely in my project 👍  Unfortunately I cannot find an easy way of doing this outside of the Xaramin.Essentials library as by that point the image has been saved. It would be inefficient to read the file back check the EXIF, rotate it and then save it again. As I read things your code is doing the rotation before saving the file which avoids the JPG recompression too presumably. The issue is specific to iOS and you are reading the attributes from the UIImage instance, not sure if that has EXIF at that point?

Confusedfish:
I could add a property to the MediaPickerOptions to allow for easily enabling/disabling the rotation if you would rather it not be on by default in the case that it caused problems?

BjornVanslembrouck:
@Confusedfish, @jamesmontemagno,

I think instead of rotating the image or even creating a method that allows the user to rotate the image, the package might want to use the UIGraphics Context to get the image as is and create the stream from that.
By using this method, it seems that whatever orientation the image was taken in, it now returns it correctly.

Tested this on iOS 12, 13 and 14 by taking pictures in different orientations.

I create a PR for this: https://github.com/xamarin/Essentials/pull/1520

Confusedfish:
Closing this PR as #1520 is better

Issue #1509, opened 2020-11-09
[Bug] MediaPicker CapturePhotoAsync and Camera permission denied inconsistency


pablolop002:
### Description
(As Xamarin.Essentials documentation says, Essentials itself takes care of permissions request)

With camera permission set on: deny & don't ask again.
On iOS: When you invoke CapturePhotoAsync it opens CameraUI and all works (without image due to denied permission). No exception is thrown when you click deny on permission request

On Android: When you invoke CapturePhotoAsync it trows a Xamarin.Essentials.PermissionException (Camera permission was not granted). This exception is throw too when you click deny on permission request

### Steps to Reproduce

1. Create a Xamarin.Forms app with Essentials.
2. Add a button that uses MediaPicker.CapturePhotoAsync() (Executing on main thread)
3. Debug app and deny camera permission

### Expected Behavior

Consistent response. Maybe both should throw an exception because open CameraUI without camera permission isn't logic.

### Actual Behavior

iOS ignore denied permission and opens cameraUI

Android throws exception

### Basic Information

- Version with issue: 1.6.0-pre3
- Last known good version: ---
- IDE: Visual Studio for mac / Jetbrains Rider
- Platform Target Frameworks:
  - iOS: 14.1
  - Android: 10
  - UWP:  ---
- Android Support Library Version: ---
- Nuget Packages: Xamarin.Forms, Xamarin.Essentials
- Affected Devices: Pixel 3 simulator, iPhone 11 Pro

PR #1517, opened 2020-11-12
[Issue-1509] Change permission request for Photo and Camera

BjornVanslembrouck:
### Description of Change ###
Explicitly check and ask for permissions as since iOS 11 the UIImagePickerController normally takes care of this. This however can lead to the camera view being open while a user denies the permission and it just showing a black screen. Might be less confusing if it lets the user know that the permission was not granted so they can ask for it again. 

Divided the checks in to camera and photo separately. It could be somebody blocking access to their Photo Library but wanting to access the camera. Therefor check what action they want to perform and check the permission accordingly.

### Bugs Fixed ###

- Related to issue #1509 

### API Changes ###

None

### Behavioral Changes ###

Instead of seeing the UIImagePickerController appearing behind the alert asking for permission and potentially resulting in a black screen, it now asks for the permission before opening UIImagePickerController

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
+++ b/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
@@ -40,11 +40,11 @@ namespace Xamarin.Essentials
             if (!photo)
                 await Permissions.EnsureGrantedAsync<Permissions.Microphone>();
 
-            // permission is not required on iOS 11 for the picker
-            if (!Platform.HasOSVersion(11, 0))
-            {
+            // Check if picking existing or not and ensure permission accordingly as they can be set independently from each other
+            if (pickExisting)
                 await Permissions.EnsureGrantedAsync<Permissions.Photos>();
-            }
+            else
+                await Permissions.EnsureGrantedAsync<Permissions.Camera>();
 
             var vc = Platform.GetCurrentViewController(true);
 
--- a/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
+++ b/Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
@@ -41,9 +41,10 @@ namespace Xamarin.Essentials
                 await Permissions.EnsureGrantedAsync<Permissions.Microphone>();
 
             // Check if picking existing or not and ensure permission accordingly as they can be set independently from each other
-            if (pickExisting)
+            if (pickExisting && !Platform.HasOSVersion(11, 0))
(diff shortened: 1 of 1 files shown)


Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
```diff
@@ -40,11 +40,11 @@ static async Task<FileResult> PhotoAsync(MediaPickerOptions options, bool photo,
             if (!photo)
                 await Permissions.EnsureGrantedAsync<Permissions.Microphone>();
 
-            // permission is not required on iOS 11 for the picker
-            if (!Platform.HasOSVersion(11, 0))
```
mattleibow, line 44: Is there a reason why this line was removed? It shows a popup on newer version which basically ask whether the app can access the specific photo or all the photos. We are never going to need all the photos, so we don't need to ask in this case.

mattleibow:
Thanks for this PR. This change is very good, so thanks for taking the time.

Xamarin.Essentials/MediaPicker/MediaPicker.ios.cs
BjornVanslembrouck: @mattleibow, to be honest, I didn't really understood why it was there in the first place. Looks like you only wanted to verify the permissions on devices lower than iOS11 and that only for photo's and not the camera. 
I assumed permissions should be asked on all supported iOS versions (I tested from 10.3 to 14.1) and it seem to be doing the same on all of them.

I know that when opening a `UIImagePickerController` on later iOS versions, it asks for the permissions when opening the controller, but that could potentially lead to the same problems as opening the camera, denying the permissions and only seeing a black screen that has to be dismissed manually.

The way I implemented this, you are always sure that the permissions you need are granted before even trying to pick or capture media.

If I'm missing something, please do let me know.

mattleibow: Ah, with the photo picker, the permissions are automatically granted on the newer devices so we don't have to explicitly ask for them.

Basically, for the picker, we need to ask before 11, but for the camera, we need to always ask.

BjornVanslembrouck: I see, I will admit I had to read the documentation a couple of times to understand it properly...
Updated the PR to ask for photo permissions when it is `pickExisting AND < iOS11`, but always ask if it is `!pikcExisting`

PR #1520, opened 2020-11-14
Issue 1514 - MediaPicker CapturePhotoAsync - Orientation of photo rotated 90 degrees

BjornVanslembrouck:
### Description of Change ###

When a picture was taken in Portrait mode, the image got rotated 90 degrees. By using the UIGrapics Context to create the stream, the image is now returned as the user has taken it. Portrait and Landscape are now both orientated correctly.

### Bugs Fixed ###

- Related to issue #1514 

### API Changes ###

None

### Behavioral Changes ###

Image now returns in the format the user took the image. It does not matter which angle (0, 90, 270) the image was take in, the UIGrapics Context will take this into account before returning the Stream.

### PR Checklist ###


vniehues:
I just had this issue on an android device. 
This PR will only affect iOS, correct?

Confusedfish:
Thanks @BjornVanslembrouck I just tested this in my solution. It worked and is a much simpler change than the one in my PR copied from @jamesmontemagno 's work.

Confusedfish:
> I just had this issue on an android device.
> This PR will only affect iOS, correct?

@vniehues this code change is iOS specific and will not change the behaviour on Android.

BjornVanslembrouck:
@vniehues, which version of Android / device / manufacturer was this on?

jamesmontemagno:
I think don't we want to check if it is a png first and save as png else check if jpg and save as jpg? it is tricky.

This does look like a better solution. I feel like we shoudl rename this method though...

Does exif data get stripped? did it before?

BjornVanslembrouck:
We could get the UTType from the underlaying image data and based on that type, convert it to PNG or JPEG.

Renaming the method might be tricky as well as it follows the method override rules that get used everywhere.

Not an expert on image processing, but what I found/read online was that a PNG normally doesn't hold any EXIF data. It is possibly to add it, but apparently that is all fairly new. A JPEG does hold some data, but that is only a limited set apparently.
Returning a JPEG (instead of the default PNG it does now) does seem to hold the basic EXIF data like orientation, size, ColorProfile, ...

Is it worth merging this in for now as it fixes the bug for the rotation and open a new one for Android and maybe a new story for the renaming, the exif data and the jpeg/png selection? 
This merge won't change anything to the file type and exif data that was already in the package. It always stored and read the image as a PNG and that hasn't changed.

ahmad-crossplatform:
Any workaround for this ?

BjornVanslembrouck:
@ahmad-crossplatform, in another PR James mentioned it needed to be revised as the code in the MediaPlugin is a bit scary sometimes (https://github.com/xamarin/Essentials/pull/1515#issuecomment-726206906).
This PR fixes the 90 degree rotation by using Apple's own SDK and without manually rotating images, but there were then discussion of checking more and returning different things on the images.

Maybe you can request to merge this fix in for now and then we can open a different story for the extensions around file types and EXIF data.

Confusedfish:
How can we request this be merged in? It fixes my problem (image rotation) and it would be great to get back to referencing the nuget instead of including this project in my build

Confusedfish (approved):
Tested on a number of physical iOS devices and resolves the issue of image rotation when taking photos

peppelorum:
Who need to review this for it to be merged?
This bug is kinda of a show stopper right now.

jamesmontemagno:
I feel as though I would like to make it an extension method or some other method. In the current instance you will get the exact image back that you requested with Exif data I am pretty sure, however if you are creating a new on, I think you wont? Which would cause issues for other libraries that read exif data.. Assumption here from looking at the data. @mattleibow thoughts here?

IngweLand:
I vote against this pull request. Just made few tests:
1) Proposed changes do remove part of the metadata (tested with MetadataExtractor library by supplying the stream after the image was captured https://github.com/drewnoakes/metadata-extractor-dotnet ). Basically, ICC profile metadata is stripped.
2) Exporting the image as JPEG (`data ??= uiImage.AsJPEG();`) fixes rotation issue because it adds rotation info into the EXIF metadata `[Exif IFD0] Orientation - Right side, top (Rotate 90 CW)`
3) Regardless of the way, the was majority of metadata is not present at all. Most of EXIF data is not there. I think this part should be redone completely - having metadata in place is crucial in a lot of scenarios. MediaPlugin saves it without issues (including location info, which might be important)

IngweLand:
Created a proposal PR which should fix this and few other issues
https://github.com/xamarin/Essentials/pull/1639

javanjoel:
Soooo... another thing not fixed??

21r8390:
When will this bug be fixed?

dimonovdd:
@npostma https://github.com/CommunityToolkit/Maui/issues/11

21r8390:
> @npostma  Will xamarin.essentials still be available in MAUI? By my understanding a lot will still be 'the same' and still usable in MAUI (in a different namespace). So this is still valuable to get fixed.

I am not an official developer of Xamarin, but what I read from the [documentations and info ](https://devblogs.microsoft.com/xamarin/xamarin-essentials-1-7-and-introducing-net-maui-essentials/)is:

Xamarin.Essentials will become **Maui.Essentials**, that means all features will be transferred.
If you want to see the available ones you can check it out at this [GitHub](https://github.com/dotnet/maui/tree/main/src/Essentials).

npostma:
Thanks. That was my suspicion. So fixing this bug is really fixing MAUI.....
I've tried to work around this problem and tried to get EXIF information out of the image but no luck there. For now i gave the user the option to rotate the image himself after taking an image.

npostma:
Sure, use this snippet. I use skiasharp. Skiasharp seems nice to do all sorts of image modifications like rotating, translating, scaling, blurring etc. But note that this could lead to issues with Samsung S8(+). For some specific devices it does not work.

```
// iOS doesn't rotate the image to match the device orientation, so we have to rotate it 90 degrees here
using (SKCanvas surface = new SKCanvas(skTargetBitmap))
{
    // Hacky workarround for issue: https://github.com/xamarin/Essentials/issues/1640 
    if (Device.RuntimePlatform == Device.iOS)
    {
        if (doRotate)
        {
            // Make sure X/Y set in the middle of the image, otherwise the pivot point is off
            surface.RotateDegrees(degrees, skSourceBitmap.Width / 2, skSourceBitmap.Height / 2);
        }
    }
    surface.DrawBitmap(skSourceBitmap, 0, 0);
}
  ```

khambley:
would someone please review this and merge it! I'm waiting.... :)

angelru:
@jfversluis ping :)

jfversluis:
@angelru thanks for the ping, but I have no powers here :)

dimonovdd:
@jfversluis Thnx. I finally got the answer to that question

> @dimonovdd we had indeed plans to move some things out of Essentials and into the Toolkit. I think we abandoned that idea for now and everything that is in Essentials today will be in .NET MAUI Essentials as well.

jfversluis:
Unfortunately we won't be able to merge this, please see [here](https://github.com/xamarin/Essentials/issues/1514#issuecomment-1004314212).

Additionally, while this fixes a specific scenario, I don't think this is the desired way to do so. And the desired way will take more work which is something we can't make happen for Essentials anymore at this point in time. I'd love for you to join the discussion in the issue I linked above. Thanks!

PR #1534, opened 2020-11-19
Get the correct name from the components

### LINKED ISSUE #1288: [Bug] AppleSignInAuthenticator doesn't return name
### Description
AppleSignInAuthenticator doesn't return name even when it is returned from Apple services.

When debugging through source code the issue seems to be in this code: 
`            appleAccount.Properties.Add("name", NSPersonNameComponentsFormatter.GetLocalizedString(creds.FullName, NSPersonNameComponentsFormatterStyle.Default, NSPersonNameComponentsFormatterOptions.Phonetic));
`
Even when cred.FullName value is populated, NSPersonNameComponentsFormatter.GetLocalizedString converts it to null.

**Note: creds.FullName populated only once, when user first allows app to be used with Apple Id.**

### Steps to Reproduce

1. Use following call from iOS app registered with Apple Sign In on iOS 13
`r = await AppleSignInAuthenticator.AuthenticateAsync(new AppleSignInAuthenticator.Options() { IncludeFullNameScope = true });
`
2. Retrieve value from r.Properties["name"]

### Expected Behavior
The value should be populated with user's name.

### Actual Behavior
The value is null

### Basic Information

- Version with issue: 1.5.3.2
- Last known good version: None
- IDE: Visual Studio
- Platform Target Frameworks: 
  - iOS:  
- Nuget Packages: 
- Affected Devices: iOS devices

### Screenshots

Non visual

mattleibow:
### Description of Change ###

Get the correct name from the components. Previously we were retrieving the phonetic name, without first checking for it.  I am not sure we actually want to do that and we rather ant the name as it is specified.

### Bugs Fixed ###

- Fixes #1288

### API Changes ###

None.

### Behavioral Changes ###

The `name` is now correctly parsed from the sign in response.

### PR Checklist ###

### DIFF
--- a/Samples/Samples/ViewModel/WebAuthenticatorViewModel.cs
+++ b/Samples/Samples/ViewModel/WebAuthenticatorViewModel.cs
@@ -44,7 +44,14 @@ namespace Samples.ViewModel
                     && DeviceInfo.Platform == DevicePlatform.iOS
                     && DeviceInfo.Version.Major >= 13)
                 {
-                    r = await AppleSignInAuthenticator.AuthenticateAsync();
+                    // Make sure to enable Apple Sign In in both the
+                    // entitlements and the provisioning profile.
+                    var options = new AppleSignInAuthenticator.Options
+                    {
+                        IncludeEmailScope = true,
+                        IncludeFullNameScope = true,
+                    };
+                    r = await AppleSignInAuthenticator.AuthenticateAsync(options);
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


DamienDoumer:
Hello @mattleibow  I just tried the corrections on your PR and I still can't get the username or email. Please can you tell me if I missed something ?

mattleibow:
@DamienDoumer did you revoke access first? This data only comes on the very first sign in. Even an uninstall doesn't clear it.

I logged out by opening Settings, then clicking my Apple ID, then Password & Security, then Apps Using Apple ID.

DamienDoumer:
Ooh, I didn't revoke it. I though your PR added a means of getting the email and password no matter what.
I guess it is impossible to get the email and name consistently. I must delegate that to my backend

mattleibow:
Apple only provides the data once... Not sure why that is really necessary, but it could be because there is a popup that is editable. I would have just cached the response and always sent it. But I guess Apple wants you to do that.

mattleibow:
I pushed the PR build to the preview feed. Let me know if this now works.

Version: `1.6.0-pr.1534.1`
Feed: `https://aka.ms/xamarin-essentials-ci/index.json`

DamienDoumer:
I just revoked the apple signin for my app, and using your code directly, I could get the email and username. 
But it does so only the first time, and never again. As expected...

mattleibow:
Thanks for confirming. I'll merge and this will get pushed out to the nightly feed.

PR #1542, opened 2020-11-23
Ensure iOS 9 is supported on all OpenUrl calls.

### LINKED ISSUE #1535: [Bug] Email.ComposeAsync and Launcher.OpenAsync not support on iOS devices
### Description
Recently I migrated Xamarin.Forms from v3.6 to the latest stable v4.8. In the process, the old code that was used for opening email was Device.OpenUri("mailto:somemail@somedomain.com"). But this code, Device.OpenUri is deprecated as of Xamarin.Forms v4.8. So I replace it with Xamarin.Essentials' Email.ComposeAsync. The problem is that I still have to support devices running iOS 9, and I tried both Email.ComposeAsync as well as Launcher.OpenAsync but neither of them support iOS 9. Thus causing a bit of inconvenience here.
### Steps to Reproduce

1. Download/Clone the project from Reproduction Link
2. Run the application on an iOS 9 device.
3. Tap on each sample email provided on the initial screen.

### Expected Behavior
Can open the email app
### Actual Behavior
Can not open the email app
### Basic Information

- IDE: Visual Studio for Mac 2019
- Platform Target Frameworks:
  - iOS: 14.4.1.3
- Affected Devices: iOS devices running iOS 9

### Reproduction Link

(https://github.com/hungxtran/XamarinEssentials-EmailAndLauncherNotSupportiOS9)[https://github.com/hungxtran/XamarinEssentials-EmailAndLauncherNotSupportiOS9]

### Workaround

The only workaround I found is switch back to Device.OpenUri and add suppress warning. But this is not pretty. Someone might be thinking "Oh, it's 2020 and nobody is using iOS 9 anymore". But the fact is there still users who hasn't update their iDevices and there are specific cases that I need to deploy to them.

jamesmontemagno:
### Description of Change ###

Use OpenUrl with only 1 parameter only on < iOS 10 and update the rest of the code for consistentcy

### Bugs Fixed ###

- Fixes #1535

### Behavioral Changes ###

Did all checks against API 10 and above to make the right calls.

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/AppInfo/AppInfo.ios.tvos.watchos.macos.cs
+++ b/Xamarin.Essentials/AppInfo/AppInfo.ios.tvos.watchos.macos.cs
@@ -24,14 +24,8 @@ namespace Xamarin.Essentials
            => NSBundle.MainBundle.ObjectForInfoDictionary(key)?.ToString();
 
 #if __IOS__ || __TVOS__
-        static void PlatformShowSettingsUI()
-        {
-            var nsurl = new NSUrl(UIApplication.OpenSettingsUrlString);
-            if (Platform.HasOSVersion(10, 0))
-                UIApplication.SharedApplication.OpenUrlAsync(nsurl, new UIApplicationOpenUrlOptions()).ConfigureAwait(false);
-            else
-                UIApplication.SharedApplication.OpenUrl(nsurl);
-        }
+        static async void PlatformShowSettingsUI()
+            => await Launcher.OpenAsync(UIApplication.OpenSettingsUrlString);
 #elif __MACOS__
         static void PlatformShowSettingsUI()
         {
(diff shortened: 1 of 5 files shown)


mattleibow (approved):
Looks good. I just refactored the code a bit to use the Launcher in all the places we launch. No use duplicating code all over.

This should be fine since the Launcher API is very simple and slim.

BjornVanslembrouck:
@jamesmontemagno, @mattleibow, maybe a stupid question, but it seems like this is a fix for iOS versions lower than 10. Does Xamarin.Essentials actually supports this? According to the documentation it supports iOS10 or higher (https://docs.microsoft.com/en-us/xamarin/essentials/platform-feature-support).

If iOS9 is something that needs to be supported, does that mean bug fixes or additions need to be tested against iOS9 before creating a PR? If so and if I'm contributing, I can do that, but according to the docs, it doesn't support anything lower than iOS10.

mattleibow:
@BjornVanslembrouck for Essentials, we typically/officially support iOS 10+, but sometimes if it is an easy fix, like this one, then we make it work on iOS 9.

As a contributor, if you _can_ test on iOS 9, then you can give it a go and see what happens. If there is some super complex logic, then maybe we can just fall back on our iOS 10+. So, for your question, if you can test on iOS 9, then maybe run it, if not, then we can wait for any issues to come in and then we can investigate.

PR #1543, opened 2020-11-24
Add DeviceDisplay for Tizen

sung-su:
### Description of Change ###

Add DeviceDisplay for Tizen

display
display2

### PR Checklist ###


Samples/Samples.Tizen/Program.cs
```diff
@@ -13,6 +13,7 @@ protected override void OnCreate()
             base.OnCreate();
 
             MaterialComponents.Init(DirectoryInfo.Resource);
+            Xamarin.Essentials.Platform.MainWindow = MainWindow;
```
mattleibow, line 16: Maybe we should follow our current "init" pattern and do something along these lines:

```
Xamarin.Essentials.Init(MainWindow);
```

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.tizen.cs
```diff
@@ -0,0 +1,100 @@
...
+        internal static extern void ScreenLockDevicePowerReleaseLock(int type);
+
+        static bool screenLock = false;
+
+        static bool PlatformKeepScreenOn
+        {
+            get
+            {
+                if (!screenLock)
+                    ScreenLockDevicePowerReleaseLock(1);
```
mattleibow, line 22: What is this `1`? Maybe we should make it a descriptive const?

Samples/Samples.Tizen/Program.cs
mattleibow: I also see a lot of `Platform.MainWindow ?? GetStaticThing`. What is the difference in values or functionality when using either?

Xamarin.Essentials/Platform/Platform.tizen.cs
```diff
@@ -22,6 +22,8 @@ public static partial class Platform
         static TizenOrientationSensor orientationSensor = null;
         static MapService mapService = null;
 
+        public static ElmSharp.Window MainWindow { get; set; }
```
mattleibow, line 25: Like my first comment, I think this should be an Init method as it is consistent with the other platforms.

However, we _can_ (and probably should) have an `internal` property for the other APIs to use.

mattleibow (commented):
Just a question on whether we should go for a `MainWindow` property or an `Init(MainWindow )` method...

Samples/Samples.Tizen/Program.cs
sung-su: I will change to Init pattern.

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.tizen.cs
sung-su: `1` is used instead of an unmanaged enum value. I will change to a descriptive const.

Xamarin.Essentials/Platform/Platform.tizen.cs
sung-su: public method is ok ?

Xamarin.Essentials/DeviceDisplay/DeviceDisplay.tizen.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        static DisplayInfo GetMainDisplayInfo()
+        {
+            var display = Platform.MainWindow;
+            return new DisplayInfo(
+                width: display.ScreenSize.Width,
+                height: display.ScreenSize.Height,
+                density: display.ScreenDpi.X / (DeviceInfo.Idiom == DeviceIdiom.TV ? 72.0 : 160.0),
+                orientation: GetOrientation(),
+                rotation: GetRotation());
```
mattleibow, line 37: I see you re no longer using `Platform.GetFeatureInfo<int>("xxx")`. Was there a reason for originally using it?

sung-su: The value through `Platform.GetFeatureInfo<int>("xxx")` is a fixed value. 
I think the current way of using changing values is more appropriate.

mattleibow (approved):

mattleibow:
Looking good! Thanks again for the work.

sung-su:
@mattleibow  Why this patch is not included in 1.6.0 ?

mattleibow:
It should be. Let me see what happened. I see it was merged some time ago.

mattleibow:
I see it there (this is the NuGet): 

image

sung-su:
oh my mistake . thank you for check

Issue #1544, opened 2020-11-24
[Bug] iOS FilePicker.PickAsync never returns after picker is dismissed


1Cor125:
## Description
`FilePicker.PickAsync` never returns if the document picker controller is dismissed using a method other than tapping the Cancel button. 

### Steps to Reproduce

1. Make a call to `PickAsync`
2. In the app, dismiss the document picker by swiping it down on the iPhone, or clicking outside the bounds of the document picker on iPad (do not tap the Cancel button).

### Expected Behavior

The method should return null.

### Actual Behavior

The method never returns (it never sets a value on the `TaskCompletionSource`).

### Basic Information

- Version with issue: 1.6.x versions
- Affected Devices: iPhone or iPad

1Cor125:
I see that `WasCancelled` is implemented. So I need to investigate this further. Perhaps something else is going wrong. Closing this.

1Cor125:
So using the Xamarin Essentials app the method never returns unless the picker is closed using the Cancel button. So this is a problem. I'm just not sure how to fix it. Will keep investigating to see if I can come up with a recommendation.

mattleibow:
Is this still happening? @BjornVanslembrouck just fixed this in https://github.com/xamarin/Essentials/pull/1518

1Cor125:
Forgot to mention that this is on iOS 14.2

1Cor125:
I remember having to fix this on our own homemade file picker @mattleibow. I think what did it was handling the newer `DidPickDocumentAtUrls` and, of course, `WasCancelled` and that seemed to cover all the scenarios.

1Cor125:
Update @mattleibow I just tried my suggestion above locally and it did not work.

BjornVanslembrouck:
@mattleibow, I fixed it on the `ImagePicker` not sure if it happens on the `FilePicker` as well.
I'll try and have a look this evening.

BjornVanslembrouck:
@mattleibow, @1Cor125, I created a PR that should fix the issue: https://github.com/xamarin/Essentials/pull/1546.

Tested both the swiping down on iPhone and the clicking outside bounds on iPad and both return a canceled action now.

1Cor125:
This looks great! Thanks @BjornVanslembrouck and @mattleibow

PR #1546, opened 2020-11-24
[Issue 1544] Fix [iOS] FilePicker.PickAsync never returns after picker is dismissed

BjornVanslembrouck:
### Description of Change ###

`FilePicker.PickAsync` never returns if the document picker controller is dismissed using a method other than tapping the Cancel button.

### Bugs Fixed ###

- Related to issue #1544 

### API Changes ###

None

### Behavioral Changes ###

The library now returns null when a user dismisses the picker by swiping down.

### PR Checklist ###


Xamarin.Essentials/FilePicker/FilePicker.ios.cs
```diff
@@ -48,6 +48,14 @@ static Task<IEnumerable<FileResult>> PlatformPickAsync(PickOptions options, bool
                 }
             };
 
+            if (documentPicker.PresentationController != null)
+            {
+                documentPicker.PresentationController.Delegate = new PickerPresentationControllerDelegate
+                {
+                    PickHandler = urls => tcs.TrySetResult(Enumerable.Empty<FileResult>())
```
mattleibow, line 55: Technically, we don't need to do what I am about to type, but maybe for consistency, we should extract the handler from `PickerDelegate.PickHandler ` and then call it here. Maybe a local method might be good?

mattleibow:
I am not able to test the file picker selection. When I pick an image on iOS, nothing happens - no visual change and the picker stays up. I can cancel, and, `MediaPicker` works. I have an iPhone X... Am I holding it wrong?

BjornVanslembrouck:
@mattleibow, not sure what is happening there. I can use all the options and pick them as expected
FilePicker_iPhoneXs

Or do you mean something else?

mattleibow:
@BjornVanslembrouck I think it is an issue (somehow) due to running from Hot Restart on Windows. I booted my macOS up and now it works. So that was fun.

Xamarin.Essentials/FilePicker/FilePicker.ios.cs
BjornVanslembrouck: That is an option, however doing it this way is consistent with how it is already done in the `MediaPicker`. Both delegates have their own functionality which they are responsible for.

mattleibow: Sounds good.

I am actually looking at the media picker API and I see it doesn't handle exceptions... I'll do a PR on that separately.

mattleibow (approved):

mattleibow:
Thanks for jumping on this!

PR #1547, opened 2020-11-25
Redirect exceptions from the pickers into the TCS

mattleibow:
### Description of Change ###

Exceptions thrown in the callbacks will not reach the end user, and the task may not even return. To avoid this, wrap the callback with a try/catch and then make sure to set the exception on the tcs.

### Bugs Fixed ###

- Related to the 3 pickers: media, file and contact

### API Changes ###

None.

### Behavioral Changes ###

Exceptions are now caught and thrown from the task.

### PR Checklist ###


pictos (approved):
LGTM!

Issue #1551, opened 2020-11-29
[Enhancement] Need for FileSystem.AppResourcesDirectory


CartBlanche:
## Summary
On _iOS_ and _macOS_ the call to **Directory.GetCurrentDirectory ()**; resolves to the **Resources** directory while on Android it just resolves to **/**.

I need all of them to resolve consistently to the **Resources** directory under _iOS_/_macOS_ and to the **Assets** folder under Android because I have a subdirectory in those locations, that vary from app to app, so I need to traverse/recurse through them to load all the required files on startup.

Now Xamarin.Essentials has **FileSystem.OpenAppPackageFileAsync**, which is great if you know the exact name of the file you want to load, but not if you want to traverse/recurse sub-directories.

## API Changes
I would like a `string FileSystem.AppResourcesDirectory` property that just resolves to **Resources** directory under _iOS_/_macOS_ and to the **Assets** folder under Android and the respective equivalents under Windows, Tizen and elsewhere.

## Intended Use Case
I need all of them to resolve consistently to the **Resources** directory under _iOS_/_macOS_ and to the **Assets** folder under Android because I have a subdirectory in those locations, that vary from app to app, so I need to traverse/recurse through them to load all the required files on startup.

If there is already a cross-platform way to do this, apologies for the waste of bandwidth :)

jamesmontemagno:
Q: Would it be better to do a PR into Mono/.NET for .NET 6 to itnroduce the change for Android?

CartBlanche:
Hi @jamesmontemagno, Hope you and your family are well during this crazy year.
Yes, I had considered that but after chatting to the Android team, it seems that might not be possible as all resource/asset handling on Android is handled by the AssetManager and not via a direct directory structure, I'm told. I'm still waiting to hear back from 1 more person on the .NET side.

So in the mean time I've been working on a possible PR for Essentials, if that sort of thing is of interest and would be a different proposition to what I suggested above. Instead the API would be something like...

`string[] FileSystem.AppResourceDirectories(string path)`
and
`string[] FileSystem.AppResourceFiles(string directoryPath)`

Therefore under the hood, they would do the correct platform specific calls to return their respective string arrays. 
It would also work under Android as the call to `Assets.List(path)` returns a string[] too and for iOS/Mac they could use the Directory.* equivalents, that we know work.

Do you think this would be useful to Essentials, or should I keep the changes to myself?

jamesmontemagno:
[...]

PR #1566, opened 2020-12-03
Add API to retrieve directories and file lists, in a cross-platform way.

CartBlanche:
### Description of Change ###

Given a Directory name,  respectively return either the sub-directories or files within that directory

### Bugs Fixed ###

- Related to issue https://github.com/xamarin/Essentials/issues/1551

### API Changes ###

Added: 

```csharp
public static partial class FileSystem {
    .
    .
    public string[] FileSystem.GetAppResourceDirectories(string path);
    public string[] FileSystem.GetAppResourceFiles(string path);
}
```

And associated cross-platform code.

### Behavioural Changes ###

None. New functionality.

### PR Checklist ###

### Sample UI Changes ###
**Android:**
Screenshot_1606994851

**iOS:**
Screenshot 2020-12-03 at 11 22 29

**Mac:**
Screenshot 2020-12-03 at 11 20 23

**Tizen:**
Tizen

**Windows.UWP:**
Windows

### DIFF
--- a/Xamarin.Essentials/FileSystem/FileSystem.android.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.android.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Android.Provider;
 using Android.Webkit;
@@ -52,6 +53,17 @@ namespace Xamarin.Essentials
                 throw new FileNotFoundException(ex.Message, filename, ex);
             }
         }
+
+        // Android does't have the concept for returning Directories (AFICT).
+        // As such, we only return string with no . in them. Not ideal but
+        // should work in most cases as Folders rarely contain dots
+        static string[] PlatformGetAppResourceDirectories(string path)
+            => Platform.AppContext.Assets.List(path).Where(file => !file.Contains(".")).Select(x => Path.Combine(path, x)).ToArray();
+
+        // This may also return Directories, but we can't restrict the list,
+        // as some files may not have an extention. So user is on their own from here.
+        static string[] PlatformGetAppResourceFiles(string path)
+            => Platform.AppContext.Assets.List(path).Select(x => Path.Combine(path, x)).ToArray();
 
         internal static Java.IO.File GetEssentialsTemporaryFile(Java.IO.File root, string fileName)
         {
--- a/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.macos.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.macos.cs
@@ -38,6 +38,22 @@ namespace Xamarin.Essentials
             }
             return dirs[0];
         }
+
+        static string[] PlatformGetAppResourceDirectories(string path)
+        {
+            if (Directory.Exists(path))
+                return Directory.GetDirectories(path);
+            else
+                return null;
+        }
+
+        static string[] PlatformGetAppResourceFiles(string path)
+        {
+            if (Directory.Exists(path))
+                return Directory.GetFiles(path);
+            else
+                return null;
+        }
     }
 
     public partial class FileBase
--- a/Xamarin.Essentials/FileSystem/FileSystem.uwp.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.uwp.cs
@@ -26,10 +26,10 @@ namespace Xamarin.Essentials
             => path.Replace('/', Path.DirectorySeparatorChar);
 
         static string[] PlatformGetAppResourceDirectories(string path)
-            => Directory.GetDirectories(Path.Combine(Application.Current.InstalledLocation, NormalizePath(path)));
+            => Directory.GetDirectories(Path.Combine(Package.Current.InstalledLocation.DisplayName, NormalizePath(path)));
 
         static string[] PlatformGetAppResourceFiles(string path)
-            => Directory.GetFiles(Path.Combine(Application.Current.InstalledLocation, NormalizePath(path)));
+            => Directory.GetFiles(Path.Combine(Package.Current.InstalledLocation.DisplayName, NormalizePath(path)));
     }
 
     public partial class FileBase
--- a/Xamarin.Essentials/FileSystem/FileSystem.uwp.cs
+++ b/Xamarin.Essentials/FileSystem/FileSystem.uwp.cs
@@ -26,10 +26,20 @@ namespace Xamarin.Essentials
             => path.Replace('/', Path.DirectorySeparatorChar);
 
         static string[] PlatformGetAppResourceDirectories(string path)
-            => Directory.GetDirectories(Path.Combine(Package.Current.InstalledLocation.Path, NormalizePath(path)));
+        {
+            if (Directory.Exists(path))
+                return Directory.GetDirectories(NormalizePath(path));
+            else
+                return null;
+        }
 
         static string[] PlatformGetAppResourceFiles(string path)
-            => Directory.GetFiles(Path.Combine(Package.Current.InstalledLocation.Path, NormalizePath(path)));
+        {
+            if (Directory.Exists(path))
+                return Directory.GetFiles(NormalizePath(path));
+            else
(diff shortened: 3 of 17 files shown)


CartBlanche:
I was unable to test the iOS Tester due to this provisioning blocker
Screenshot 2020-12-03 at 11 38 41

I believe My provisioning is correct as I've recently built and Code signed MacOS apps.

CartBlanche:
Should I raise an issue about the theming in iOS Sample? This is in the screenshot above but I'm adding it here to highlight it here. Basically we are getting white text on a white background. Interestingly the 1st screen that lists all the Essentials seems to be using dark mode, but the next screens like this one, on iOS don't...
Screenshot 2020-12-03 at 11 22 29

CartBlanche:
@mattleibow, @jamesmontemagno Finally borrowed my son's Windows machine to test this out and the sample, builds and runs on there too now. I think this is ready for you gents to take a look.

CartBlanche:
Looks like a timeout failure on Android 29. While on MacOS Cake.Core is blowing up. 
If there is anything I can do to fix either of these error, let me know.

CartBlanche:
@mattleibow Happy New Year! I was going to rebase this branch/PR, but seems I've had my permissions revoked. Does that mean I need to fork and submit PRs that way in future?

mattleibow:
Ah, you might need to use a fork then. I don't think there is anything specific blocking, just haven't had a chance to test the idea in some corner cases that I don't know about yet. I'm trying to merge PRs now, so I'll be looking at this.

Xamarin.Essentials/FileSystem/FileSystem.android.cs
```diff
@@ -53,6 +54,17 @@ static Task<Stream> PlatformOpenAppPackageFileAsync(string filename)
             }
         }
 
+        // Android does't have the concept for returning Directories (AFICT).
+        // As such, we only return string with no . in them. Not ideal but
+        // should work in most cases as Folders rarely contain dots
+        static string[] PlatformGetAppResourceDirectories(string path)
+            => Platform.AppContext.Assets.List(path).Where(file => !file.Contains(".")).Select(x => Path.Combine(path, x)).ToArray();
```
mattleibow, line 61: Can we wrap this for easier reading?

Xamarin.Essentials/FileSystem/FileSystem.android.cs
```diff
@@ -53,6 +54,17 @@ static Task<Stream> PlatformOpenAppPackageFileAsync(string filename)
...
+
+        // This may also return Directories, but we can't restrict the list,
+        // as some files may not have an extention. So user is on their own from here.
+        static string[] PlatformGetAppResourceFiles(string path)
+            => Platform.AppContext.Assets.List(path).Select(x => Path.Combine(path, x)).ToArray();
```
mattleibow, line 66: Maybe wrap this too.

Xamarin.Essentials/FileSystem/FileSystem.android.cs
```diff
@@ -53,6 +54,17 @@ static Task<Stream> PlatformOpenAppPackageFileAsync(string filename)
...
         }
 
+        // Android does't have the concept for returning Directories (AFICT).
+        // As such, we only return string with no . in them. Not ideal but
+        // should work in most cases as Folders rarely contain dots
+        static string[] PlatformGetAppResourceDirectories(string path)
+            => Platform.AppContext.Assets.List(path).Where(file => !file.Contains(".")).Select(x => Path.Combine(path, x)).ToArray();
+
+        // This may also return Directories, but we can't restrict the list,
+        // as some files may not have an extention. So user is on their own from here.
```
mattleibow, line 64: Is there no way to detect this? Not ending in a slash or something?

Xamarin.Essentials/FileSystem/FileSystem.ios.tvos.watchos.macos.cs
```diff
@@ -38,6 +38,22 @@ static string GetDirectory(NSSearchPathDirectory directory)
             }
             return dirs[0];
         }
+
+        static string[] PlatformGetAppResourceDirectories(string path)
+        {
+            if (Directory.Exists(path))
```
mattleibow, line 44: What happens if I change my current directory? Does it always work, or maybe we need to first get the package directory and then append? What if I pass a rooted path, will that still be inside the app resources dir?

Xamarin.Essentials/FileSystem/FileSystem.uwp.cs
```diff
@@ -24,6 +24,22 @@ static Task<Stream> PlatformOpenAppPackageFileAsync(string filename)
 
         internal static string NormalizePath(string path)
             => path.Replace('/', Path.DirectorySeparatorChar);
+
+        static string[] PlatformGetAppResourceDirectories(string path)
+        {
+            if (Directory.Exists(path))
```
mattleibow, line 30: Should this also be normalized?

mattleibow: Also, what happens if I pass a rooted path or change the cwd? Should we be checking for the location first?

Xamarin.Essentials/FileSystem/FileSystem.android.cs
mattleibow: Android is dodgy and it seems if I pass in `/` then I might get the root of the APK or something based on SO. I tested with Android 30 and I get an empty result. Which is wrong. If you use `""`, then you get the root. Might be worth adding some checks to first make sure that the path coming in is translated to the Android ways.

However, Android is bananas (probably fitting since we support the mono monkey) because Android adds junk:

```
> Platform.AppContext.Assets.List("");
{string[4]}
	[0]: "FileSystemTemplate.txt"
	[1]: "Folder"
	[2]: "images"
	[3]: "webkit"
```

This guy person reports that is also has a `sounds` dir which I don't have: https://stackoverflow.com/a/8366081 However, we cannot filter any of them out since that really is a valid folder and I can add things to it... I can't see the folder in the APK, but I can add things to it and then it exists. Bonkers. It must be a pseudo file structure that is merged with the real APK.

mattleibow: I also don't think we can use the filter of "contains period" because I can create an extensionless file. Maybe if we switch to EnumerateFiles, we can also do a `Assets.List("<filename>")` and see if it returns anything:

```
var dirs = Assets.List(path);
foreach (var dir in dirs) 
    if (Assets.List(dir).Length > 0)
        yield return dir
    else
        // this is a file
```

And we can do the reverse for the EnumFiles

mattleibow (changes requested):
This AIP is a bit ore complex based on things I mention below and in my comments. Let me know your thoughts and the things that didn't work for you.

1. The name of the method should maybe be `GetAppPackage[Files|Directories]` as the "package" idea is what the existing AIP uses.
2. We should probably work similarly to the `Directory` API does
   https://docs.microsoft.com/dotnet/api/system.io.directory.getfiles
   https://docs.microsoft.com/dotnet/api/system.io.directory.getdirectories
   1. Throw null exceptions for bad paths/incorrect paths/incorrect file/folder types.
   2. Throw IO exception the GetDirectories was a file
   3. Return an array of the full names (including paths) for the items in the specified directory, or an empty array if no items are found.
3. I notice that on some cases you assume the dev is a good guy and won't pass in rooted paths or change the working dir.
4. I would also suggest using the EnumerateFiles/Directories instead of Get and returning an `IEnumerable<string>`. If the result list is large, then there will be a significant delay. We can always have another method that does ToArray adjacent.

mattleibow:
If you can't work on this branch, then just open a new PR and link from here.

CartBlanche:
@mattleibow Thanks for the feedback. Looks like I may have to send a new PR from my fork. I'm currently in the middle of finishing and releasing my own app (hence the TTS PR), so won't get to resubmit a PR until next week. I definitely will revisit though, because as I mentioned this code is currently being used in a friend's language app on iOS, Android, Mac and they'll be releasing on  Windows by end of Feb.

CartBlanche:
@mattleibow I'll close this PR as I've just created a draft PR from my fork. So future changes will be from there.

Issue #1592, opened 2020-12-16
[Bug] iOS: Launcher.OpenAsync() Bugs out


Edgaras91:
### Description
Using Xamarin.Forms, iOS device OpenAsync overlay that appears in iPhone bugs out.
Other apps do below without any issues.
File is  .xlsx format.
### Steps to Reproduce

1. Save file to iPhone using this:
```
            string path = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
            string filePath = Path.Combine(path, filename.xlsx);
            try
            {
                FileStream fileStream = File.Open(filePath, FileMode.Create);
                fileStream.Write(bytes);
                fileStream.Flush();
                fileStream.Close();
            }
```
2. Open File: 
`await Launcher.OpenAsync(new OpenFileRequest { File = new ReadOnlyFile(filePath) });`
3. Select "Save to Files"
4. Press Cancel
5. Select "Save to Files" again
6. It will bug out - not go in to "Save to Files", screen flash and stay where it is

### Expected Behavior
To be able to "Save to Files" when you go back and forth in the menu.
### Actual Behavior
Not able to "Save to Files" on 2nd attempt in the same overlay session
### Basic Information

- Version with issue:
- Last known good version:
- IDE: VS2019
- Platform Target Frameworks: 
  - iOS:  14.2 
  - Android: 11.0 (N/A)
- Nuget Packages: 
Xamarin.Essentials. 1.6.0-pre2
Xamarin.Forms 4.8.0.1687

- Affected Devices:
Simulator and Physical, both iPhone 12 Pro Max (iOS 14.2)
### Screenshots
![image](https://user-images.githubusercontent.com/32236075/102375649-ab93f700-3fba-11eb-8dd8-e5fc2b161ded.png)

dimonovdd:
I also encountered this problem

> File is .xlsx format.

this applies to files with different extensions

dimonovdd:
this correctly work on iPhone 7 Plus with iOS 13.5

dimonovdd:
@mattleibow I noticed that it works on ios 14, but the opening "save to files" dialog takes a very long time. But the second and next times the opening is fast.
it works faster on ios 13.

**Videos**:
1. [iOS 13.6](https://yadi.sk/i/gbH71xc__SV9UQ)
2. [ iOS 14.3](https://yadi.sk/i/V46BhVQbIeW94A)

mattleibow:
Thanks for the PR.

PR #1609, opened 2020-12-27
[Launcher][iOS] Fixing save to files bug

dimonovdd:
### Bugs Fixed ###

- Related to issue #1592

### PR Checklist ###


Samples/Samples/View/LauncherPage.xaml
```diff
@@ -22,7 +22,7 @@
                     <ColumnDefinition Width="*" />
                     <ColumnDefinition Width="*" />
                 </Grid.ColumnDefinitions>
-                <Entry Grid.Row="0" Grid.ColumnSpan="2" Text="{Binding LaunchUri}" />
+                <Entry Grid.Row="0" Grid.ColumnSpan="2" Text="{Binding LaunchUri}" TextColor="Black" />
```
mattleibow, line 25: These should no longer be needed. I just added a new style to fix this for iOS:
https://github.com/xamarin/Essentials/pull/1610

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -30,47 +31,27 @@ static Task<bool> PlatformTryOpenAsync(Uri uri)
         }
 
 #if __IOS__
+
```
mattleibow, line 34:
```suggestion
```

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -30,47 +31,27 @@ static Task<bool> PlatformTryOpenAsync(Uri uri)
...
         static UIDocumentInteractionController documentController;
 
         static Task PlatformOpenAsync(OpenFileRequest request)
         {
-            var fileUrl = NSUrl.FromFilename(request.File.FullPath);
-
-            documentController = UIDocumentInteractionController.FromUrl(fileUrl);
-            documentController.Delegate = new DocumentControllerDelegate
+            documentController = new UIDocumentInteractionController()
```
mattleibow, line 39: What is the difference between the `ctor()` or `UIDocumentInteractionController.FromUrl`? Is there a reason?

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -30,47 +31,27 @@ static Task<bool> PlatformTryOpenAsync(Uri uri)
         }
 
 #if __IOS__
+
         static UIDocumentInteractionController documentController;
```
mattleibow, line 35: Will this leak? Do we get an event when a save is actually made? If not, then maybe add a comment saying this is here to prevent the GC from clearing up but should not be referenced outside the method.

I'll also do some tests to see if we actually need this field is the simple act of having the picker visible prevents the GC. Based on our chats in Discord, it seems it is fine, but I'll still test on my devices.

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -30,47 +31,27 @@ static Task<bool> PlatformTryOpenAsync(Uri uri)
...
-            else
-            {
-                rect = vc.View.Bounds;
-            }
-
-            documentController.PresentOpenInMenu(rect.Value, vc.View, true);
+            var view = Platform.GetCurrentUIViewController().View;
+            var rect = DeviceInfo.Idiom == DeviceIdiom.Tablet
+                ? new CGRect(new CGPoint(view.Bounds.Width / 2, view.Bounds.Height), CGRect.Empty.Size)
+                : view.Bounds;
```
mattleibow, line 49: Not for this PR, but maybe we should think of adding a `PresentationSourceBounds` property to file request so that the user can potentially control this. Could you open an issue for that?

mattleibow (commented):
This looks good. Just a few Q's and then a quick revert of one file.

After I do some testing in release on a device, we can approve.

Samples/Samples/View/LauncherPage.xaml
dimonovdd: We can roll back, I just didn't see the text I'm typing

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
dimonovdd: https://github.com/xamarin/Essentials/issues/1592#issuecomment-751392147

I think it works a little bit faster

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
dimonovdd: yes, this should be tested. but I haven't found a better option yet

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
dimonovdd: Ок

dimonovdd: https://github.com/xamarin/Essentials/issues/1611

Samples/Samples/View/LauncherPage.xaml
mattleibow: Yeah, thanks. We can fix this in a new PR so we don't have too much in this one.

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
mattleibow: Awesome, thanks.

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
mattleibow: I saw the delay on the sim, but only on the first run. When I clicked it again, it was fast. But both work so it is ok.

mattleibow: actually, i noticed the delay on my device too... seems like a bug in iOS as we have already sent things over. It can't be our file as we just create on with bytes. So, i think the OS has to load all the providers.

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
mattleibow: I think this is fine for now as we don't have open streams or file handles. it is just strings and uris.

mattleibow (approved):

PR #1610, opened 2020-12-28
Update some libraries in the samples

mattleibow:
### Description of Change ###

Update libraries and fix the issue where iOS text boxes have invisible text.


Samples/Samples/App.xaml
```diff
@@ -10,6 +10,21 @@
             <Style TargetType="Page" ApplyToDerivedTypes="True">
                 <Setter Property="Visual" Value="{x:Static local:App.PreferredVisual}" />
             </Style>
+            <Style TargetType="Editor">
```
mattleibow, line 13: For some reason, the default text on iOS with material is invisible in text boxes. So we just make it black for this sample. Probably a real reason and might be a bad style. But, for this sample app, no need to spend much time on it. :)

PR #1614, opened 2021-01-03
[Enhancement][HapticFeedback] Prepare generator

dimonovdd:
### Description of Change ###

This revision is necessary for applications that often use HapticFeedback. This changes will improve performance on iOS, UWP and Tizen. Using the new method, we will not create a new object for HapticFeedback every time.
It doesn't break the old API.

### API Changes ###
Added: 
 
 ```csharp
public static class HapticFeedback
{
    public static void Perform(HapticFeedbackType type = HapticFeedbackType.Click)
    {
        using var generator = PrepareGenerator(type);
        generator.Perform();
    }
   
    //new method
    public static HapticFeedbackGenerator PrepareGenerator(HapticFeedbackType type = HapticFeedbackType.Click);
}
```

 ```csharp
public class HapticFeedbackGenerator : IDisposable
{
    public void Perform();
    public void Dispose();
}
```

### PR Checklist ###

- [x] Updated documentation


Xamarin.Essentials/HapticFeedback/HapticFeedback.shared.cs
```diff
@@ -5,10 +5,45 @@ namespace Xamarin.Essentials
     public static partial class HapticFeedback
     {
         public static void Perform(HapticFeedbackType type = HapticFeedbackType.Click)
+            => PlatformPerform(type);
+
+        public static HapticFeedbackGenerator PrepareGenerator(HapticFeedbackType type = HapticFeedbackType.Click)
+#if MONOANDROID || WINDOWS_UWP || __MACOS__ || TIZEN
```
mattleibow, line 11: If possible, can we move this to platform specific files? This will keep the shared code files clean and also forms a "interface" of sorts. You can create a a HapticFeedback.android.uwp.macos.tizen.cs just or these to reduce code and files.

mattleibow:
Maybe I am not understanding this, but what is the benefit/purpose of this split logic. I get for iOS that is how it works, but how does this split benefit over calling static haptic methods twice or more?

dimonovdd:
@mattleibow 
I have a case in which I very often perform Haptic Feedback. 
and each time there are 4 actions with `UIImpactFeedbackGenerator:`
1) initialize
2) Prepare
3) ImpactOccurred
4) Dispose

probably the new method is more correct, and we should have made it right away and abandoned `static HapticFeedback.Perform()` method

Xamarin.Essentials/HapticFeedback/HapticFeedback.shared.cs
dimonovdd: Yes, well, I'll split this into specific platforms files

jamesmontemagno (dismissed):
Looks fine to me. API remains in tact.

Issue #1548, opened 2020-11-25
[Enhancement] SaveToAlbum -> MediaPicker.CapturePhotoAsync


noxe:
Hi,

will SaveToAlbum be added to the MediaPickerOptions - currently we use the "old" CrossMedia Plugin where we used 

var options = new MediaPickerOptions()
                {
                    SaveMetaData = true,
                    CompressionQuality = 92,
                    SaveToAlbum = true
                };

which is not available in Essentials?

thank you!

dimonovdd:
@noxe read [this issue](https://github.com/xamarin/Essentials/issues/1455). 
I would like to do pull request. but I don't know where it should be implemented yet

noxe:
ohh - thx for the update!

atexinspect:
CompressionQuality & CustomPhotoSize would also be a great from the old media picker.

Issue #1455, opened 2020-10-17
[Enhancement] Save to gallery!


dimonovdd:
## Summary
Saving to the gallery is a very useful feature for saving media files

## API Changes
I have a [draft for iOS and Android](https://gist.github.com/dimonovdd/07ecb6ceb14bf72c7e54dc0210e4d0c0), but I do not know in which the Essentials API class should implement this

mattleibow:
Ah... I think maybe we named the media picker wrong... Maybe should have been media gallery or something.

What type name sounds correct? I think the past can be fixed, but we just need to think of a better name, or a new api. But I think a new api might not be good as it separates the haller access into bits.

dimonovdd:
> Maybe should have been media gallery

I think it's a good name

dimonovdd:
@mattleibow  Can I rename the `MediaPicker` class and everything referenced with it to `MediaGallery`?

jamesmontemagno:
There are different things...

Like MediaPicker let's you pick or take photos and videos. It could be from the gallery or may not be.

I prefer almost 2 classes.... like "Camera" and "MediaPicker" and picker can TakePhotoAsync or TakeVideoAsync and MediaPicker lets you pick a photo/video.

The Camera could have a feature to save it to the gallery on the device. However, my recommendation is don't as it is extremely complicated and metadata and all this stuff. I have done it, it is terrible.

dimonovdd:
I just need the ability to save images to the gallery. So I wanted to do it. 
@jamesmontemagno let me know if there's anything I can do to help

jamesmontemagno:
I mostly feel like we shouldn't introduce it into the core as it is more of a feature that is very pick per device (especially on android) and will cause a lot of issues. I think it is more of a document how to do it somewhere or write a good blog post.

dimonovdd:
I don't see anything special about this feature. Why should Xamarin developers add an image saving implementation to their projects? We can do it well, and everyone will take advantage of this feature

The more features out of the box, the bigger our community will be

jamesmontemagno:
I agree, I just know that from my MediaPlugin that the API to save to gallery doesnt' work on all platforms and on Android isn't consistent across devices. Consistency here is important.

dimonovdd:
I don't think it's easy. But it works on Android in many other(Telegram, Whatsapp, etc.)
But what API class can we add this to?

PR #1628, opened 2021-01-09
[Enhancement] Save To Gallery

dimonovdd:
### Description of Change ###

implementation to saving media files with metadata on iOS, MacOS, Android and UWP.

### Related to issue ###

-  #1455 
-  #1548 
-  #1668

### API Changes ###
```csharp
public static class MediaGallery
{
    public static Task SaveAsync(MediaFileType type, string filePath);
    public static Task SaveAsync(MediaFileType type, byte[] data, string fileName);
    public static Task SaveAsync(MediaFileType type, Stream fileStream, string fileName);
}
```
```csharp
public enum MediaFileType
{
     Image,
     Video
}
```

### PR Checklist ###

- [x] Has tests 
- [x] Has samples 
- [x] Updated documentation


Xamarin.Essentials/SaveToGallery/SaveToGallery.ios.cs
```diff
@@ -0,0 +1,17 @@
...
+using UIKit;
+
+namespace Xamarin.Essentials.SaveToGallery
+{
+    public static partial class SaveToGallery
+    {
+        static Task PlatformSaveImageAsync(byte[] data, string filename, string albumName)
+        {
+            var image = new UIImage(NSData.FromArray(data));
+            image.SaveToPhotosAlbum((UIImage img, NSError error) => { });
```
IngweLand, line 13: `UIImage.SaveToPhotosAlbum` does not save metadata

dimonovdd: yes I know, I'm trying to find the right cross-platform solution

IngweLand: Might be a better option (that's for iOS). 
```
var library = new ALAssetsLibrary();
var albumSave = await library.WriteImageToSavedPhotosAlbumAsync(cgImage, meta);
var albumSave = await library.WriteVideoToSavedPhotosAlbumAsync(new NSUrl(path));
```

dimonovdd:
Hi @mattleibow. I finished a draft implementation of saving video and image files with metadata on ios.
But as you remember, I don't have a real iphone. Could you check it on real devices please?

dimonovdd:
@jamesmontemagno 
You wrote about the complexity of the implementation in [the comments on this issue](https://github.com/xamarin/Essentials/issues/1455#issuecomment-714762221). Could you check this pull request, please?
This is still a draft, but I tested it as best I could

Xamarin.Essentials/SaveToGallery/SaveToGallery.ios.macos.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        static async Task PlatformSaveAsync(MediaFileType type, string filePath, string albumName)
+        {
+            if (!File.Exists(filePath))
+                throw new ArgumentNullException(nameof(filePath));
+
+            await Permissions.EnsureGrantedAsync<Permissions.Photos>();
+
+            using var album = GetAlbum(albumName) ?? await CreateAlbumAsync(albumName);
```
follesoe, line 56: This part will fail on iOS 14 if limited access to the Media Library has been granted. See https://github.com/xamarin/Essentials/issues/1635 for details.

Basically, iOS 14 introduces a new access level to the gallery, allowing you to read selected photos and save new photos, but not create or retrieve named albums.

Using `await Permissions.EnsureGrantedAsync<Permissions.Photos>();` will not be enough, as it will pass but the code will later crash with a runtime exception missing permission to read or create the album.

follesoe:
Great to see work on Essentials API to store media in the gallery! We have this exact need in our app, and as some of the APIs are obsolete in Android 11 I am about to start working on that area of the app again. I will try to copy in the work-in-progress version of this API and test it in practice in our app, and report back any findings. For now, the first thing that needs to be addressed is the new permission level for Photos in iOS 14 (on the iOS side).

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        static readonly DateTime jan1st1970 = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        // JavaSystem.CurrentTimeMillis()
+
+        static long TimeMillis(DateTime current)
+            => (long)CalcTimeDifference(current).TotalMilliseconds;
+
+        static long TimeSeconds(DateTime current)
+            => (long)CalcTimeDifference(current).TotalMilliseconds;
```
follesoe, line 108: Guess this one should be `TotalSeconds`.

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            var relativePath = type == MediaFileType.Image
+                ? Environment.DirectoryPictures
+                : Environment.DirectoryMovies;
+
+            if (Platform.HasApiLevelQ)
+            {
+#if MONOANDROID10_0
+                values.Put(MediaStore.MediaColumns.RelativePath, Path.Combine(relativePath, albumName));
+                values.Put(MediaStore.MediaColumns.DateTaken, TimeMillis(dateTimeNow));
```
follesoe, line 64: Would be misleading to use explicitly set the `DateTaken` to be equal to the import time (`DateAdded´).

My current scenario is downloading and importing photos taken by an external system (drone), which sets the date taken data as EXIF columns. At least this should be documented. It's fair that the Essentials should be the "base path", and in many cases, this is probably fine...

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
dimonovdd: oh, yes

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
dimonovdd: We can try to remove this. The metadata should remain unchanged

dimonovdd:
Hi @sung-su 
I made a draft implementation for several platforms. Is it possible to do this for Tizen?

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
dimonovdd: @follesoe, can you give me a sample file with a metadata that the drone made?

Xamarin.Essentials/SaveToGallery/SaveToGallery.ios.macos.cs
dimonovdd: Yes, I know about the problem with iOS. I don't have a good idea for this yet

Xamarin.Essentials/SaveToGallery/SaveToGallery.android.cs
dimonovdd: I checked, this does not affect the date of capture a photo, which is in the metadata

|__________________________| Sample Video |__________________________|
|:---------------------------:|:-------------:|:---------------------------:|
||Video||

follesoe: Will find a picture for you @dimonovdd, but looking at the documentation: https://developer.android.com/reference/android/provider/MediaStore.MediaColumns#DATE_TAKEN

> Indexed value of MediaMetadataRetriever#METADATA_KEY_DATE or ExifInterface#TAG_DATETIME_ORIGINAL extracted from this media item.

So I don't really see why it would make sense to set this for `DateTime.Now`, from my understanding it should not be set and then Android could index it from the EXIF metadata.

However, during our testing we did find cases where this field was not set correctly on all Android devices, so we explicitly set this field to a value manually read from the EXIF information of our file.

This is how we do it:

```csharp
var directories = ImageMetadataReader.ReadMetadata(fileStream);
var exif = directories.OfType<ExifSubIfdDirectory>().FirstOrDefault();
var dateTaken = exif?.GetDateTime(ExifDirectoryBase.TagDateTimeOriginal);

if (dateTaken.HasValue)
{
    values.Put(MediaStore.IMediaColumns.DateTaken, dateTaken.HasValue ? dateTaken.Value.ToUnixTimeMs() : dateTimeNow.ToUnixTimeMs());
}

// Set position to 0 after reading Exif.
if (fileStream.CanSeek)
{
    fileStream.Position = 0;
}
```

The EXIF is extracted using the `MetadataExtractor` NuGet package.

dimonovdd: Metadata remains the same. Perhaps not all devices display the date when a photo was captured which is in a metadata.
I definitely don't see the point in adding extra dependencies. 
Android documentation is good, but device manufacturers often don't follow it. That's why the documentation is useless

follesoe: @dimonovdd Sure, I am not arguing taking this dependency or reading the EXIF, but I am arguing that it makes no sense trying to set the DATE_TAKEN field at all, as the picture most definitely was NOT taken at `DateTime.Now` as the current implementation does, and Android should extract this information from the EXIF metadata so makes no sense trying to explicitly set it to a value by default.

jamesmontemagno:
Are all of these APIs available on iOS 10+ and Android 5+?

dimonovdd:
> Are all of these APIs available on iOS 10+ and Android 5+?

Yes. I tested this on android 5.0. Apple [documentation](https://developer.apple.com/documentation/photokit/phassetchangerequest/1624057-creationrequestforassetfromvideo?language=objc) says this should work on iOS 8.0. But I only tested this on version 12, 13, and 14.
[There are tests for this feature](https://github.com/xamarin/Essentials/blob/5d9592332bfdb54ed9a1dcc82241a480f7e30445/DeviceTests/DeviceTests.Shared/MediaGallery_Tests.cs#L9).

jamesmontemagno (dismissed):

jfversluis:
We won't be doing this for Xamarin.Essentials anymore. Thanks for the time and effort here!

Issue #1638, opened 2021-01-17
[Bug] [iOS] Exif and other metadata is missing when capturing photos with MediaPicker


IngweLand:
### Description
There is no EXIF and some other metadata included into file when capturing photos with MediaPicker

### Steps to Reproduce

1. Capture photo
2. Check metadata

### Expected Behavior

Metadata should be present

### Actual Behavior

Metadata is absent

### Basic Information

- Version with issue: 1.6.0
- Last known good version: none
- IDE: VS 16.8.4
- Platform Target Frameworks: 
  - iOS:  14.3
  
- Affected Devices: all

jfversluis:
Thanks for the report! More on this and the background can be found in this issue #1514

TL;DR We won't be able to add this to Xamarin.Essentials anymore, you can either use the MediaGallery plugin or consider voting for adding it to the .NET MAUI Community Toolkit for the time being. Thanks!

Issue #1631, opened 2021-01-13
MediaPicker Take Photo iOS incorrect FullPath[Bug] 


oleksandrtretiak:
FullPath looks like the same as FileName.
### Screenshots

<img width="494" alt="Screenshot 2021-01-13 at 14 19 50" src="https://user-images.githubusercontent.com/6115977/104460176-6a413800-55b6-11eb-9a5f-e81899d7e464.png">

IngweLand:
I do not think this is a bug. As far as I understand, the physical file is not created when the photo is captured. Not sure why they extend the result from `FileResult` in that case, probably for convenience. As such, they had to include that FullPath.

        internal UIImageFileResult(UIImage image)
            : base()
        {
            uiImage = image;


            FullPath = Guid.NewGuid().ToString() + FileSystem.Extensions.Png;
            FileName = FullPath;
        }

olitee:
I find this inconsistency between iOS and Android odd: 

iOS does not create a file - but Android does. So the handling and clean-up logic I write ends up being platform specific. Unless I've missed something?

IngweLand:
My pull request fixes that issue

mattleibow:
The file path returned but the pickers are all dependent on where the file comes from. Some files are from a library, so making a local copy is not always done. For example, picking files from the iOS photos library will return the original. Android does something similar if you have the permissions to the SD card and the photo was on there.

Instead of making a copy, you can just open the file and read it. If you want a copy, then just copy the stream using `OpenReadAsync`. The path is just where the file was picked from. If the OS does not want to give this out, then we generate a guid so at least the file names are unique.

kunalenterprise:
@mattleibow What happens in case when using PickPhotoAsync(). I am getting guid in filename and path.

PR #1639, opened 2021-01-17
[Draft] [MediaPicker] [iOS] Captured file metadata

IngweLand:
### Description of Change ###

1. Preserves metadata of captured image
2. Saves the image into app specific temp location
3. MediaPicker sample updated to show metadata

### Questions ###

1. I think we should add some custom exception for the cases when MediaPicker's capturing fails. Not sure where though.
2. Should we rename UIImageFileResult into ImageFileResult or something? After all, UIImage is used just for input there
3. Should we include new options into this PR (e.g., if image should or should not preserve metadata)? If yes, I feel like current `MediaPickerOptions` are not a best choice because there might be different options for picking and capturing media. 

### Bugs Fixed ###

- Related to issue #1638, #1631, #1514

### API Changes ###

Internal API changed

New:

- `FileResultCreationException`

Changed:

 - `internal UIImageFileResult(UIImage image)` => `internal UIImageFileResult(UIImage image, NSDictionary metadata)`
 

### Behavioral Changes ###

Images are saved as JPEG now instead of PNG

### PR Checklist ###

### DIFF
--- a/Samples/Samples/View/MediaPickerPage.xaml
+++ b/Samples/Samples/View/MediaPickerPage.xaml
@@ -1,10 +1,9 @@
-<views:BasePage
-    x:Class="Samples.View.MediaPickerPage"
-    xmlns="http://xamarin.com/schemas/2014/forms"
-    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-    xmlns:viewmodels="clr-namespace:Samples.ViewModel"
-    xmlns:views="clr-namespace:Samples.View"
-    Title="Media Picker">
+<views:BasePage xmlns="http://xamarin.com/schemas/2014/forms" 
+                xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
+                xmlns:viewmodels="clr-namespace:Samples.ViewModel" 
+                xmlns:views="clr-namespace:Samples.View" 
+                x:Class="Samples.View.MediaPickerPage" 
+                Title="Media Picker">
     <views:BasePage.BindingContext>
         <viewmodels:MediaPickerViewModel />
     </views:BasePage.BindingContext>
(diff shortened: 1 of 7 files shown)


Xamarin.Essentials/FileSystem/FileSystem.ios.cs
```diff
@@ -176,25 +179,62 @@ internal override Task<Stream> PlatformOpenReadAsync()
...
+            Stream fileStream = File.OpenRead(FullPath);
+
+            return Task.FromResult(fileStream);
+        }
+
+        void SaveImageWithMetadata(UIImage image, NSDictionary metadata)
+        {
+            try
+            {
+                var imageData = image.AsJPEG();
```
mattleibow, line 205: The down side of JPEG is that is compressed and loses transparency. For example, I might pick a png from my downloads folder.

Or maybe this is not the case that it will go down this path?

Samples/Samples/View/MediaPickerPage.xaml
```diff
@@ -1,23 +1,72 @@
-﻿<views:BasePage xmlns="http://xamarin.com/schemas/2014/forms"
-                xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-                xmlns:views="clr-namespace:Samples.View"
-                xmlns:viewmodels="clr-namespace:Samples.ViewModel"
-                x:Class="Samples.View.MediaPickerPage"
-                Title="Media Picker">
+<views:BasePage
```
mattleibow, line 1: If possible, can we preserve the XAML formatting for consistency?

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
mattleibow: Is it possible to maybe detect the source image and use that? Can a iOS save PNG data?

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
```diff
@@ -176,25 +179,62 @@ internal override Task<Stream> PlatformOpenReadAsync()
...
             : base()
         {
-            uiImage = image;
-
-            FullPath = Guid.NewGuid().ToString() + FileSystem.Extensions.Png;
-            FileName = FullPath;
+            FileName = Guid.NewGuid() + FileSystem.Extensions.Jpg;
+            var tempDir = NSFileManager.DefaultManager.GetTemporaryDirectory().Path;
+            FullPath = Path.Combine(tempDir, FileName);
+            SaveImageWithMetadata(image, metadata);
```
mattleibow, line 191: Since this is now a more complex ctor that does a fair bit of work, maybe this should really be transformed as your comment suggests. Rename this to `ImageFileResult` and then use a ctor that takes a path. Then maybe add a static method such as `Create` and give an indication that work is being done. Or since this is now not doing anything, just move the logic to a method on FileSystem and use that to get a path to create a plain old FileResult?

mattleibow: In fact, it might just be work reverting this class and keeping it that simple logic, and then moving your code to FileSystem. We might not use this class, but we can remove it later as it still might have a bit of value now for other cases.

In the actual code, just call the new method to get a string path, and return the new FileResult(path). This is what we do for Android anyway.

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
```diff
@@ -176,25 +179,62 @@ internal override Task<Stream> PlatformOpenReadAsync()
...
+                using var ciImage = CIImage.FromData(imageData);
+                var originalColorSpace = ciImage.ColorSpace;
+                using var newImageSource = ciImage.CreateBySettingProperties(metadata);
+                using var ciContext = new CIContext();
+                var result = ciContext.WriteJpegRepresentation(newImageSource, NSUrl.FromFilename(FullPath), originalColorSpace ?? CGColorSpace.CreateSrgb(), new NSDictionary(), out var error);
+                if (result)
+                {
+                    if (error != null)
+                    {
+                        // TODO: Throw custom exception and include the error
```
mattleibow, line 222: You can throw an NSErrorException with the error.

mattleibow: The logic may cause this to never be reached:

> Return Value
> If true, file export succeeded. If false, examine the errorPtr parameter for possible failure reasons.

https://developer.apple.com/documentation/coreimage/cicontext/1642218-writejpegrepresentation

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
```diff
@@ -176,25 +179,62 @@ internal override Task<Stream> PlatformOpenReadAsync()
...
+                    }
+
+                    // TODO: Throw custom exception
+                }
+            }
+            catch (Exception ex)
```
mattleibow, line 228: No need for the catch here as it is fine to have the NSErrorException go out as it will contain all the info needed to debug.

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
dimonovdd: > The down side of JPEG is that is compressed and loses transparency

Yes, that's where the difficulty lies, I myself can not figure out how to do it correctly to keep images and photos original with all the data without compression and loss of transparency.

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
IngweLand: > The logic may cause this to never be reached:
> 
> > Return Value
> > If true, file export succeeded. If false, examine the errorPtr parameter for possible failure reasons.
> 
> https://developer.apple.com/documentation/coreimage/cicontext/1642218-writejpegrepresentation

True. My mistake, that second prospect exception should be in `else` clause

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
IngweLand: That's the whole point. I think we should better create our custom exception, so the user will know what to expect. We should add internal exceptions if the user wants to analyze exact reason.

IngweLand: There is another cases, where there is no internal exception, like above in that code, where the image data might be null. Which, I believe, is a good place for custom exception as well. Something like, ImageCreationException

Xamarin.Essentials/FileSystem/FileSystem.ios.cs
IngweLand: > Or since this is now not doing anything, just move the logic to a method on FileSystem and use that to get a path to create a plain old FileResult?

This Class is for creating files from camera. There is no path at the moment there (I think camera creates just UIImage when taking images, unlike with videos, when it creates the file in the temp dir)

IngweLand: > In fact, it might just be work reverting this class and keeping it that simple logic, and then moving your code to FileSystem. We might not use this class, but we can remove it later as it still might have a bit of value now for other cases.
> 
> In the actual code, just call the new method to get a string path, and return the new FileResult(path). This is what we do for Android anyway.

I think about this new implementation as a new replacement of current way of saving images. Current implementation does not really do what it should. Check the list if bugs which this PR suppose to fix.

IngweLand:
I think it would be a good idea to postpone merging this PR until specs are finalized for adding other APIs available in the original MediaPlugin. https://github.com/xamarin/Essentials/issues/1640
This way we might avoid introducing breaking changes later.

henrikweimenhog:
Just saying, PNGs also have EXIF so saving as JPG wont help.
Either add EXIF or at an absolute minimum rotate the image to counter the missing EXIF.

jfversluis:
Hey @IngweLand thank you so much for your time and effort on this. Unfortunately, at this time we won't be able to get this in Xamarin.Essentials anymore. For more background and discussion please see #1514

PR #1673, opened 2021-02-02
Fix Secure storage group not shareable between iOS applications

kiddailey:
### Description of Change ###

Resolves secure storage access groups not working on iOS by ensuring that specified shared access groups are not overridden with a bundle-specific access group.

### Bugs Fixed ###

Implementation of changes discussed in thread below, removing bundle-specific override of access group for iOS:

https://github.com/xamarin/Essentials/pull/1351#discussion_r485128322

### API Changes ###

Changed:

 - When access group is specified, SecureStorage explicitly uses shared access group provided.
 - SecureStorage backwards compatibility if no access group supplied, ie. still uses old XaraminEssentials alias as documented

### DIFF
--- a/Xamarin.Essentials/SecureStorage/SecureStorage.shared.cs
+++ b/Xamarin.Essentials/SecureStorage/SecureStorage.shared.cs
@@ -10,7 +10,7 @@ namespace Xamarin.Essentials
 
         // Special alias needed to used grouping
         internal static string GetAlias(string accessGroup)
-           => string.IsNullOrWhiteSpace(accessGroup) ? Alias : $"{accessGroup}.{Alias}";
+           => string.IsNullOrWhiteSpace(accessGroup) ? Alias : accessGroup;
 
         public static Task<string> GetAsync(string key, string accessGroup)
         {
--- a/Xamarin.Essentials/SecureStorage/SecureStorage.ios.tvos.watchos.macos.cs
+++ b/Xamarin.Essentials/SecureStorage/SecureStorage.ios.tvos.watchos.macos.cs
@@ -17,7 +17,7 @@ namespace Xamarin.Essentials
 
         public static Task SetAsync(string key, string value, SecAccessible accessible, string accessGroup)
         {
-            var alias = GetAlias(accessGroup);
+            var alias = string.IsNullOrWhiteSpace(accessGroup) ? Alias : accessGroup;
             if (string.IsNullOrWhiteSpace(key))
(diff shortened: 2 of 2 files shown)


ali-moussa:
Can we get this reviewed please? Its quite the blocker.

Cheers

Xamarin.Essentials/SecureStorage/SecureStorage.shared.cs
```diff
@@ -10,7 +10,7 @@ public static partial class SecureStorage
 
         // Special alias needed to used grouping
         internal static string GetAlias(string accessGroup)
-           => string.IsNullOrWhiteSpace(accessGroup) ? Alias : $"{accessGroup}.{Alias}";
+           => string.IsNullOrWhiteSpace(accessGroup) ? Alias : accessGroup;
```
jamesmontemagno, line 13: This one concerns me as this is shared so would break all other platforms. We should not change this.

jamesmontemagno (changes requested):

jamesmontemagno:
This is a big oversight in the library for sure. I left some comments as we need to not break other platforms.

Xamarin.Essentials/SecureStorage/SecureStorage.shared.cs
kiddailey: Thanks for reviewing!  Without this on iOS, shared, secure access won't work, so it sounds like this aliasing mechanism will have to be a bit more platform specific.  Will try to come up with something different and resubmit as soon as I can.  Thanks again!

jamesmontemagno: I mean in that case I would just change your logic not to call this method and just call this locally in each method:

```
string.IsNullOrWhiteSpace(accessGroup) ? Alias : accessGroup;
```

jamesmontemagno: checkout my commit

jamesmontemagno:
Ah i now see that this is a PR for my changes that i had in there a while ago. So not a blocker as it isn't in secure storage yet :)

So actually your previous is fine with changing the alias.

Xamarin.Essentials/SecureStorage/SecureStorage.shared.cs
jamesmontemagno: nm this is good, saw other comments

Issue #409, opened 2018-07-25
Please support Background task!


nguyenthanhliemfc:
Thank you for Essentials. It's help me a lot in development app
It will be nice if support Background Task!
Thank you!

> VS bug [#735660](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/735660)

jamesmontemagno:
Can you describe more as to what you mean here.

nguyenthanhliemfc:
Sorry for my bad English.
I mean please support easy way to help developer implement Background Task in app. Thank!

jamesmontemagno:
Do you have an example of the APIs on each platform you are looking to have abstracted?

nguyenthanhliemfc:
Example I want to send a GET method to get news every 15 min without open app!

cfl777:
Andriod new component for this is the Android WorkManager api - doesn't seem like Xamarin supports this yet. But it would be great to have a scheduler for both platforms supported on Essentials

Mrnikbobjeff:
I strongly support this proposal. We needed this for several apps and always rolled our own implementation, which always added overhead until you got everything right. Scheduling actions is very platform specific, and there is a huge potential for miususe. I.E. iOS, if you take a look at the [docs](https://developer.apple.com/library/archive/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html) you can see that for apps that need to update from time to time as i guess @nguyenthanhliemfc has to do you can see that it states 

> Enabling this mode is not a guarantee that the system will give your app any time to perform background fetches.
The only apps that can permanently run in the background on ios are voip, location and audio.

Android is a lot more liberal with its Job scheduler ([docs](https://docs.microsoft.com/en-us/xamarin/android/platform/android-job-scheduler)). It can specify a deadline, indicate that the job is to be persisted across reboots etc. It is available from Android 5.0 onward, for 4.4 we would have to look at the [Alarm Service](https://developer.android.com/reference/android/app/AlarmManager).  This is also more permissive than iOS, thus we would still be limited by iOS.

Finally, UWP is also mightier that iOS in that regard ([docs](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/support-your-app-with-background-tasks)). You have certain restrictions such as the minimum interval allowed between internally triggered background events is 15 minutes, and that your app only has 30 seconds to process your tasks. 

[...]

PR #1674, opened 2021-02-03
Background tasks support

knocte:
### Description of Change ###

Introduces cross-platform (except Tizen/WatchOS) support for Background Tasks.

### Bugs Fixed ###

https://github.com/xamarin/Essentials/issues/409

### API Changes ###

Consumer of this API would do this:

```csharp
public class SomeBackgroundWork : IBackgroundTask
{
    public SomeBackgroundWork()
    {
    }

    public Task StartJob()
    {
        // YOUR CODE HERE

        return;
    }
}
```

Then call: `Background.StartBackgroundWork()`

### PR Checklist ###

- [ ] Has tests (please help us test it manually or tell us how to write tests about this, we think it's hard)
- [ ] Has samples (community please help us add them! or... **let's see first if maintainers agree with the API, and after that we will add samples**)
- [x] Rebased on top of `main` at time of PR (yes)
- [ ] Changes adhere to coding standard (hopefully?)
- [ ] Updated documentation (**let's see first if maintainers agree with the API, and after that we will add documentation**)

### DIFF
--- a/Xamarin.Essentials/Background/Background.android.cs
+++ b/Xamarin.Essentials/Background/Background.android.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Android.Content;
+
+namespace Xamarin.Essentials.Background
+{
+    public static partial class Background
+    {
+        static ContextWrapper context;
+
+        public static void Init(ContextWrapper ctx)
+        {
+            context = ctx;
+        }
+
+        internal static void PlatformStart()
+        {
+            if (context == null)
+                throw new InvalidOperationException("Background service is not initialized yet");
+            var intent = new Intent(context, typeof(BackgroundService));
+            context.StartService(intent);
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


jamesmontemagno:
While I super appreciate this PR, I think that anything with Backgrounds need serious design considerations. There is advanced complexities with scheduling and additional APIs that need to be brought on. I do think that for now it is better served to use Shiny - https://github.com/shinyorg/shiny/tree/master/src/Shiny.Core/Jobs

knocte:
> I do think that for now it is better served to use Shiny

Until any BackgroundTasks PR is merged in Essentials, sure, I agree.
In fact my team tried to use Shiny at first but we found some problems. That's why we designed what you see here as a very minimalistic first version (moving forward, more things could be added to this first version in a second version, such as Periodic tasks).

jfversluis:
Hey @knocte thanks so much for your time and effort here. As James already pointed out I think there is a lot to consider to implement something like this. We're definitely not going to do that for Xamarin.Essentials anymore. If you're still interested in this feel free to open an issue or discussion on the .NET MAUI repo to discuss this further and see what we can do there.

Thanks!

PR #1713, opened 2021-03-03
[Enhancement] Feature quick preview file

andrk0p:
### Description of Change ###

Added the ability to quickly view the file inside the iOS app.

### API Changes ###

Added: 
`public bool QuickLook { get; set; } = false;`
To
`OpenFileRequest`

### PR Checklist ###

- [ ] Updated documentation


Xamarin.Essentials/Launcher/Launcher.netstandard.watchos.cs
```diff
@@ -14,6 +14,9 @@ public static partial class Launcher
         static Task PlatformOpenAsync(OpenFileRequest request) =>
             throw ExceptionUtils.NotSupportedOrImplementedException;
 
+        static Task PlatformOpenAsync(OpenFileRequest request, bool openInApp) =>
```
jamesmontemagno, line 17: No need for another parameter, it should just be in the OpenFileRequest

jamesmontemagno (changes requested):
I would only add it to the OpenFileRequest. This simplifies the API  so we don't need a new one and can document that it is only an iOS feature.

Xamarin.Essentials/Launcher/Launcher.android.cs
```diff
@@ -57,6 +57,9 @@ static Task PlatformOpenAsync(OpenFileRequest request)
             return Task.CompletedTask;
         }
 
+        static Task PlatformOpenAsync(OpenFileRequest request, bool openInApp) =>
```
jamesmontemagno, line 60: I think we would just add this to "OpenFileReqauest", no need to chagne teh singautre at all

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -62,9 +65,60 @@ static Task PlatformOpenAsync(OpenFileRequest request)
             return Task.CompletedTask;
         }
 
+        static Task PlatformOpenAsync(OpenFileRequest request, bool openInApp)
```
dimonovdd, line 68: There is no need to create this method. See @jamesmontemagno  comment.

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -62,9 +65,60 @@ static Task PlatformOpenAsync(OpenFileRequest request)
...
+        {
+            if (openInApp)
+            {
+                UIApplication.SharedApplication?.KeyWindow?.RootViewController?.
```
dimonovdd, line 72: use [Platform.GetCurrentUIViewController()](https://github.com/xamarin/Essentials/blob/7218ab88f7fbe00ec3379bd54e6c0ce35ffc0c22/Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs#L79)

Xamarin.Essentials/Launcher/Launcher.ios.tvos.cs
```diff
@@ -62,9 +65,60 @@ static Task PlatformOpenAsync(OpenFileRequest request)
...
+
+        public override nint PreviewItemCount(QLPreviewController controller) => 1;
+    }
+
+    public class PreviewItem : QLPreviewItem
+    {
+        public PreviewItem(string filePath, string fileName)
+        {
+            ItemUrl = NSUrl.FromFilename(filePath);
+            ItemTitle = fileName;
```
dimonovdd, line 116: `fileName` will be used as a window titlel.  Maybe, we should be able to change it.
A file name doesn't always look good

dimonovdd (commented):
LGTM.
But I wrote a few comments. Most importantly we need to change the way we get `UIViewController`

andrk0p:
@jamesmontemagno @dimonovdd Thank you for your feedback. I made the changes

IngweLand:
I do not think that QuickLook should be added to the Essentials. It's a dedicated single-platform feature only, while Essentials is supposed to have cross-platform implementations. 
If you still decide to add it, I think you should also add following things:

- Checking if file is supported in QL (https://developer.apple.com/documentation/quicklook/qlpreviewcontroller/1617016-canpreviewitem)
- Disposing of QLPreviewController and its data source

jfversluis:
Hey everyone! Thanks for all the effort and input here! At this point in time we're aiming to not add any new functionality anymore as we're transitioning into .NET MAUI. Also, there seems to be some discussion about if this should be in Essentials because it's very specific to one platform.

If this is still of interest to you, please open a discussion on the .NET MAUI repo and if people seem to be in favor we might be able to port this code there. I'm closing if for here for now. Thank you Andrey for your time here!

PR #1731, opened 2021-03-17
Fixes #1729 check for ios 13 on max frame rates

### LINKED ISSUE #1729: [Bug] DisplayInfo.GetMainDisplayInfo() on iOS <10.3 and lower causes crash
### Description
When running iOS, lower than 10.3 and call DisplayInfo.GetMainDisplayInfo(), you'll get crash with exception Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInvalidArgumentException Reason: -[UIScreen maximumFramesPerSecond]: unrecognized selector sent to instance 0x11be36af0

Main issue is this commit in Xamarin.Essentials ver.1.6.1: https://github.com/xamarin/Essentials/commit/90a40a38a76024c35936f732f0828dccb993b9b7

It calls UIScreen.MainScreen.MaximumFramesPerSecond, that is available only in iOS 10.3+ 
https://developer.apple.com/documentation/uikit/uiscreen/2806814-maximumframespersecond

You should check iOS version to be at least 10.3 before calling UIScreen.MainScreen.MaximumFramesPerSecond.
### Steps to Reproduce

1. Run iOS 9-12
2. Call DisplayInfo.GetMainDisplayInfo()
3. Get crash

### Expected Behavior
No crash
### Actual Behavior
Crash
### Basic Information

- Version with issue: 1.6.1
- Last known good version: 1.6.0
- IDE:
- Platform Target Frameworks: 
  - iOS:  14.4
  - Android:  
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices: iOS devices with iOS <10.3

jamesmontemagno:
### Description of Change ###

Describe your changes here. 

### Bugs Fixed ###

- Related to issue #1729

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
```

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/DeviceDisplay/DeviceDisplay.ios.cs
+++ b/Xamarin.Essentials/DeviceDisplay/DeviceDisplay.ios.cs
@@ -24,7 +24,7 @@ namespace Xamarin.Essentials
                 density: scale,
                 orientation: CalculateOrientation(),
                 rotation: CalculateRotation(),
-                rate: UIScreen.MainScreen.MaximumFramesPerSecond);
+                rate: Platform.HasOSVersion(10, 3) ? UIScreen.MainScreen.MaximumFramesPerSecond : 0);
         }
 
         static void StartScreenMetricsListeners()


jamesmontemagno:
Ready @mattleibow

mattleibow (approved):
`0` is probably not the greatest value, but we do that for Android so better to be consistent where 0 == unknown.

Redth (approved):

Issue #1771, opened 2021-04-25
[Bug] PermissionException: Permission request must be invoked on the main thread (with Geolocation on UWP)


janusw:
### Description

A `Xamarin.Essentials.PermissionException` is thrown whenever `Geolocation.GetLocationAsync` is called outside of the main thread ("Permission request must be invoked on the main thread"). This happens even if I make sure that the permission has been granted beforehand.

AFAICS it is ok to make a location request outside of the main thread, only requesting the permission must be done on the main thread. The problem is that each call to `Geolocation.GetLocationAsync` seems to trigger a permission request, even if the permission has already been granted. This effectively makes it impossible to query the location outside of the main thread on UWP. On Android and iOS this works fine.

### Steps to Reproduce

1. Check out the branch `janusw:uwp_geoloc_bug_repro` (https://github.com/janusw/Essentials/commits/uwp_geoloc_bug_repro)
2. Run Samples.UWP.
3. Go to the Geolocation section.
4. Click the lower Refresh button.

### Expected Behavior

The current location info is shown.

### Actual Behavior

Exception thrown: 'Xamarin.Essentials.PermissionException'
Unable to detect Location. Exception: Permission request must be invoked on the main thread.

### Basic Information

- Version with issue: 1.6.1
- Last known good version: ?
- IDE: VS 2019
- Platform Target Frameworks: 
  - UWP:  18362 (Windows 10 v1903)

### Reproduction Link

https://github.com/janusw/Essentials/commits/uwp_geoloc_bug_repro

jfversluis:
See https://github.com/xamarin/Essentials/pull/1777#issuecomment-1845430702

PR #1777, opened 2021-04-26
Permissions: buffer location permission status on UWP

janusw:
### Description of Change ###

This fixes a problem with location permissions on UWP. The problem is caused by a combination of several points:
* Requesting the location permission on UWP can only be done on the main thread.
* Querying the location automatically checks the permissions.
* Checking the permission status on UWP does the same thing as requesting the permission, see https://github.com/xamarin/Essentials/blob/7218ab88f7fbe00ec3379bd54e6c0ce35ffc0c22/Xamarin.Essentials/Permissions/Permissions.uwp.cs#L40:L41

All of this together means that it is currently impossible to query the location outside of the main thread, because a PermissionException is triggered ("Permission request must be invoked on the main thread").

To fix it, I'm buffering the location permission status, so that it only needs to be checked/requested once (if granted), and after that we can just return the buffered value, so that querying the location outside of the main thread becomes possible.

### Bugs Fixed ###

- Related to issue #1771

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###


Xamarin.Essentials/Permissions/Permissions.uwp.cs
```diff
@@ -125,18 +125,24 @@ public override Task<PermissionStatus> CheckStatusAsync()
                 return RequestLocationPermissionAsync();
             }
 
+            static PermissionStatus status = PermissionStatus.Unknown;
+
             internal static async Task<PermissionStatus> RequestLocationPermissionAsync()
             {
+                if (status == PermissionStatus.Granted)
```
jamesmontemagno, line 132: what if the user changes the permission while running?

janusw: Huh, good point. On the one hand, the docu of [Geolocator.RequestAccessAsync](https://docs.microsoft.com/en-us/uwp/api/windows.devices.geolocation.geolocator.requestaccessasync) says:

> The RequestAccessAsync method prompts the user for permission to access their location. The user is only prompted once (per app). After the first time they grant or deny permission, this method no longer prompts for permission.

This is what led me to believe that buffering might be a good idea (since repeated calls will apparently have no effect?).

OTOH, there is the [Geolocator.StatusChanged](https://docs.microsoft.com/en-us/uwp/api/windows.devices.geolocation.geolocator.statuschanged) event. So, in order to catch changes to the permissions, we should probably react to that event, right?

jamesmontemagno: Since we haven't created that event for the other platforms our recommendation is to just query the API before you call a method that requires it. 

I do see your issue though that you state that you want to query on a background thread accordingly, but we check on the UI thread. This probably is a better all up change for .NET MAUI and we should strip away permission checks and make users do it ahead of time.

janusw: > Since we haven't created that event for the other platforms our recommendation is to just query the API before you call a method that requires it.

... but one could still use the `StatusChanged` event internally in the UWP parts only (to update the buffered permission), right? We don't need it on the other platforms. (I was not suggesting to make the status event part of the public XE Permissions API or anything like that.)

> I do see your issue though that you state that you want to query on a background thread accordingly, but we check on the UI thread. This probably is a better all up change for .NET MAUI and we should strip away permission checks and make users do it ahead of time.

I don't actually think that making sure the permissions are granted is a bad idea. If you strip away the check, then there is no guarantee any more that querying the location will work.

Another alternative might be to simply make sure that the permissions are checked from the main thread (via `MainThread.InvokeOnMainThreadAsync` or similar), instead of throwing an error if this is not the case ...

janusw: > > Since we haven't created that event for the other platforms our recommendation is to just query the API before you call a method that requires it.
> 
> ... but one could still use the `StatusChanged` event internally in the UWP parts only (to update the buffered permission), right?

I have figured out how to do that. It's rather straightforward. If anyone is interested, I'll be happy to share it.

However, I don't really think it is necessary. The current UWP behavior on the main branch is as follows: If one removes the location permissions in the system settings while the app is running, the next location request simply triggers an exception ("Unable to detect location. Exception: Access is denied." or similar). The permission check does not seem to request the permission from the user again.

Therefore my buffering of the permission status should not be a problem: It is only used to skip the permission request, which is skipped anyway upon a change of the permission status while running. The behavior on my branch is the same as on main in this respect.

janusw: > > ... but one could still use the `StatusChanged` event internally in the UWP parts only (to update the buffered permission), right?
> 
> I have figured out how to do that. It's rather straightforward. If anyone is interested, I'll be happy to share it.
> 
> However, I don't really think it is necessary.

Sorry, I think I was wrong here. For the case where one starts with a granted permissions, which is buffered upon the first request and then denied in the system settings, everything was working well indeed.

However, I missed to handle the inverted case, where one starts with a buffered `Denied `permission that is changed to `Granted` subsequentially. I fixed that in c097ddd91c1f980ac598cd70bf07db463da56730. And in fact this case also introduces the necessity to handle permission status changes during runtime, as remarked by @jamesmontemagno. This is added in 9e80679fb83e2244bb627d7b8674ea4821dc307b.

@jamesmontemagno I hope you are happy with these additions.

janusw:
Rebased to 1.7.0.

jfversluis:
I'm on the fence about still adding this one. It seems safe enough, on the other hand, it has been open for some time without too much traction from other people. Therefore it seems a potential risk of breaking people at this time while there are not a lot of people waiting for this change.

I guess the workaround here is to request the permission manually instead of relying on our code to do it.

Thank you for all the work and effort on this!

Issue #1489, opened 2020-11-02
on iOS 14 the "Allow Location" prompt is show very briefly, then it instantly goes away before I'm able to click "Use location while in App"


aherrick:
What I'm seeing is when requesting location, on iOS 14 the "Allow Location" prompt is show very briefly, then it instantly goes away before I'm able to click "Use location while in App" 

Is this issue related? Below is my simple code. Wasn't sure if its somehow related to this:

https://stackoverflow.com/questions/7888896/current-location-permission-dialog-disappears-too-quickly
https://github.com/xamarin/Essentials/issues/1066

```
     while (true)
                {
                    if (await Permissions.CheckStatusAsync<Permissions.LocationWhenInUse>() == PermissionStatus.Granted)
                       break;

                    await AppHelper.RequestAsync_Fixed<Permissions.LocationWhenInUse>();
                }

                var request = new GeolocationRequest(GeolocationAccuracy.Medium)
                {
                    Timeout = TimeSpan.FromSeconds(30)
                };

                var location = await Geolocation.GetLocationAsync(request);

```
        
Video showing the issue:

https://streamable.com/3o51ah

_Originally posted by @aherrick in https://github.com/xamarin/Essentials/issues/1390#issuecomment-718288385_

mattleibow:
This will probably be fixed once we merge https://github.com/xamarin/Essentials/pull/1487

Then you won't need to use this code as either.

aherrick:
@mattleibow awesome thanks for the response! once the new version is pushed i'll update and try it out. if its successful i'll close this puppy out. 👍

mattleibow:
You can test the fix in the nuget from the artifacts here:
https://dev.azure.com/xamarin/public/_build/results?buildId=30010&view=results

I'm going to just make sure that VS2017 is not broken with this, and then I'll merge and then the build will go out to the preview feed: https://aka.ms/xamarin-essentials-ci/index.json

aherrick:
@mattleibow hi! is this part of the beta 3 release that just dropped? if so i'm not seeing the issue fixed :(

aherrick:
when i keep this code in 

  while (true)
                {
                    if (await Permissions.CheckStatusAsync<Permissions.LocationWhenInUse>() == PermissionStatus.Granted)
                       break;

                    await AppHelper.RequestAsync_Fixed<Permissions.LocationWhenInUse>();
                }

[...]

PR #1815, opened 2021-05-25
set result after cleanup of locationmanager #1489

SiNeumann:
### Description of Change ###

On IOS if you like to request LocationAlways you have to request locationWhenInUse first. This means in an average implementation one would have 2 consecutive calls
await Permissions.RequestAsync<Permissions.LocationWhenInUse>();  
await Permissions.RequestAsync<Permissions.LocationAlways>();  
Both use the static LocationManager instance.
Since RequestAsync<Permissions.LocationWhenInUse>() sets the result of the task completition source before cleaning up its locationManager there is a race condition where RequestAsync<Permission.LocationAlways>() already runs and thus this cleanup disposes the locationManager used by the second call. This will not always be the case but in my test around 50% of the time. Workaround wait 3-10 seconds in between the calls

### Bugs Fixed ###

- Related to issue #1489 

### API Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/Permissions/Permissions.ios.tvos.watchos.cs
@@ -186,9 +186,9 @@ namespace Xamarin.Essentials
                         }
 
                         del.AuthorizationStatusChanged -= LocationAuthCallback;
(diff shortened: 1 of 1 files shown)


jamesmontemagno:
This looks fine to me.

jfversluis:
LGTM Thanks for your contribution to our little project @SiNeumann !

PR #1837, opened 2021-06-24
Implement prefersEphemeralWebBrowserSession Fixes #1712

### LINKED ISSUE #1712: WebAuthenticator iOS - Sign out does not delete the cookies and we are unable to sign in as a different user
### Description
Using WebAuthenticator for app sign in/out. Even after successful sign out, cookies aren't cleared and next time sign in logs in without prompting for user details.

### Expected Behavior
After sign out, when the user tries to sign in again, it should open the page which accepts user id (ex. Azure Active directory login page)

### Actual Behavior
App just signs in with the previous user. Unable to really switch users.

### Basic Information
Only workaround to achieve it is, 
Settings --> Safari --> Clear History and development data

###Note - Android seems to be working fine.

Tried below option, but no cookies are found
authContext.TokenCache.Clear();
var cookieStorage = Foundation.NSHttpCookieStorage.SharedStorage;
foreach (var cookie in cookieStorage.Cookies)
{
   cookieStorage.DeleteCookie(cookie);
}

Have gone through other similar issues in this repo like,
https://github.com/xamarin/Essentials/issues/1223
https://github.com/MicrosoftDocs/xamarin-docs/issues/2814

but unable to find a concrete solution.

Any suggestions/solutions would be of great help

Thanks

jamesmontemagno:
Fixes #1712

### DIFF
--- a/Xamarin.Essentials/AssemblyInfo/AssemblyInfo.ios.tvos.watchos.macos.cs
+++ b/Xamarin.Essentials/AssemblyInfo/AssemblyInfo.ios.tvos.watchos.macos.cs
@@ -1,7 +1,5 @@
 ﻿using Foundation;
 
-[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
-
 #pragma warning disable CS0618 // Type or member is obsolete
 [assembly: LinkerSafe]
 #pragma warning restore CS0618 // Type or member is obsolete
--- a/Xamarin.Essentials/AssemblyInfo/AssemblyInfo.ios.tvos.watchos.macos.cs
+++ b/Xamarin.Essentials/AssemblyInfo/AssemblyInfo.ios.tvos.watchos.macos.cs
@@ -1,3 +1,7 @@
 ﻿using Foundation;
 
+[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
+
+#pragma warning disable CS0618 // Type or member is obsolete
 [assembly: LinkerSafe]
+#pragma warning restore CS0618 // Type or member is obsolete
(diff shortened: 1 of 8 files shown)


dalexsoto (commented):
These look right to me but not my area of expertise @mandel-macaque could you also have a look? Thanks!

mattleibow (dismissed):

Xamarin.Essentials/WebAuthenticator/WebAuthenticator.ios.tvos.cs
```diff
@@ -82,6 +88,18 @@ static void AuthSessionCallback(NSUrl cbUrl, NSError error)
...
+            {
+                NSUrlCache.SharedCache.RemoveAllCachedResponses();
+                WKWebsiteDataStore.DefaultDataStore.HttpCookieStore.GetAllCookies((cookies) =>
+                {
+                    foreach (var cookie in cookies)
+                    {
+                        WKWebsiteDataStore.DefaultDataStore.HttpCookieStore.DeleteCookie(cookie, null);
+                    }
+                });
+            }
```
mandel-macaque, line 101: I do not believe that the fix is that easy. AFAIK the WKWebsiteDataStore.DefaultDataStore.HttpCookieStore returns a WKHttpCookieStore that will have a shred cookies for the process pool, this object does not allow to remove the cookies that have been set server side, at least the last time I checked, this means that all those cookies will be present as long as you are still using the same WKProcess. 

Have you tested this via a cookie set from the server side? I think that you will need to remove the cookies AND create a new WKProcess. 

Another way to do this is to use the ephemeral setting in the NSUrlSessionHanlder, which will not save the cookies and won't make you do this dance. Do you have a pointer in the code where you are making the call. The Ephemeral setting by the OS might be a cleaner way.

PS: Sorry, I don't know the code base enough, but I do know that if you do not change the WKProcess the cookies are there.

mandel-macaque (commented):
Looks good in the surface, I fear that server side set cookies are going to be a problem,

Xamarin.Essentials/AssemblyInfo/AssemblyInfo.ios.tvos.watchos.macos.cs
```diff
@@ -1,3 +1,7 @@
 ﻿using Foundation;
 
+[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
```
Redth, line 3: Let's introduce this in a different PR in the future.
```suggestion
```

Redth (approved):

PR #1846, opened 2021-07-12
Improving the operation of PresentationControllers

dimonovdd:
### Description of Change ###

Using [`Rg.Plugins.Popup`](https://github.com/rotorgames/Rg.Plugins.Popup) and `Essentials` together, I noticed some problems:
- Closing Popups closes pickers and etc., but the tasks remain not finished.
- We don't have the option to specify a UIViewController before opening pickers

This is due not only to `Rg.Plugins.Popup`, but also to any projects that use custom controllers for rendering pages. 

### API Changes ###
Added: 
 
```csharp
public static partial class Platform
{
    public static void Init(Func<UIViewController> getCurrentUIViewController)
}
```

### Behavioral Changes ###
- All tasks will be finished when the controller is closed
- Ability to change the parent controller for pickers

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
@@ -152,8 +152,11 @@ namespace Xamarin.Essentials
         {
             public Action DismissHandler { get; set; }
 
-            public override void DidDismiss(UIPresentationController presentationController) =>
+            public override void DidDismiss(UIPresentationController presentationController)
+            {
                 DismissHandler?.Invoke();
+                DismissHandler = null;
+            }
 
             protected override void Dispose(bool disposing)
             {
--- a/Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
+++ b/Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
@@ -73,12 +73,20 @@ namespace Xamarin.Essentials
 
 #if __IOS__ || __TVOS__
 
+        static Func<UIViewController> getCurrentController;
+
+        public static void Init(Func<UIViewController> getCurrentUIViewController)
+            => getCurrentController = getCurrentUIViewController;
+
         public static UIViewController GetCurrentUIViewController() =>
             GetCurrentViewController(false);
 
         internal static UIViewController GetCurrentViewController(bool throwIfNull = true)
         {
-            UIViewController viewController = null;
+            var viewController = getCurrentController?.Invoke();
+
+            if (viewController != null)
+                return viewController;
 
             var window = UIApplication.SharedApplication.KeyWindow;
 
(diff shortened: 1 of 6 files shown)


Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
```diff
@@ -138,5 +146,21 @@ internal static UIWindow GetCurrentWindow(bool throwIfNull = true)
...
+        internal class UIPresentationControllerDelegate : UIAdaptivePresentationControllerDelegate
+        {
+            public Action DismissHandler { get; set; }
+
+            public override void DidDismiss(UIPresentationController presentationController) =>
+                DismissHandler?.Invoke();
+
+            protected override void Dispose(bool disposing)
+            {
+                DismissHandler?.Invoke();
```
dimonovdd, line 160: Experimentally, I found out that `DidDismiss` method is not always called when closing parent controllers.  And this is probably the only workaround to finish a `TaskCompletionSource`.

mattleibow (approved):

Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
```diff
@@ -73,12 +73,20 @@ internal static string GetSystemLibraryProperty(string property)
 
 #if __IOS__ || __TVOS__
 
+        static Func<UIViewController> getCurrentController;
+
+        public static void Init(Func<UIViewController> getCurrentUIViewController)
+            => getCurrentController = getCurrentUIViewController;
```
mattleibow, line 79: @dimonovdd  What would be a use case for this? Just for reference.

dimonovdd: There was such a case in my application:
1) A popup appears
2) FilePicker appears
3) The popup closes and it closes the picker

because the picker was opened from the popup controller, and not from main navigation controller

dimonovdd:
Will it be merged? It is a big problem. It has been fixed in many libraries for example in [userdialogs](https://github.com/aritchie/userdialogs#ios-and-windows):
>Nothing is necessary any longer as of v4.x.  There is an Init function for iOS but it is OPTIONAL and only required if you want/need to control
the top level viewcontroller for things like iOS extensions.  Progress prompts will not use this factory function though!

mattleibow (dismissed):

Xamarin.Essentials/Platform/Platform.ios.tvos.watchos.cs
```diff
@@ -138,5 +146,21 @@ internal static UIWindow GetCurrentWindow(bool throwIfNull = true)
...
         internal static NSOperationQueue GetCurrentQueue() =>
             NSOperationQueue.CurrentQueue ?? new NSOperationQueue();
+
+#if __IOS__
+        internal class UIPresentationControllerDelegate : UIAdaptivePresentationControllerDelegate
+        {
+            public Action DismissHandler { get; set; }
+
+            public override void DidDismiss(UIPresentationController presentationController) =>
+                DismissHandler?.Invoke();
```
mattleibow, line 156: Can we set this to null ofter invoking? This way we don't get a double finish. Same with the Dispose invocation. I think we are fine now because all our usage is with a TrySetResult, so it makes no difference at the moment. But just in case there are other things in the future it is good to be safe.

mattleibow (approved):

PR #1900, opened 2021-10-11
Fix Share API does not correctly block when awaited

### LINKED ISSUE #1899: Awaited Calls to Share.RequestAsync do not block as expected
### Description
Share API is async, however it immediately returns despite not having completed the Share operation.
This is a problem, because it means there is no way to disable a button while the platform Share window is open, then re-enable it when the share operation has completed.
On iOS, if you request a UIActivityViewController be presented twice, the application may freeze.

### Steps to Reproduce

1. Call await Share.RequestAsync(...)

### Expected Behavior
Call should block until user either completes the sharing operation (i.e, picks a mode, print, dropbox, etc), or cancels the operation.

### Actual Behavior
The execution immediately proceeds.

### Basic Information

  - iOS:  iOS15

Adam-Langley:
### Description of Change ###

Fixes #1899 Share API does not correctly block when awaited. Should not return from async call until user has dismissed share UI.

### Bugs Fixed ###

- Related to issue #1899

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -6,22 +6,4 @@ labels: feature-request
 assignees: ''
 
 ---
-
-## Summary
-Please provide a brief summary of your proposal. Two to three sentences is best here.
-
-## API Changes
-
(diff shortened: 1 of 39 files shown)


jfversluis:
Took us a liiiiiitle while. But this looks good to me. Thanks so much for your very first contribution @Adam-Langley !

jfversluis (approved):

PR #1957, opened 2022-01-21
Fixed local debugging builds

dimonovdd:
After the last changes, the project is not building because of `mdoc` and [`Microsoft.DotNet.XHarness.TestRunners.Xunit`](https://github.com/xamarin/Essentials/blob/2627bf2db0c009392e6558f31b9844fa2d4185fd/DeviceTests/DeviceTests.Android/DeviceTests.Android.csproj#L64)

### DIFF
--- a/Xamarin.Essentials/Xamarin.Essentials.csproj
+++ b/Xamarin.Essentials/Xamarin.Essentials.csproj
@@ -31,11 +31,12 @@
     <MDocDocumentationDirectory>$(MSBuildThisFileDirectory)..\docs\en</MDocDocumentationDirectory>
     <DebugType>portable</DebugType>
     <Configurations>Debug;Release</Configurations>
+		<_IsRelease Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">true</_IsRelease>
  </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)'=='Debug' ">
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">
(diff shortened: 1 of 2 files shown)


dimonovdd:
I can see that the build failed, but unfortunately I don't have access. 

image

mattleibow:
The build failed because you are not an internal contributor and probably should not have run at all. I'll check why.

dimonovdd:
@mattleibow There are still problems.


image

Xamarin.Essentials/Xamarin.Essentials.csproj
```diff
@@ -31,11 +31,12 @@
     <MDocDocumentationDirectory>$(MSBuildThisFileDirectory)..\docs\en</MDocDocumentationDirectory>
     <DebugType>portable</DebugType>
     <Configurations>Debug;Release</Configurations>
+		<_IsRelease Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">true</_IsRelease>
```
dimonovdd, line 34: Can we describe a more specific condition for release builds? Maybe, does MSBuild on CI add any specific Properties?

mattleibow: Maybe instead of the _IsRelease variable, we can use the `'$(TF_BUILD)' == 'True' Or '$(CI)' == 'True'`.

That only gets triggered on CI.

But, I was unable to reproduce your errors. Did you check out submodules with the repo? The docs repo is needed in order to generate the xml docs.

dimonovdd: Yep, Does this really work on mac?

jamesmontemagno: Any additional thoughts on this to push it through? @mattleibow ?

jfversluis:
Not sure if we still need this at this point. Closing for now. Thanks for the time and effort on this one <3

PR #1976, opened 2022-03-01
Fix #1972 - allow restricted permissions for geolocation on android

### LINKED ISSUE #1972: [Bug] Approximate location on Android 12 doesn't work correct.
### Description
      Can not get approximate location on android 12 because:
      line (Permissions.EnsureGrantedAsync<Permissions.LocationWhenInUse>())  expected only permissions granted:
  https://github.com/xamarin/Essentials/blob/72dda322aaa583873eb517b44fdcba3e87e5f35f/Xamarin.Essentials/Geolocation/Geolocation.android.cs#L40

### Steps to Reproduce

1. Select approximate location in popup
2. Try to get permissions for locationInUse => Restricted 
3. Try to get location => inner exception

### Expected Behavior
     Geolocation gets approximate location on android 12
### Actual Behavior
    Geolocation throw exception
 ### Basic Information

- Version with issue:
- Platform Target Frameworks: 
  - Android:

aritchie:
### Description of Change ###
Allow restricted permissions for "gatekeepers" on android geolocation methods

### Bugs Fixed ###

- Related to issue #
(https://github.com/xamarin/Essentials/issues/1972)

### API Changes ###

Optional parameter added to Permissions.EnsureGrantedAsync that also allows restricted to pass internal gatekeepers.  Default is set to false, so only methods where this is required functionality will care

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Geolocation/Geolocation.android.cs
+++ b/Xamarin.Essentials/Geolocation/Geolocation.android.cs
@@ -19,7 +19,7 @@ namespace Xamarin.Essentials
 
         static async Task<Location> PlatformLastKnownLocationAsync()
         {
-            await Permissions.EnsureGrantedAsync<Permissions.LocationWhenInUse>();
+            await Permissions.EnsureGrantedAsync<Permissions.LocationWhenInUse>(true);
 
             var lm = Platform.LocationManager;
             AndroidLocation bestLocation = null;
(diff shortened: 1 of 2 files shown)


jamesmontemagno:
Should we add  a new method instead called "EnsureGrantedOrRrestrictedAsync". I am not a super fan of more overrides.. but an option. @jfversluis @Redth

jfversluis:
I think I have to agree with James. Although it's a nitty thing, I think a new method that explicitly mentions `Restricted` is a bit more discoverable and clear. 

Can we steer it to that? I think if we do that we can merge this pretty quickly. Let me know if you can do it Allan otherwise I can do it :)

aritchie:
@jfversluis Update sent with reverted signature and new signature methods

jfversluis (approved):
LGTM, thanks again Allan!

Issue #2009, opened 2022-05-10
[Bug] Permissions.CheckStatusAsync<Permissions.LocationWhenInUse> throws exception with coarse location


jtorvald:
### Description
I'm trying to get the location "while in use" from an Android device. So I first check if I have permission.
It can be that I don't understand how to call the API, but I expect to do something like this before calling the GetLocationAsync, so that I can present some information to the user first.

But when I call:
`Permissions.CheckStatusAsync<Permissions.LocationWhenInUse>`

I get an exception thrown:
`You need to declare using the permission: android.permission.ACCESS_FINE_LOCATION`

While Android docs say that if you have only COARSE location permission it should still work. The code suggests the same. So it's either because it's already too late or I'm just no used to reading this coding style, but doesn't this row make both permissions required:
[public override (string androidPermission, bool isRuntime)[] RequiredPermissions =>](https://github.com/xamarin/Essentials/blob/main/Xamarin.Essentials/Permissions/Permissions.android.cs#L246)

@aritchie can it be that this is something from 1.7.3 release, or a general bug, or should I call the API in an other way?

aritchie:
@jtorvald add the permission to the manifest.  I don't know what the docs say.  My two cents is that the docs are wrong, but the Microsoft guys can decide that.

jtorvald:
Thanks for the quick reply. I'm talking about the Android docs so I assume they are correct. The exception is from the check permissions code in essentials. If I'm not using fine location, I also prefer not to request permissions for it.

jtorvald:
I believe these methods should not even throw an exception since it's just checking for a status. This works fine on Android without having the permission in the manifest. Therefor I created a PR #2010 to change this to a permission.Denied and a Debug log line.

aritchie:
The issue is that this doesn't behave consistently with iOS in two aspects. 
1. iOS always requests precise location
2. iOS will throw an exception in a plist permission is missing

I believe Essentials is going to need a new permission set for their API LocationWhenInUse(bool fine) and LocationBackground(bool fine).

jtorvald:
@aritchie are you sure about 2.?

If I run it on iOS without the permission permission in the info.plist:
```
            Console.WriteLine("Enabled: {0} ", CoreLocation.CLLocationManager.LocationServicesEnabled);
            Console.WriteLine("Status: {0} ", lm.AuthorizationStatus);
```
Enabled: True 
Status: NotDetermined 

[...]

PR #2010, opened 2022-05-11
Permissions check method shoud not throw exception when permissions not in manifest

jtorvald:
### Description of Change ###

I believe that the Check methods should not throw exceptions since it just asks if it has permissions.
If you request the same with the Android API you also don't get exceptions, just Permission.Denied.

To be able to use Geo Location without FINE_LOCATION ([which is recommended now](https://developer.android.com/training/location/permissions#approximate-request)) it should not crash when only COARSE_LOCATION permission is set.

Also added permission to write external storage in the Android Sample app since it's needed for the MediaPicker.

### Bugs Fixed ###

Related to issue(s)
- #2009

### API Changes ###

No API changes. Just removed exceptions and replace with line in debug console to set the correct permission.
 
### Behavioral Changes ###

People might try/catch for the exception but get a permission denied now. Which is only an improvement.

### PR Checklist ###

- [ ] Has tests => simple change checked with the sample repo
- [ ] Has samples => simple change checked with the sample repo
- [ ] Updated documentation

### DIFF
--- a/Samples/Samples.Android/Properties/AndroidManifest.xml
+++ b/Samples/Samples.Android/Properties/AndroidManifest.xml
@@ -1,8 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.xamarin.essentials" android:installLocation="auto">
 	<uses-sdk android:minSdkVersion="19" android:targetSdkVersion="30" />
-	<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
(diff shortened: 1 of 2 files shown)


jfversluis:
Thanks for the input and the PR here @jtorvald!

However, I can't merge it this way. This does not only affect the location permissions, the changes are happening in the base permission which will now affect _all_ permissions and is therefore a big breaking changes across the whole Essentials library.

Ideally, especially for Xamarin.Essentials, we will want to keep the impact as small as possible. If there is something that needs to be tweaked for the location that is fine, but we should aim to keep the scope only to locations on Android.

jtorvald:
@jfversluis I understand. I will see if I've time to create a new PR specific to location on android.

Issue #2035, opened 2022-08-12
[Bug] Application hangs when using Connectivity on x86_64 emulator


Snegovikufa:
### Steps to Reproduce

1. Create a new Android application with LLVM and AOT enabled.
2. Add `Connectivity.PlatformConnectionProfiles` somewhere
3. Run app on x86_64 emulator.

### Expected Behavior

Application does not hang.

### Actual Behavior

Application hangs.

### Basic Information
- Platform Target Frameworks: 
  - Android: 12
- Affected Devices: x86_64 emulator, Chromebook with Intel

jfversluis:
See https://github.com/xamarin/Essentials/pull/2036#issuecomment-1845465648

PR #2036, opened 2022-08-12
Reading network state in the modern way

Snegovikufa:
### Description of Change ###

I faced the issue with the x86_64 emulator (and Chromebook as well) when I was compiling it with the LLVM+AOT. The app started and hanged. After hours of investigation, with lots of logs added, I found the issue with the `Connectivity`. As I see, Connectivity used a lot of deprecated API. I updated its code according to Android recommendations:

https://developer.android.com/training/basics/network-ops/reading-network-state

### Bugs Fixed ###

- Related to issue #

#2035 

### PR Checklist ###


jfversluis:
After much consideration I have decided not to still merge this PR. While I think it definitely makes sense to get rid of the obsolete APIs, at this point in the Essentials lifespan this is a risky change that can cause breaking changes to people.

Thank you so much for your time and effort with this change!

Issue #2007, opened 2022-05-03
[Bug] Launcher doesn't open URL if ResizableActivity = false, behavior changes between 1.4 & 1.5


czuck:
### Description

Opening a uri using Launcher.OpenAsync using Version 1.4 on a Pixel 3 XL opens in a browser window full screen.  Using 1.5, the launcher opens the browser in multi window instead of full screen with no other changes to the app.  Adding ResizableActivity=false to either the MainApplication or MainActivity results in the browser not opening at all.

### Steps to Reproduce

1. Run the attached test app.  Press the button "Launch Website" and the web page opens full screen.  Note that is it running Essentials 1.4.0
2. Open NuGet on the solution and install Essentials 1.5.0.  Run the app and press the same button, note that the web page opens in multi screen mode.
3. In MainActivity.cs comment out the activity on lines 8-9 and uncomment the one on 10-11.  Note the only difference is the ResizableActivity=false added on line 10.  Run the app and press the button, nothing happens.  Open MainPage.xaml.cs and add a breakpoint on the Launcher command, line 16.  Press the button again and the breakpoint will hit, continue and observe that still nothing happens.
4. Return MainActivity.cs to it's initial state.  Open MainApplication.cs and uncomment the code.  Run again, again the breakpoint will hit but the web page will not open.
5. Install Essentials 7.2 and repeat steps 2-4.

### Expected Behavior
Upgrading from Essentials 1.4.0 to 1.5.0 shouldn't change how the browser is opened.
Adding ResizableActivity=false shouldn't stip the launcher from opening

### Basic Information

- Version with issue: 1.5.0 - 1.7.2
- Last known good version: 1.4.0
- IDE: VS2019
- Platform Target Frameworks: 
  - Xamarin:16.11.000.197
  - Android: 12.0.0.3
- Android Support Library Version: 
- Nuget Packages:
- NETStandard.Library 2.0.3
- Xamarin.Forms 5.0.0.2401
- Affected Devices: Pixel 3 XL

  
### Reproduction Link

[LauncherTest.zip](https://github.com/xamarin/Essentials/files/8614602/LauncherTest.zip)

jamesmontemagno:
Will be fixed in 1.7.4

PR #2058, opened 2022-11-18
Remove launch adjacent Fixes #2023

### LINKED ISSUE #2023: [Bug] Launcher: Pixel 4 or newer, Android 12 forces split screen when opening files with default app
### Description
On Pixel 4 devices or newer with Android 12, if a file is opened using `Launcher.OpenAsync()` the file is opened in split screen mode adjacent to the source app regardless of whether `android:resizeableActivity` is `false` in the manifest file.

The issue is that when it goes into split screen mode the source app is reloaded and any in-progress state is lost in that app.

I've narrowed the problem down to the following logic in the `PlatformOpenAsync()` method of the `Launcher.android.cs` file:

`#if __ANDROID_24__
            if (Platform.HasApiLevelN)
                flags |= ActivityFlags.LaunchAdjacent;
#endif`

Removing this logic fixes the issue.

### Steps to Reproduce

1. Deploy your Xamarin App to a Pixel 4 or newer device
2. Open a PDF or DOCX file from your Xamarin App using `Launcher`

### Expected Behavior
PDF or DOCX file should be opened in a new window and the source app state should remain as before if `android:resizeableActivity` is `false`

### Actual Behavior
PDF or DOCX file is open adjacent in split screen mode and the source app is reloaded, losing any in-progress state

### Basic Information

- Version with issue:
- Nuget Packages: Xamarin.Forms 5.0.0.2012, Xamarin.Essentials 1.7.1
- Affected Devices: Pixel 4a, Pixel 5

jamesmontemagno:
### Description of Change ###

We shouldn't do this. Browser has an option

### Bugs Fixed ###

- Related to issue #2023

### DIFF
--- a/Xamarin.Essentials/AppInfo/AppInfo.android.cs
+++ b/Xamarin.Essentials/AppInfo/AppInfo.android.cs
@@ -59,10 +59,6 @@ namespace Xamarin.Essentials
 
             var flags = ActivityFlags.NewTask | ActivityFlags.NoHistory | ActivityFlags.ExcludeFromRecents;
 
-#if __ANDROID_24__
-            if (Platform.HasApiLevelN)
-                flags |= ActivityFlags.LaunchAdjacent;
-#endif
             settingsIntent.SetFlags(flags);
 
             context.StartActivity(settingsIntent);
--- a/Xamarin.Essentials/Email/Email.android.cs
+++ b/Xamarin.Essentials/Email/Email.android.cs
@@ -21,10 +21,7 @@ namespace Xamarin.Essentials
         {
             var intent = CreateIntent(message);
             var flags = ActivityFlags.ClearTop | ActivityFlags.NewTask;
-#if __ANDROID_24__
-            if (Platform.HasApiLevelN)
-                flags |= ActivityFlags.LaunchAdjacent;
-#endif
+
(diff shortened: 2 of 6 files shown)


andreasbrostencab:
Any reason why no one is reviewing this?

jfversluis (approved):

Bejasc:
When can 1.7.4 be expected with this fix? 
In the meantime - is there a workaround?

nikolajskov:
> When can 1.7.4 be expected with this fix? In the meantime - is there a workaround?

I'm also waiting for the next release with a fix for this. In the meantime, I've stopped using the Xamarin Essentials Launcher and am using my own implementation without the LaunchAdjacent flag for now.

jamesmontemagno:
Fixes #2007

shnaz:
We have exact same problem which should be fixed by this.

@jfversluis I see the 1.7.4 was released 3 days ago, according to https://github.com/xamarin/Essentials/releases/tag/1.7.4.

But it is not updated in nuget.org, is this intentional and why?
https://www.nuget.org/packages/Xamarin.Essentials

jamesmontemagno:
I think we had some CI/CD problems, should hopefully be fixed soon.

Issue #2061, opened 2022-11-24
[Bug] [UWP] App crashes when picking more than 1,000 files


DanielGilbert:
### Description
When picking more than 1000 files on UWP at once using `PickMultipleAsync()` or when picking 1000 different files in a row with `PickAsync()`, the application will crash with a `System.Exception`, stating that:

`The maximum number of items in the access list has been reached. An item must be removed before another item is added.`

This is because the [platform specific implementation](https://github.com/xamarin/Essentials/blob/df5ddea17601c788595842b91ac1db38df61227a/Xamarin.Essentials/FilePicker/FilePicker.uwp.cs#L39) for UWP is adding every file to [StorageApplicationPermissions.FutureAccessList](https://learn.microsoft.com/en-us/uwp/api/windows.storage.accesscache.storageitemaccesslist?view=winrt-20348), which has a maximum storage capacity of 1000 files. This list is kept between app restarts and presumably also updates, and should only be managed by the app itself.

The issue surfaced for us after many months of usage, were a single file was added on a day-by-day base.

### Steps to Reproduce

1. Call `PickMultipleAsync` for UWP
2. Pick more than 1000 files (or less, depending on how many have been picked before)
3. Click "Open" in the file dialog.

-- or --

Checkout the attached archive, which has a sample UWP app and 1001 random files. Simply click on _PickFiles_ and select all files from the _Sample Data_ folder. Then press "Open" and wait for the app to crash.

### Expected Behavior
The app should continue to work as expected.

### Actual Behavior
The app takes a long time to load all files and will finally crash with a `System.Exception`, after the maximum amount of items has been reached.

### Basic Information

- Version with issue: 1.3.1+, after the FilePicker got added
- Last known good version: None
- IDE: Visual Studio 2022
- Platform Target Frameworks: 
  - UWP: 10.0.17763  

### Reproduction Link

[FilePickerUwpTest.zip](https://github.com/xamarin/Essentials/files/10084013/FilePickerUwpTest.zip)

DanielGilbert:
In case you need help implementing a fix on this item (PR or anything), I'm happy to sign a CLA and contribute. :)

jfversluis:
@DanielGilbert your best shot at getting this fixed is probably if you indeed contribute. So if you're willing and able, you might want to give that a shot. Thanks!

DanielGilbert:
[...]

PR #2068, opened 2022-12-18
FilePicker in UWP now handles more than 1,000 files

DanielGilbert:
### Description of Change ###
This moves the addition of every file to the `FutureAccessList` out of the library, as the returned token was never used. If really needed, the app can add the `StorageFile` object to the `FutureAccessList` on their own, as stated in [Microsoft's Remarks on this class.](https://learn.microsoft.com/en-us/uwp/api/Windows.Storage.AccessCache.StorageApplicationPermissions?redirectedfrom=MSDN&view=winrt-22621#Windows_Storage_AccessCache_StorageApplicationPermissions_FutureAccessList)

I couldn't see a way to add tests for the file picker, as its part of the system UI, so I omitted this. Also, no samples were needed, as it's not a feature that got added. Same goes for the documentation, as the previous behaviour wasn't documented.

Of course, I'm open to suggestions. :)

### Bugs Fixed ###
- #2061
 
### API Changes ###
None

### Behavioral Changes ###
The library no longer adds every `StorageFile` object to the `FutureAccessList`.

### PR Checklist ###


jfversluis:
So, in order to not lose this entire functionality here, could we consider adding the first 1000? Or better yet I guess, you want to have the latest ones, so make sure to only take the _last_ 1000 of whatever was picked?

DanielGilbert:
Yeah, that won't solve the problem either... I just realized that there is no way to get the `IStorageFile` object from the `FileResult`. I'm currently thinking about how to tackle that. The main issue is that, after time, the list will run full if it's in the scope of the library - and you, the consuming developer, will simply have no clue what's going on. 😅
And also, it's not possible to get the hash you'd need to access the item from the list in the current implementation.

I'm afraid this needs a cleaner implementation. So, the problem we currently have is quite easy to reproduce. Taking my example project from the PR - if you extend it by this:

```csharp
var fileResult = await FilePicker.PickMultipleAsync(); //Pick at least one file

var firstResult = fileResult.FirstOrDefault();
string fullPath = firstResult.FullPath;

//Let's pretend we do not use firstResult.OpenReadAsync(),
//but instead create a new object:

var fileStream = File.OpenRead(fullPath); //System.UnauthorizedAccessException: 'Access to the path 'D:\Sample Data\random00000.test' is denied.'
```
You won't be able to access the file from the path alone. The implementation as-is is kind of useless - unless you go and iterate through the list itself, which is quite time consuming.

**For UWP, it is necessary to have access to the `IStorageFile` object that the FilePicker returns.** Either that, or we need a way to explicitly add it to the `StorageApplicationPermissions.FutureAccessList` from the FileResult, where we also return the associated hash for that list:

```csharp
string hash = FileResult.AddToFutureAccessList();
```
With that, each item in the UWP implementation could be stored for later use without exposing the `IStorageFile`. And the function would return the hash that is necessary to retrieve the item later. Which would, of course, be a `IStorageFile`. So that would need a platform specific wrapper. 

I would love to have a clean implementation on this one.

jfversluis:
But we know how many files have been picked, right?

So can't we just add something smart like (pseudo code):

```csharp
var skipCount = resultList.Count - 1000;

foreach (var file in resultList.Skip(skipCount))
    StorageApplicationPermissions.FutureAccessList.Add(file);
```

If we picked 1001 files, `skipCount` will now be 1 and the foreach will skip the first one and then proceeds to add the next 1000 entries. That way we still always get the last 1000, not losing any existing functionality, but it shouldn't break anymore. Right?

DanielGilbert:
> That way we still always get the last 1000, not losing any existing functionality, but it shouldn't break anymore. Right?

I wish it'd be that easy. But it's possible that there are already entries in this list, as the list is managed by the app itself. Sure, we could check how many entries are already in this list, and substract that from the amount of entries we could add. But that feels like duct-taping. Also, when selecting the first 1000 files, and then again selecting 1000 files, they won't get added to the list. But yes, it would solve the exception. 

How do you think about my last paragraph, regarding the proposed cleaner implementation? I'd be willing to extend my PR in this regard, I just didn't manage to build `Xamarin.Essentials` on my system yet, as I'm running into errors regarding several dependencies.

jfversluis:
I think that solution would imply adding more (public-ish) APIs which is what we try to avoid mostly unless we really need to do that for support scenarios and I wouldn't think this falls into that category to be honest.

@jamesmontemagno do you have any objections to this piece being removed? Technically its a breaking change from a functional standpoint, but my gut feeling says not too many people are heavily relying on this and they can easily implement something of their own in their own app which is "the right way"™️ of doing this anyway imo

jfversluis:
I'm not sure what that means @mattleibow ?

PR #2077, opened 2023-02-23
Update info about Xamarin.Essentials in .NET 6

jamesmontemagno:
Updating about .net 6 and 7 and maui stuff in docs. Thoughts?


jfversluis (approved):
Good idea, just fixed a couple of typo looking things. Thanks!

PR #2093, opened 2023-07-06
Make UWP FilePicker FutureAccessList more robust

### LINKED ISSUE #2061: [Bug] [UWP] App crashes when picking more than 1,000 files

jfversluis:
### Description of Change ###

When using FilePicker on UWP it will add the entries to something that is called the FutureAccessList. However, that list has a maximum amount of entries and when it reaches that, it will break. This change will add some more robustness.

This adds:
* Check if we're going over the limit with the picked files
* If yes; clean up the list starting with the oldest entries (assumed)
* Add the last _n_ number of picked files back to the list, where _n_ is the amount of files picked, or if that exceed the maximum of the FutureAccessList, it will take that maximum.
* Wraps this whole process in a try/catch so we don't break users apps when something goes wrong, its not that important.

### Bugs Fixed ###

- Fixes #2061
- Closes #2068

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/FilePicker/FilePicker.uwp.cs
+++ b/Xamarin.Essentials/FilePicker/FilePicker.uwp.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Windows.Storage;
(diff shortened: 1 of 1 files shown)


jfversluis:
@DanielGilbert are you still interested in this one? How does this look?

DanielGilbert:
> @DanielGilbert are you still interested in this one? How does this look?

Hi @jfversluis !

It's an improvement to the existing solution, thanks! Although I still stand by my comment, that a library should never interfere with the list. But it is how it is, I guess. :)

PR #2111, opened 2023-12-02
[Android] [Battery] Remove code duplication

Samidarex:
### Description of Change ###

Extracted duplicate code, into a new method. Tests and samples were ommited as the changes doesn't alternate the current class behavior.

Duplicate code:

```csharp
try
{
    Platform.AppContext.UnregisterReceiver(receiver);
}
catch (Java.Lang.IllegalArgumentException)
{
    System.Diagnostics.Debug.WriteLine("{Receiver name} receiver already unregistered. Disposing of it.");
}
```

### Bugs Fixed ###

- None

### API Changes ###
None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/Xamarin.Essentials/Battery/Battery.android.cs
+++ b/Xamarin.Essentials/Battery/Battery.android.cs
@@ -59,7 +59,7 @@ namespace Xamarin.Essentials
             {
                 Platform.AppContext.UnregisterReceiver(receiver);
             }
-            catch(Java.Lang.IllegalArgumentException)
+            catch (Java.Lang.IllegalArgumentException)
             {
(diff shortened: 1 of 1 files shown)


jfversluis:
Hey @Samidarex thanks so much for your work on this!

Looking at the Xamarin.Essentials lifespan that remains we want to only change the bare minimum so that we don't risk any regressions. Since this only addresses a thing to make the code a bit cleaner and not much else I'm going to close this one for now. You might want to check if this is still the same in .NET MAUI and maybe make your change also there!

Again thanks, much appreciated. If you're looking for any work to do here please reach out and we can surely find you something 😉

Issue #2122, opened 2024-02-01
OrientationSensor on iOS does not work when Location Services are off.


robheffo79:
### Description
OrientationSensor in iOS does not fire events due to an underlying issue where the sensor is intialised in such a way as it has a dependency on the Compass Calibration, and thus a dependency on location services.

### Steps to Reproduce

1. Attempt to enable the OrientationSensor on iOS
2. Wait for an event that never comes.

### Expected Behavior
You expect the ReadingChanged event to start firing as the orientation of the device changes.

### Actual Behavior
You do not get a ReadingChanged event to fire as the orientation of the device changes.

### References
[https://github.com/xamarin/Essentials/issues/1473](https://github.com/xamarin/Essentials/issues/1473)

jfversluis:
See https://github.com/xamarin/Essentials/pull/2123#issuecomment-1923452692

Issue #19, opened 2018-02-24
Exceptions


Redth:
There are some patterns in several API's that we might want to have some common exception types for.

## Feature Availability
Things like GPS, Cameras, Phone, SMS, etc are not always available on a given device.  When a method is called which tries to make use of such a device, we should throw a consistent exception to let the developer know the feature is unsupported on the current device:

`NotSupportedOnDeviceException : NotSupportedException`

There may also be the case where a feature is technically supported by the device, but not enabled:

`NotEnabledOnDeviceException : InvalidOperationException`

Open to ideas on naming, and other common exceptions we may require...

mattleibow:
What if you use the system exceptions as the base types:

 - `NotSupportedOnDeviceException : NotSupportedException`
 - `NotEnabledOnDeviceException  : InvalidOperationException`

Redth:
Yes!

mattleibow:
I think we also need to investigate the name of the exceptions: `NotSupportedOnDevice` vs `DeviceNotSupported` vs `FeatureNotSupported`.

F1nZeR:
We might use capability prefix for that exceptions: 
`CapabilityNotSupportedException` and `CapabilityNotEnabledException`.

mattleibow:
@jamesmontemagno @Redth we need to properly decide what exceptions we throw in cases, and when we just do nothing.

We currently have different ways of indicating that nothing can be done:

 - do nothing
 - throw `FeatureNotSupportedException`
 - throw `PlatformNotSupportedException`
 - throw `FeatureNotEnabledException`

One case where we do nothing is in Android's energy saver API: if we are not Lollipop, then we just fall through:
https://github.com/xamarin/Essentials/blob/1.0.0/Xamarin.Essentials/Battery/Battery.android.cs#L12-L48

One case where we throw `FeatureNotSupportedException` is with the sensors: if there is no sensor X, then we throw.

One case (and the only case so far) where we throw `PlatformNotSupportedException` is in Android's text-to-speech API: if we try and speak, but we couldn't initialize, then we throw.

So far, I was able to determine that we throw `FeatureNotSupportedException` for:
 - the sensors on all platforms if we aren't able to access the hardware
    - we throw in the start and the stop (this one may be overkill, we can probably first check to see if it is started, and if not then just do nothing)
 - the Android external browser if there was no browser installed
 - the email API
    - Android: if there is no `message/rfc822` intent handler
[...]

PR #2123, opened 2024-02-01
Change CMAttitudeReferenceFrame to XMagneticNorthZVertical to remove …

robheffo79:
…the dependency on the compass calibration, thus removing the dependency on location services.

### Description of Change ###

Changed the CMAttitudeReferenceFrame value from XTrueNorthZVertical to XMagneticNorthZVertical in OrientationSensor.ios.watchos.cs line #19.

This change is trivial and requires no tests, samples or documentation.

### Bugs Fixed ###

[- Related to issue #2122](https://github.com/xamarin/Essentials/issues/2122)

### API Changes ###

None

### Behavioral Changes ###

Orientation Sensor takes calibration from Magnetic North instead of True North. True north requires Location Services to be enabled for the app, introducing a sensitive permission that is not strictly necessary. This change pulls the functionality into line with the behavior of Android, which uses Magnetic North for its calibration.

### PR Checklist ###

- [ ❌ ] Has tests (if omitted, state reason in description)
- [ ❌ ] Has samples (if omitted, state reason in description)
- [ ✔️ ] Rebased on top of `main` at time of PR
- [ ✔️ ] Changes adhere to coding standard
- [ ❌ ] Updated documentation ([see walkthrough](https://github.com/xamarin/Essentials/wiki/Documenting-your-code-with-mdoc))

### DIFF
--- a/Xamarin.Essentials/OrientationSensor/OrientationSensor.ios.watchos.cs
+++ b/Xamarin.Essentials/OrientationSensor/OrientationSensor.ios.watchos.cs
@@ -16,7 +16,7 @@ namespace Xamarin.Essentials
             manager.DeviceMotionUpdateInterval = sensorSpeed.ToPlatform();
 
             // use a fixed reference frame where X points north and Z points vertically into the sky
-            manager.StartDeviceMotionUpdates(CMAttitudeReferenceFrame.XTrueNorthZVertical, Platform.GetCurrentQueue(), DataUpdated);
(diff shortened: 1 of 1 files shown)


jfversluis:
This would be a breaking change and I don't think we want to do that in this phase of the Xamarin.Essentials lifecycle as we're only months away from sunsetting it.

Thanks for your time and effort here, you might want to consider proposing this still for .NET MAUI as that likely has the same behavior.

PR #2137, opened 2024-04-29
Update readme may 1

rachelkang:
### Description of Change ###

Update README in preparation for May 1


mattleibow:
Not sure if it is worth mentioning that there is a maui nuget that is basically the same and used the same as the Xamarin.Essentials: https://www.nuget.org/packages/Microsoft.Maui.Essentials

We currently support netCurrent and netCurrent-1. For example, the 8.x series supports .net 8 and .net 7. The upcoming 9.x series will support .net 9 and .net 8.

jfversluis (approved):

